PR #39, opened 2018-08-17
Fix for issue eclipse/thingweb.node-wot#34

6d77:
* parse multiple parameters of content type and pass them to the codec
* pass parameters and schema during encoding and decoding

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -158,7 +158,7 @@ class ConsumedThingProperty extends TD.PropertyFragment implements WoT.ThingProp
                 reject(new Error(`ConsumedThing '${this.getThing().name}' did not get suitable client for ${form.href}`));
             } else {
                 console.log(`ConsumedThing '${this.getThing().name}' writing ${form.href} with '${value}'`);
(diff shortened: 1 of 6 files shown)


danielpeintner:
Minor question: Does it make sense to still have  a *default* schema `{}` if the second parameter in `valueToContent` is not set? Hence the parameter `schema` is optional and internally `{}` could kick in indicating that there are no restrictions on the value?

6d77:
Definitely yes. However this rises another implementation question to me as a TypeScript novice: 

The schema parameter is defined as `schema: WoT.DataSchema` and according to the definition of that type __MUST__ have a `type` attribute and programmers would not see why they should check for existence of that attribute. However as you mentioned, `type` is indeed optional. So is the definition of the `valueToContent`and `contentToValue` methods ok? 

Also `{}` would not be an instance of a `WoT.DataSchema` according to that definition.

danielpeintner:
> Also {} would not be an instance of a WoT.DataSchema according to that definition.

I opened https://github.com/w3c/wot-scripting-api/issues/133 and I think we should fix this.

mkovatsc:
https://github.com/w3c/wot-scripting-api/issues/133 was closed; `type` is not required anymore.

PR #45, opened 2018-10-16
Observe property

sebastiankb:
- enables 'observe-able' Properties in node-wot for consumer and exposer wot servient
- tested with MQTT
- open: rel assignment for wot exposer (tbd: Form class has to be extended)


danielpeintner:
I assume we can close this PR, correct @sebastiankb ?

PR #55, opened 2018-12-06
fix(core): fix unhandled property write rejection

### LINKED ISSUE #54: ExposedThingProperty.write() does not handle rejected promises
The `write(value)` method of the `ExposedThingProperty` class which is called when a request is received to change the value of a property,  returns a promise. 

It checks if a property `writeHandler` is defined, and if so, get's the promise this handler returns, waits for it to resolve, does some logging and internal state updating, and then resolves it own promise accordingly. 

However, if the promise returned by writeHandler is rejected, there is not catch() to handle this rejection. This means that the promise returned by write() is never rejected and just hangs.

As an example, if my writeHandler rejects a certain property write (for example because it's invalid) , this results in an `UnhandledPromiseRejectionWarning` being thrown and in the write request hanging indefinitely.

A possible solution would be to add:
`.catch((err) => { reject(err); })`
after the then() block in exposed-thing.ts line 294.

hasbel:
fix(core): fix unhandled property write rejection

if the promise returned by a custom writeHandler for a property of ExposedThing is rejected, this rejection should be handled.

fixes #54

### DIFF
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -291,6 +291,10 @@ class ExposedThingProperty extends TD.ThingProperty implements WoT.ThingProperty
                             }
                             this.getState().value = customValue;
                             resolve();
+                        })
+                        .catch((customError) => {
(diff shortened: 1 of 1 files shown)


danielpeintner:
Look good to me.
+1 for merging

PR #57, opened 2019-01-11
fix: add the correct op to forms if readOnly or writeOnly are true

### LINKED ISSUE #49: op field in generated Thing Descriptions
In the TDs automatically generated by node-wot, the op field in the forms of a properties is always an array that contains both "readproperty" and "writeproperty". (even if writable is set to false.)

This field is defined as a string and not an array according to the spec draft. 
So the expected behaviour is for it to be set to "readproperty" if the property is not writable and to be set to "writeproperty" if the property is write only. 

I am not sure what the correct value should be when a property is both readable and writable though.

hasbel:
Fixes #49 


danielpeintner:
Thanks! I will merge it.

Please not that are discussions in the WoT TD group to slightly change the op fields (see https://github.com/w3c/wot-thing-description/issues/236#issuecomment-449293523).
We should take this into account if this is going to happen

PR #59, opened 2019-01-16
Add clientId

wiresio:


sebastiankb:
There seems to be some issues. Please check the travis report.

Btw: It would be great if you can update the README of the mqtt packacke as well as the https://github.com/eclipse/thingweb.node-wot/blob/master/packages/cli/src/cli.ts regarding this new setting option.

PR #60, opened 2019-01-16
Op forms

sebastiankb:
- includes an op-forms handler for the consumer
- includes observable properties for an http based provider (longpoll)


danielpeintner:
Thanks!

I assume https://github.com/eclipse/thingweb.node-wot/issues/56#issuecomment-452646911 is resolved with this PR also, correct?

PR #75, opened 2019-03-18
Fix(http-client): Fix content type handling

### LINKED ISSUE #73: WoT.fetch(): contentType check always returns false
In the current `wot-impl.ts` implementation, when a `fetch()` operation is performed, the `contentType` of the response is checked to make sure it is either `"application/td+json"` or `"application/ld+json"`.

```typescript
if (content.type !== ContentSerdes.TD &&
    content.type !== ContentSerdes.JSON_LD ) {
    console.warn(`WoTImpl received TD with media type '${content.type}' from ${uri}`);
}
```

However this check always returns false when using a http client right now. 
this is because the http client uses the function checkResponse() to generate the content object, and this function currently creates an object with the wrong attributes.

hasbel:
Closes #73


sebastiankb:
thanks

PR #79, opened 2019-03-26
fix(binding-http): Add support for CORS preflight requests

### LINKED ISSUE #78: CORS pre-flight requests are rejected
When trying to access a node-wot Thing (server mode) from a browser application hosted on a different domain, a Cross-Origin Resource Sharing (CORS) pre-flight request is needed [in certain cases](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).

This is an HTTP OPTIONS request, and currently it always returns with a 405 Error, making the Thing inaccessible from a browser when it sends automatic CORS preflight requests.

hasbel:
fix(binding-http): Add support for CORS preflight requests

closes #78


sebastiankb:
thanks

PR #82, opened 2019-04-02
Browser

hasbel:
Add a browser-bundle package and an example of how to use it. Update docs accordingly.

The browser-bundle packages build to a 167kb javascript file that can be used as browser-side library and includes:
- core
- binding-http (client only)
- binding-websocket (client-only)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,15 @@ W3C Web of Things implementation on NodeJS
 
 [![Build Status](https://travis-ci.org/eclipse/thingweb.node-wot.svg?branch=master)](https://travis-ci.org/eclipse/thingweb.node-wot)
 
+### Table of Contents
+- [Licence](#licence)
+- [Prerequisites](#prerequisites)
+- [How to get the library](#howtoget)
+- [Start with an example](#example)
+- [How to use the library](#api)
+
+<a name="license"/>
+
 ## License
 Dual-licensed under both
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -98,7 +134,24 @@ To reduce the size of the installation from about 800 MByte down to about 200 MB
    * try `npm run unlock` from project root before calling `[sudo] npm run link`
    * try `npm link` in each package directory in this order: td-tools, core, binding-\*, cli, demo-servients
 
-## No time for explanations - I want to start from something running!
+### As a browser library
+
+Node-wot can also be imported as browser-side library. To do so, include the following `script` tag in your html:
+```html
+<script src="https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@latest/dist/wot-bundle.min.js"></script>
```
danielpeintner: I assume this CDN links works once the wot-bundle is pushed on NPM, Correct?
Currently it fails.

Existing packages like https://cdn.jsdelivr.net/npm/@node-wot/td-tools@latest/dist/td-tools.min.js work.

Shall we add a note to readers?

examples/browser/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <title>Browser Native node-wot tryout</title>
+    <!-- Foundation CSS framework -->
+    <link rel='stylesheet' href='https://cdn.jsdelivr.net/foundation/5.0.2/css/foundation.min.css'>
+
+    <!-- Font Awesome icons -->
+    <link rel='stylesheet' href='https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.0.3/css/font-awesome.css'>
+
+    <!-- defer to not block rendering -->
+    <script src="https://cdn.jsdelivr.net/npm/@json-editor/json-editor@latest/dist/jsoneditor.min.js" defer></script>
+    <script src="https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@latest/dist/wot-bundle.min.js" defer></script>
```
danielpeintner: Hence, this fails also...

examples/browser/index.html
```diff
@@ -0,0 +1,58 @@
...
+    <script src="index.js" defer></script>
```
danielpeintner: When loading in FF66 I besides the failing resource I get some more warnings.

e.g., ReferenceError: JSONEditor is not defined

Maybe it would be good to add some try catch code to let the user know that somethings did not work as expected.

README.md
```diff
@@ -3,6 +3,15 @@ W3C Web of Things implementation on NodeJS
...
 [![Build Status](https://travis-ci.org/eclipse/thingweb.node-wot.svg?branch=master)](https://travis-ci.org/eclipse/thingweb.node-wot)
 
+### Table of Contents
+- [Licence](#licence)
+- [Prerequisites](#prerequisites)
+- [How to get the library](#howtoget)
+- [Start with an example](#example)
+- [How to use the library](#api)
+
+<a name="license"/>
```
danielpeintner: All the anchors, also the others are not needed and I would not add them because they are already there and automatically updated (see https://github.com/eclipse/thingweb.node-wot#prerequisites etc)

danielpeintner:
@hasbel thank you for your great work and efforts. It looks very promising. I added some comments (see above).

At the moment the HTML example is not working given that the bundle is not on NPM yet.
Shall we put in on some other resource until we are ready. 

The problem I see is that otherwise it might confuse people...

danielpeintner:
One more comment: we have a *running* thing instance on http://plugfest.thingweb.io:8083/TestThing.

We could say that instead of starting the counter example one can use that URL instead.
(see http://plugfest.thingweb.io:8088/test/fullscreen/default)

README.md
hasbel: Yes, once it is on NPM it will work. This is actually the best approach in my opinion as the alternative would be to upload the compiled code on `/dist` to github and use the github link, which should be avoided, and it should not be a problem if a snapshot is released on NPM directly after this PR is merged.

I also think that it would be useful to have more SNAPSHOT releases available on NPM for node-wot, as currently anyone using NPM to get node-wot would be missing on all the changes made in the last 2 months.

The typescript-definitions for example are already on 0.7.0-SNAPSHOT-5.

examples/browser/index.html
hasbel: The first defer should be removed. Thanks I will fix this.

hasbel:
@danielpeintner Thanks a lot for the quick feedback! 

Regarding NPM, as I commented above, I think that it is the best approach when combined with a snapshot release. Having more snapshot releases of node-wot would be a good thing IMO.

Thanks for letting me know about the testThing. I replaced the counter and updated the table of content as you suggested.

PR #89, opened 2019-05-15
feat: add support for URI Variables on client side

danielpeintner:
TODO check other *possible* RFC6570 syntax (https://tools.ietf.org/html/rfc6570)


6d77:
See also the implementation at https://www.npmjs.com/package/uritemplate Perhaps this is useful instead of trying to implement this again.

danielpeintner:
> See also the implementation at https://www.npmjs.com/package/uritemplate Perhaps this is useful instead of trying to implement this again. This would also implement path parameters and not only query parameters.

@6d77 thanks, will look into that

I will also create a new PR because somehow Eclipse ECA seems to cause issues

PR #90, opened 2019-05-15
feat: add support for URI Variables on client side

danielpeintner:
TODO check other possible RFC6570 syntax (https://tools.ietf.org/html/rfc6570)

Work in progress (do not merge)!! will look into https://www.npmjs.com/package/uritemplate


danielpeintner:
ECA issue again

PR #99, opened 2019-05-21
Fix(cli): correctly read config file

### LINKED ISSUE #83: CLI does not correctly read config files
Loading a configuration from the command line in the CLI does not work as documented.

In the help, the function is documented as:
```
options:
-f, --configfile=file   load configuration from specified file
```

Doing this does not work.

hasbel:
Fixes #83


danielpeintner:
looks good to me

PR #106, opened 2019-05-30
update default conf

### LINKED ISSUE #104: wot-servient.conf is "misleading" for mqtt
There is no documentation when used `wot-servient -h` on how to use it for mqtt. If someone builds node-wot and does the initial tests with `counter.js` it will not work out of the box since the configuration uses username, password and client id. None of these are necessary to set up a connection.
I would suggest to remove username, password and client id from the default configuration but document them on `wot-servient -h` such that it works out of the box

egekorkan:
fixed #104


egekorkan:
Closing due to new PR #107

PR #107, opened 2019-05-31
fix: remove username, password and clientid from default config

egekorkan:
cli: add mqtt information on the output of wot-servient -h


egekorkan:
fixes #104

PR #120, opened 2019-06-24
fix(binding-http): correctly use security proxy

hasbel:
HttpClient says it will overwrite the local proxy if a security proxy is given. however, a bug prevented this from happening.


danielpeintner:
Looks good -> merge

PR #122, opened 2019-07-12
fix(http-binding): stop polling in case of error

hasbel:
When a subscription returns a response with an error code, don't keep polling forever.

Right now, when a subscription return a 404 for example (not a request error), polling() is directly called again, and return another 404, which results in an endless loop.


danielpeintner:
Thanks! @hasbel do you think you can create a test-case also that elaborates the expected behavior? That would be great!

hasbel:
@danielpeintner sure!

hasbel:
@danielpeintner tests added

danielpeintner:
@hasbel can you please look into the travis build issues

danielpeintner:
Thanks @hasbel

PR #129, opened 2019-10-16
Add support for debugging thing scripts

relu91:
Vm scripts must have `filename` option to be inspected. Filename option allows the node inspector to locate script source files and to support breakpoints inside the script. 
Example:
`node --inspect-brk packages/cli/dist/cli.js ../thing.js`

Futher details on [docs](https://nodejs.org/api/vm.html#vm_class_vm_script).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,25 @@ npm install @node-wot/binding-coap
 ```
 Alternatively you can add `@node-wot/<package-name>`as a dependency to your `package.json`.
 
+#### As a dev dependency
+
+If you want to develop applications for wot-runtime, you can use the command-line interface to run and debug your local scripts. First, install the CLI module as a dev-dependency:
+
+```
+npm install @node-wot/cli --save-dev
+```
+Then to start all `.js` files in the current directory use the following command:
+```
+wot-runtime
+```
+If you want to run a specific file or a list of files just append the file paths to the previous command:
+```
+wot-runtime script1.js ./src/script2.js
+```
+Finally, to debug use the option `--inspect` or `--inspect-brk` if you want to hang until your debug client is connected. Then start [Chrome Dev Tools](chrome://inspect) or [vscode debugger](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_attaching-to-nodejs) or your preferred v8 inspector to debug your code.
+
+For further details check: `wot-runtime --help`
+
 ### As a standalone application
 #### Clone and build
 Clone the repository:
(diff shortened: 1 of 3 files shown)


relu91:
Please wait to merge. I'm trying to start the inspector programmatically so that one can start the runtime like the following bash command: `wot-runtime --inspect thing.js`

sebastiankb:
ok.
I added the 'WIP' in the title to make the PR status clear.

danielpeintner:
@relu91 may I ask you to provide some documentation in [README.md](https://github.com/eclipse/thingweb.node-wot/blob/master/README.md) also.

Thanks!

relu91:
Done! 
Some points that I want to highlight about my implementation:
- I decided to handle the activation of the inspector in `cli.ts` instead of `servient.ts`. It seemed to be more natural to have such logic inside the CLI module.
- The `--inspect-brk` option waits for a debug client before to run the first script. It also adds a breakpoint on the first line of such a script. This is consistent with the `--inspect-brk` option of node.js. It allows debug clients to synchronize with the inspector server (as pointed in the commit message).

If you have any requests or comments, please let me know.

relu91:
Hi @danielpeintner, thanks for your feedback.
>Hence I understood your work to also support attaching VisualStudio and/or Chrome to a running instance. Correct? This would be very nice!

Yes, exactly! My contribution is twofold:
- Correctly load script inside the VM context so that can be recognized from the node inspector (first commit)
- Programmatically start node inspector from CLI (second commit)

I've just reproduced your steps and the servient started correctly. I'm using node v10.16.0, which version are you using?

That kind of error is usually generated if there are two debugger session attached to the same process. Have you tried to restart vscode? Anyway, I'll do my best to reproduce it!

>Moreover, I am not sure where the notion/command of wot-runtime comes from. I had to replace it with node packages\cli\dist\cli.js

`wot-runtime` is the command name published in your package.json. Users of the library can find it inside the `node_module\.bin` folder. Although, for testing purposes, it's correct to start the CLI as you did.

relu91:
Sorry for the double comment, but I have another question: 
Are you lunching the following command `node packages\cli\dist\cli.js --clientonly --inspect-brk examples\scripts\counter-client.js` from the command line or using `launch.json`?

danielpeintner:
Thanks for the quick feedback.

> I've just reproduced your steps and the servient started correctly. I'm using node v10.16.0, which version are you using?

I am using version v8.12.0.

> Are you lunching the following command node packages\cli\dist\cli.js --clientonly --inspect-brk examples\scripts\counter-client.js from the command line or using launch.json?

I am launching the command from the command line.

AndreMaz:
Sorry to intrude, but node v8 currently in maintenance mode is will reach EOL 2019/12/31.
Does it worth to support it?

danielpeintner:
Makes sense to use at least v10. Let me do some more checks and I will get back to you.

danielpeintner:
I did some further checks with v10 and it seems to work as expected. Thanks!

PR #138, opened 2019-11-05
Support MQTT 5 and cleanup servient configuration

wiresio:
Fixes [https://github.com/eclipse/thingweb.node-wot/pull/65](url)

Signed-off-by: Christian Glomb christian.glomb@siemens.com


danielpeintner:
Looks good, thanks!

PR #154, opened 2019-12-03
feat (binding-netconf): add support for NetConf binding

lukesmolo:
Add NetConf client protocol binding.
It currently supports the followings NetConf methods:

- GET-CONFIG
- EDIT-CONFIG
- COMMIT
- (CUSTOM) RPC

Subscriptions are still not implemented.

Signed-off-by: lukesmolo <lukesmolo@gmail.com>

### DIFF
--- a/packages/binding-netconf/README.md
+++ b/packages/binding-netconf/README.md
@@ -1,0 +1,130 @@
+# NETCONF Client Protocol Binding
+ 
+W3C Web of Things (WoT) Protocol Binding for NETCONF [RFC6241](https://tools.ietf.org/html/rfc6241)
+ 
+## Getting Started
+ 
+### Optional: NETCONF Server Simulator
+ 
+If a NETCONF server is needed, the simulator [Netopeer2](https://github.com/CESNET/Netopeer2) can be used.
+A Dockerfile for building the simulator Docker image can be found [here](https://github.com/opennetworkinglab/ODTN-emulator/blob/master/emulator-oc-cassini/Dockerfile), then just create a Docker image and launch the Docker container:
+ 
+```bash
+$ docker image build -t netconf
+$ docker run -d --name NETCONF netconf
+```
+ 
+The Docker container for the Server Simulator already comes with the necessary YANG modules loaded.
+Otherwise to manually install them just download the files and then:
+ 
+```bash
+$ sysrepoctl --install --yang=ietf-interfaces@2014-05-08.yang --owner=root:root --permissions=666
+$ sysrepoctl --install --yang=ietf-ip@2014-06-16.yang --owner=root:root --permissions=666
+$ sysrepoctl --install --yang=iana-if-type@2014-05-08.yang --owner=root:root --permissions=666
+```
+ 
+### Prepare the NETCONF Server
+ 
+Three YANG files must be loaded by the server in order to be able to run the examples:
+ 
+* [ietf-interfaces@2014-05-08.yang](https://github.com/YangModels/yang/blob/master/standard/ietf/RFC/ietf-interfaces%402014-05-08.yang)
+* [ietf-ip@2014-06-16.yang](https://github.com/YangModels/yang/blob/master/standard/ietf/RFC/ietf-ip%402014-06-16.yang)
+* [iana-if-type@2014-05-08.yang](https://github.com/YangModels/yang/blob/master/standard/ietf/RFC/ietf-interfaces%402018-02-20.yang)
+
+### Run the Example App
+ 
+The Binding example in the `./examples` directory provides a TD (`netconf-thing.jsonld`) and an app script (`netconf-example.js`) .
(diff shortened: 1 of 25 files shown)


sebastiankb:
Thank you so much for this PR! May I ask you to provide a README about this binding and an example script + TD. That would be awesome.

One more question: what kind of (secured) transport protocol is supported in this binding?

lukesmolo:
> Thank you so much for this PR! May I ask you to provide a README about this binding and an example script + TD. That would be awesome.

 Sure, no problem. I'll add them as soon as possible.
> One more question: what kind of (secured) transport protocol is supported in this binding?

This binding uses [node-netconf](https://www.npmjs.com/package/netconf), that relies on ssh for sending RPC requests. So I would answer that this binding supports TCP for the SSH connection. The security of the connection is then granted by the user/password or user/privateKey combinations provided in the servient configuration.

lukesmolo:
Ok, @sebastiankb you can find now an example and the README as well.

danielpeintner:
Question: you also changed `ProtocolClient` by adding schema information. Is this actually needed?

Moreover, if we change it we should use `TD.DataSchema` consistently and not mix `TD.DataSchema` and `TD.BaseSchema`.

lukesmolo:
> Question: you also changed `ProtocolClient` by adding schema information. Is this actually needed?

Yes, it is definitely needed actually for all those cases in which you need to be able to convert JSON structure into something different on the wire. This could be a general improvement for several other bindings too.
> Moreover, if we change it we should use `TD.DataSchema` consistently and not mix `TD.DataSchema` and `TD.BaseSchema`.

 I agree, the problem is that for some reasons the compiler complains if I use only DataSchema.

danielpeintner:
Excuse the long idle time due to the holiday season.

>> Moreover, if we change it we should use TD.DataSchema consistently and not mix TD.DataSchema and TD.BaseSchema.
>
> I agree, the problem is that for some reasons the compiler complains if I use only DataSchema.

Mhh, I will try to look into the issue and get back to you.

BTW, the `README.md` states as TODO "tests".
Do you plan to provide tests in this PR? We encourage to have at least some basic tests before merging...

lukesmolo:
I am sorry for this long period of silence, but I am actually busy with other stuff. I added some tests for the binding. In this case, since I cannot instantiate a Netconf server, I made some basic tests.

danielpeintner:
Thanks for the updates.

>>    Question: you also changed ProtocolClient by adding schema information. Is this actually needed?
>
>Yes, it is definitely needed actually for all those cases in which you need to be able to convert JSON structure into something different on the wire. This could be a general improvement for several other bindings too.

 May I ask you again to explain the reasons in more details since we didn't encounter any issues with the current approach. This a _core_ change and we would like to make sure it actually makes sense.

Can you share a concrete example so that we understand the issue better?
Did you look into `ContentSerdes` which is meant to handle different content types.

wiresio:
@lukesmolo I agree with @danielpeintner - would you mind giving us an explanation. Thanks!

lukesmolo:
Sorry again for the delay, I first wanted to discuss with @mkovatsc some details of the bindings. Yes, you are right, for this specific binding it should be feasible to use a new ContentSerdes. In the next days I should be able to work on it and to make a new commit. Thank you for the advice!
PS: Concerning the other OPCUA PR, I will answer on that page since the situation is a little bit different.

lukesmolo:
Ok, now hopefully it should be fine. Maybe the new ContentSerdes in the future could inspire a more generic XML ContentSerdes.

packages/binding-netconf/README.md
```diff
@@ -0,0 +1,128 @@
...
+ 
+Please note that, in order to make this binding work, each href should always contain a leaf that refers to a property of the YANG module.
+ 
+## TODO
+ 
+- [ ] Subscriptions implementation (EVENTS)
+- [x] TEST
+- [ ] (NETCONF Server Protocol Binding ?)
+ 
+## Acknowledgments
```
danielpeintner: Please remove any acknowledgments. We do not have it elsewhere in the project.

packages/browser-bundle/package.json
```diff
@@ -19,8 +19,7 @@
     "browserify": "16.2.3",
     "tinyify": "2.5.0"
   },
-  "dependencies": {
-  },
+  "dependencies": {},
```
danielpeintner: Please revert this change since this does not belong to this PR.

packages/binding-netconf/README.md
lukesmolo: Ok, done

packages/browser-bundle/package.json
lukesmolo: sure, done

danielpeintner:
@lukesmolo Thank you fort your work!

PR #156, opened 2019-12-19
feat (binding-opcua): add support for OPC UA binding

lukesmolo:
Add OPC UA client protocol binding.
It currently supports:

- reading/writing a property
- invoking an action (call a OPC method)
- observing a property (monitor an item)

Subscriptions with Pub/Sub protocol are still not implemented.

OPC UA uses custom Datatypes on the wire, which requires additional translations from the JSON model used at WoT application level. To enable support for a proper Datatype translation, the DataSchema information has to be extended and the schema also be passed down to the Binding.
Please take a look at the README file I included.

Signed-off-by: lukesmolo <lukesmolo@gmail.com>


danielpeintner:
Excuse the long idle time due to the holiday season and thanks for the contribution.

The README.md states as TODO "tests". Do you plan to provide tests in this PR? We encourage to have at least some basic tests before merging...

lukesmolo:
I am sorry for this long period of silence, but I am actually busy with other stuff. I added some tests for the binding.

lukesmolo:
Sorry again for the delay, I first wanted to discuss with @mkovatsc some details of the bindings.
Concerning [the problem](https://github.com/eclipse/thingweb.node-wot/pull/154#issuecomment-582322043) of using a new ContentSerdes for this binding, differently from the NetConf case, it is not possible to make a generic ContentSerdes used by this binding and (possibly) by others, since the binding itself is not handling directly the raw bytes for OPC UA, but instead an intermediate format to be passed to another nodejs library ([node-opc](https://github.com/node-opcua/node-opcua)) that is in charge to make the raw conversion. For this reason, this is a good example where it could be good to have the schema in the binding since, in practice, the binding is not using a different Content Type but a different way of handling another Content Type (in this case JSON).
The other option, of course, is to have a custom/specific ContentSerdes only for this binding.

packages/binding-opcua/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+A basic OPC UA server crawler and a basic OPC UA -> TD translator can be found [here](https://github.com/lukesmolo/wot-utils/tree/master/opcua-crawler)
+
+## TODO
+ 
+- [ ] Subscriptions implementation (EVENTS) with Sub/Pub protocol
+- [x] TEST
+- [ ] (OPC UA Server Protocol Binding ?)
+ 
+## Acknowledgments
```
danielpeintner: Please remove any acknowledgments. We do not have it elsewhere in the project.

packages/browser-bundle/package.json
```diff
@@ -19,8 +19,7 @@
     "browserify": "16.2.3",
     "tinyify": "2.5.0"
   },
-  "dependencies": {
-  },
+  "dependencies": {},
```
danielpeintner: Please revert this change since this does not belong to this PR.

packages/core/src/consumed-thing.ts
```diff
@@ -275,7 +275,7 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
                 let input;
 
                 if (parameter !== undefined) {
-                    input = ContentManager.valueToContent(parameter, <any>this, form.contentType);
+                    input = ContentManager.valueToContent(parameter, <any>ta, form.contentType);
```
danielpeintner: Change should be reverted?

packages/core/src/consumed-thing.ts
```diff
@@ -343,7 +343,7 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
                 reject(new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`));
             } else {
                 console.log(`ConsumedThing '${this.title}' unobserveing to ${form.href}`);
-                client.unlinkResource(form);
+                client.unlinkResource(form, tp);
```
danielpeintner: Change should be reverted?

packages/core/src/consumed-thing.ts
```diff
@@ -387,7 +387,7 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
                 reject(new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`));
             } else {
                 console.log(`ConsumedThing '${this.title}' unsubscribing to ${form.href}`);
-                client.unlinkResource(form);
+                client.unlinkResource(form, te.cancellation);
```
danielpeintner: Change should be reverted?

packages/core/src/protocol-interfaces.ts
```diff
@@ -32,7 +32,7 @@ export interface ProtocolClient {
   invokeResource(form: TD.Form, content: Content): Promise<Content>;
 
   /** this client is requested to perform an "unlink" on the resource with the given URI */
-  unlinkResource(form: TD.Form): Promise<void>;
+  unlinkResource(form: TD.Form, schema?: TD.BaseSchema): Promise<void>;
```
danielpeintner: Change should be reverted?

packages/binding-opcua/README.md
lukesmolo: ok, done

packages/browser-bundle/package.json
lukesmolo: sure, done

packages/core/src/consumed-thing.ts
lukesmolo: Actually here I do not understand why the whole Consumed Thing is passed down to the ContentSerdes  instead of only the Action Affordance. In my case I would have some problems in the Content Serdes. Maybe can you explain me why it is like this only for actions?

packages/core/src/consumed-thing.ts
lukesmolo: Sorry, I forgot to fix it. Done

packages/core/src/consumed-thing.ts
lukesmolo: Sorry, I forgot to fix it. Done

packages/core/src/protocol-interfaces.ts
lukesmolo: Sorry, I forgot to fix it. Done

packages/core/src/consumed-thing.ts
danielpeintner: You are right, this was a mistake on our side. Thanks!

lukesmolo: ok, good!

danielpeintner:
@lukesmolo Thanks for your work.

I merge the PR but may I ask you too look into the following issue which pops up sometimes (see https://travis-ci.org/eclipse/thingweb.node-wot/builds/651900865?utm_source=github_status&utm_medium=notification)

```
  1 failing

  1) OPCUA client test

       "before all" hook for "should read a property":

     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/eclipse/thingweb.node-wot/packages/binding-opcua/test/opcua-client-test.ts)
```

PR #158, opened 2020-01-08
Implement observable property

wiresio:
Please test. Not sure whether cancelling of observation works properly.

Signed-off-by: Christian Glomb christian.glomb@siemens.com


wiresio:
Withdrawn

Issue #161, opened 2020-01-17
[testing] mqtt test is not working/old


egekorkan:
The mqtt testing file is using skip (so not testing anything?) but at the same time, using the old API. I have tried the following but I haven't managed to call done() the way Mocha wants. Some more information available [here](https://blog.revathskumar.com/2016/11/javascript-testing-promises.html)
```
 WoT.produce({ 
                    title: "TestWoTMQTT",
                    events: {
                        event1: {
                            data: { type: "number"}
                        }
                    } 
                })
                .then((thing) => {
                            
                    thing.expose();
                    
                    console.info("Exposed", thing.getThingDescription().title);

                    WoT.consume(thing.getThingDescription())
                        .then((client) => {
                            let check = 0;
                            client.subscribeEvent("event1",
                                x => {
                                    expect(x).to.equal(++check);
                                    if (check === 3) done();
                                    
                                }
                            )
                            
    
                            var job = setInterval(() => {
                                ++counter;
                                thing.emitEvent("event1",counter); // sends data to the topic /TestWoTMQTT/events/event1
    
                                if (counter === 3) clearInterval(job);
                            }, 100);
                        });
                    });
            });
```

danielpeintner:
@egekorkan can you check whether https://github.com/eclipse/thingweb.node-wot/pull/398 resolved this issue?

egekorkan:
Yup, thank you @JKRhb ! :)

Issue #152, opened 2019-11-26
[proposal][binding-mqtt] Adding writeable property feature


egekorkan:
For a node-wot exposed thing:
In the current state, there is no way to write to a property through MQTT since only `observeproperty` and `unobserveproperty` are added for the forms no matter what, as mentioned in #126. In addition to `readproperty`, we can add the writeproperty functionality where a separate form is created with an `href` like `mqtt://myBroker/myThing/properties/myProperty/writeproperty` where the expected MQTT method would be to publish to this topic. This implies that the Exposed Thing should subscribe to this topic, like an action. [Bindings Template](https://w3c.github.io/wot-binding-templates/#mqtt-vocabulary) already explains how the default value for a `writeproperty` in MQTT should be. This would also imply that [line 150](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-mqtt/src/mqtt-broker-server.ts#L150) should be changed since there would be 5 segments in the topic.

Finally, a `writeproperty` form would look like the following:
```
        {
          "href": "mqtt://myBroker/myThing/properties/myProperty/writeproperty",
          "op": ["writeproperty"],
          "contentType": "application/json",
          "mqv:controlPacketValue": "PUBLISH"
        }
```

danielpeintner:
@wiresio any opinion/proposal?

wiresio:
I do not like the fact that we introduce the "deeper" level _writeproperty_. Isn't it sufficient to evaluate the "op" key to find out about the actual operation?

egekorkan:
In this case it is not possible since if I have two subscribers to the property of the exposed thing (publisher) and when one of them also publishes to the topic, then the other subscribers will think that the exposed thing has published a new property value. Maybe the exposed thing was offline but the broker will still notify the subscribers. In the end, all we are doing with a TD is describing the broker and not necessarily the exposed thing.

If we don't want to change the topic, we will need to identify the publisher and make sure that it is the exposed thing. This would imply a change in the broker since the brokers should not do this and the subscribers should not care who publishes (source [here](https://stackoverflow.com/a/48114255/3806426)).

I think it is necessary to change the topic (i.e. href). We do this already for observable properties in the http binding since a GET request can be for a `readproperty` or `observeproperty` and there is the additional `/observable` in the href as a result.

[...]

Issue #144, opened 2019-11-13
[question] How to get the TD with MQTT Subscribe


egekorkan:
For HTTP and CoAP, the TD is always found at `http://myIP:port/myDevice`. Where is this for the MQTT? I can think of the servient publishing its TD upon successful connection to the broker with retain set to true. A python example with the paho library from Eclipse would be:
```
def on_connect(client, userdata, flags, rc):
    client.publish("td", td, retain=True)
```

danielpeintner:
@egekorkan let's do the same as for http and coap, report td based thing `title`

egekorkan:
PR available at #162

egekorkan:
Available on master branch, closing

PR #162, opened 2020-01-17
MQTT features

egekorkan:
Added features:
- Before the `expose()` returns, the TD is published to the broker, addresses issue #144. This is also published with retain set to true
- readOnly and writeOnly properties are handled. Before even writeOnly properties were readable and there was no way to write to properties via MQTT. Now, only readable properties are published to and there is handling of subscribed topics to handle writing, like with actions. Addresses issue #152 

I am not able to write tests for this due to 2 reasons:
1. See #161 
2. A test for this would require to fetch the TD from the broker. This is not possible with the `fetch()` function since it always maps to `readResource()` which is not implemented in the mqtt binding. Then I have tried the following but was blocked by the same reason as in the point 1:
```
                            let broker = mqtt.connect("test.mosquitto.org");
                            broker.on("connect", () => {
                                broker.subscribe("TestWoTMQTT");
                                broker.on("message", (receivedTopic: string, payload: string) => {
                                    if (receivedTopic == "TestWoTMQTT"){
                                        expect(payload).contain("TestWoTMQTT")
                                        console.log(payload)
                                    }
                                });
```
However, I have tested using mosquitto_sub and mosquitto_pub and gotten the TD, even multiple times or published before subscription. Only writeable properties are written to and their form show up in the TD. Same for readable properties.

### DIFF
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -106,28 +106,41 @@ export default class MqttBrokerServer implements ProtocolServer {
         let topic = "/" + encodeURIComponent(name) + "/properties/" + encodeURIComponent(propertyName);
         let property = thing.properties[propertyName];
 
-        thing.observeProperty(propertyName,
-        // let subscription = property.subscribe(
-          (data) => {
-            let content;
-            try {
-              content = ContentSerdes.get().valueToContent(data, property.data);
-            } catch(err) {
-              console.warn(`MqttServer cannot process data for Property '${propertyName}': ${err.message}`);
-              // subscription.unsubscribe();
-              thing.unobserveProperty(propertyName);
-              return;
+        if(!property.writeOnly ){
+          thing.observeProperty(propertyName,
+          // let subscription = property.subscribe(
+            (data) => {
+              let content;
+              try {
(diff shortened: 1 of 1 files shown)


egekorkan:
I am not fully understanding the error produced by Travis but it seems that it is failing since it is taking too long to build node.js from the sources. There are the following lines in the raw log file:
```
Downloading https://nodejs.org/dist/v13.6.0/node-v13.6.0.tar.xz...
Computing checksum with sha256sum
Checksums matched!
$>./configure --prefix=/home/travis/.nvm/versions/node/v13.6.0 <
Node.js configure: Found Python 2.7.12...
[1m[93mWARNING[0m: C++ compiler (CXX=g++, 5.4.0) too old, need g++ 6.3.0 or clang++ 8.0.0
[1m[93mWARNING[0m: warnings were emitted in the configure phase
[1m[32mINFO[0m: configure completed successfully
make -C out BUILDTYPE=Release V=0

... (lots of difficult to understand lines)
The job exceeded the maximum time limit for jobs, and has been terminated.
```

danielpeintner:
1 out of 4 Travis builds failed. I guess there was another issue since the Travis build was stopped after 50 minutes while all other builds finished after around 4-6 minutes (see https://travis-ci.org/eclipse/thingweb.node-wot/builds/638673659)

I will restart the build and let's see whether the issue is resolved...

PR #178, opened 2020-02-26
(Partially) Automate generating examples from TypeScript to JavaScript

danielpeintner:
I started a *test* PR for issue https://github.com/eclipse/thingweb.node-wot/issues/171.

The idea is to have a package `examples` where we can use TypeScript to create examples and we are sure that the syntax and the API usages are correct.

To be able to run it as JavaScript with the CLI one needs to copy the code between

`// -->  BEGIN CODE for scripts `

and 

`// <--  END CODE for scripts `

to the scripts folder (e.g., https://github.com/danielpeintner/thingweb.node-wot/commit/fe19c7644453e06efa0f38f0237fdd51051deb8f) so that the CLI can properly run it.

This is not ideal but I think doing so at least there is a guarantee that the code fits the API and the syntax is correct.

Any opinions?

Note: I also wonder whether the package `demo-servients` should be combined.


egekorkan:
I would support this PR and also combining demo-servients. I don't understand however the following:
> 
> // --> BEGIN CODE for scripts
> 
> and
> 
> // <-- END CODE for scripts

danielpeintner:
The reason for begin/end is that one needs to manually copy the code between the 2 markers to scripts in https://github.com/eclipse/thingweb.node-wot/tree/master/examples/scripts (as I also did in this PR for the counter.js example).
This test also showed that we were using `thing.title` incorrectly and should rather use `thing.getThingDescription().title` instead.

The package uses TS while the scripts folder is meant to be used for CLI environment.

@egekorkan  do you have a better approach. I am all open!

danielpeintner:
The recent update simplifies the process/workflow
see https://github.com/eclipse/thingweb.node-wot/pull/178/files#diff-7f6c29f1c24e5ac1f9a491db698a6cd2

On improvements or automatic scripts can be worked on later.

danielpeintner:
I asked @sebastiankb for review.

@egekorkan  If you can check as well would be great!

I think this PR contains NOW all changes I wanted to work on.
* TS counterpart for each JS script
* half automatic generation of JS scripts
* all JS scripts are in sync now with the latest Scripting API

Remaining future topics:
* fully automate TS to JS conversion

packages/examples/README.md
```diff
@@ -0,0 +1,38 @@
...
+* `src\testthing`
+
+The idea of these folders is to use TypeScript to work on the examples which offers support in being up-to-date with the current API.
+
+see https://github.com/eclipse/thingweb.node-wot/issues/171.
+
+### Workflow
+
+1. Run `npm run build`
+2. Remove the following 3/4 lines in JS files of folder `dist/`
```
sebastiankb: I'm not sure what I have to do here? 3/4 lines (do you mean 3-4 lines)?

packages/examples/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. Copy the according JS file(s) to
+* `<node-wot>/examples/scripts`
+* `<node-wot>/examples/proxy-scripts`
+* `<node-wot>/examples/testthing`
+
+
+## Test Thing 
+
+Test thing and client is located in `src\testthing`.
```
sebastiankb: Is the test thing related to the example? or is this a generic test thing?

danielpeintner: The TestThing is developed in TypeScript in the example package. The result is copied to the examples folder so that people don't need to build/run node-wot but instead have a JavaScript file that can be used right away.

packages/examples/README.md
danielpeintner: the `tsc` command creates the 3/4 lines mentioned below which at the moment need to be removed manually. In the future we should have an automates travis script

sebastiankb:
Thanks @danielpeintner for clarification and for this PR. This will improve the quality of node-wot and its examples. Let's merge

egekorkan:
Couldn't be sure where to put this information but instead of doing this removal of 3-4 lines in Travis, I think it would be better to have a postbuild script that gets triggered by npm once the npm run build is finished. This way, it would be easier to develop new examples and not need to push them to GitHub to trigger the Travis script. However, it must be made sure that this script runs on WIndows and Linux systems, which probably means a node.js script

PR #187, opened 2020-03-11
Add support for contentTypes other than JSON …

danielpeintner:
This is work in progress and needs more discussions.

I started to report SVG with contentType `"image/svg+xml"`
Will try to add other content types like PNG etc

### DIFF
--- a/examples/scripts/counter.js
+++ b/examples/scripts/counter.js
@@ -17,7 +17,6 @@
 // * basic properties, actions, events
 // * uriVariables
 // * multi-language
-
 WoT.produce({
     title: "counter",
     titles: {
@@ -44,6 +43,14 @@ WoT.produce({
             },
             "iot:Custom": "example annotation",
             observable: true,
+            readOnly: true
+        },
+        countAsImage: {
+            description: "current counter value as SVG image",
+            forms: [{
+                    contentType: "image/svg+xml"
+                }],
+            observable: false,
             readOnly: true
         },
         lastChange: {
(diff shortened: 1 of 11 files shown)


sebastiankb:
the direction looks very good. go ahead.

egekorkan:
I think that at least in the code, there should be a comment saying that it will apply to all the forms of the interaction since it is something normally unexpected.

danielpeintner:
> I think that at least in the code, there should be a comment saying that it will apply to all the forms of the interaction since it is something normally unexpected.

@egekorkan, what do you mean by "it will apply" ?
The `contentType`? Where should a comment go to?

egekorkan:
> @egekorkan, what do you mean by "it will apply" ?

The contentType is specified only for the 0th form but the exposed TD will have the custom contentType in all form field of the interaction.

Before the WoT.produce() there are some comments, it would be nice to indicate what happens in this example and say that the contentType put into forms apply to all the forms

relu91:
It seems a nice starting point. I'm going to add just a general comment and then maybe some considerations about some specific line of codes. 

It might be better to extract the "form-fill-contentType" (let me use this term) logic outside from the specific HttpServer implementation. It is quite a general feature that could be used by many other protocol server implementations. So I think it could be worth doing a little bit of engineering and move this feature to a specialized class or function. Plus, server protocol should just be servers and concern more about protocol functionalities and less about hight level configuration stuff. As always just my 2 cents.

packages/binding-http/src/http-server.ts
```diff
@@ -529,7 +539,27 @@ export default class HttpServer implements ProtocolServer {
                   thing.readProperty(segments[3], options)
                   // property.read(options)
                     .then((value:any) => {
-                      let content = ContentSerdes.get().valueToContent(value, <any>property);
+                      // try to find contentType (How to do this better)
+                      // Should readProperty() return an encapsulated value container with value&contenType
+                      // as sketched in https://github.com/w3c/wot-scripting-api/issues/201#issuecomment-573702999
+                      let contentType;
```
relu91: That's a really good point, I'm going to follow w3c/wot-scripting-api#201 and try to contribute there. From the developer's point of view, the problem is how to know when to DataSchema make sense and when is not.

packages/binding-http/src/http-server.ts
```diff
@@ -216,6 +216,16 @@ export default class HttpServer implements ProtocolServer {
             let propertyNamePattern = this.updateInteractionNameWithUriVariablePattern(propertyName, thing.properties[propertyName].uriVariables);
             let href = base + "/" + this.PROPERTY_DIR + "/" + propertyNamePattern;
             let form = new TD.Form(href, type);
+            
+            if(tdTemplate && tdTemplate.properties[propertyName] && tdTemplate.properties[propertyName].forms) {
+              for (let formTemplate of tdTemplate.properties[propertyName].forms) {
+                // TODO match HTTP only?
```
relu91: As I said in my previous https://github.com/eclipse/thingweb.node-wot/pull/187#issuecomment-598839011 it might worth exploring the possibility to move this logic outside of this class.

packages/binding-http/src/http-server.ts
danielpeintner: Agree. This is a *broader* issue that needs to be solved in the Scripting API itself.

packages/binding-http/src/http-server.ts
danielpeintner: This is a very good point. I will extract the code in some sort of a helper class. The PR is still work in progress and I hope I get to work on it the next days! Will keep you informed!

danielpeintner:
I extracted the code to a helper class. It makes it usable in other bindings. Thanks @relu91 for pointing this out. I will look now in other interaction affordances besides properties.

relu91:
It seems perfect, later on we might add a pre-processing step where this substitution is done automatically; so that protocol-bindings implementations do not even need to call the helper function. Great job!

danielpeintner:
I removed the [WIP] label and I think it is ready for review/merge.

@sebastiankb @egekorkan @relu91 please take another look

examples/scripts/counter.js
```diff
@@ -17,7 +17,6 @@
 // * basic properties, actions, events
```
egekorkan: I would add a comment like the following:
```js
// This is an example Thing script. 
// It has a count property that can be incremented or decremented via actions  and its changes are reported via events.
// It also has two properties that return an image. The SVG property is also influenced by the increment and decrement actions.
// Features:
// * basic properties, actions, events
// * uriVariables for actions
// * multi-language for descriptions and titles
// * image contentTypes for properties. Note: the contentType applies to all forms of the property
```

packages/core/src/codecs/base64-codec.ts
```diff
@@ -0,0 +1,44 @@
+/********************************************************************************
+ * Copyright (c) 2018 - 2019 Contributors to the Eclipse Foundation
```
egekorkan: 2020 instead of 2019

It is probably needed in other files

danielpeintner: updated all copyright years affected by this PR

examples/scripts/counter.js
danielpeintner: Makes sense. Done as suggested

danielpeintner:
> I would also say that since this involves changes in the core components, a bump to version 0.7.1 is justified

In the past we usually did that once the scripting API changed.
This PR does not introduce any breaking change. Hence I propose not increasing the version but I am open to other opinions.

egekorkan:
Isn't scripting API changes result in bumping the version to 0.8?

danielpeintner:
Let's resolve the version issue later...

relu91:
I am late for the review, but it seems already great! nice work @danielpeintner 👍

PR #196, opened 2020-03-25
fix: allow explicitely configure no security in HTTP

danielpeintner:
see https://github.com/eclipse/thingweb.node-wot/issues/181

besides the fix it also changes the internal variable name "securityScheme" to "httpSecurityScheme" to make clear the difference


danielpeintner:
@egekorkan can yuo check whether it fixes *your* issue reported in https://github.com/eclipse/thingweb.node-wot/issues/181

Issue #197, opened 2020-04-01
Run node-wot in docker container


wiresio:
@hyperloris - @relu91 pointed me to: https://github.com/hyperloris/dockerfiles/blob/master/node-wot/Dockerfile which looks good to me. Would you mind submitting a PR such that we can add your code? Thanks!

hyperloris:
No problem, I'll submit a PR as soon as possible.

wiresio:
Great, thank you!

wiresio:
Was merged into master via: https://github.com/eclipse/thingweb.node-wot/pull/198

PR #198, opened 2020-04-02
feat: add Dockerfile

hyperloris:
#197

The docker image can be built in development (default) or production mode, the only difference is that the second will automatically execute the prune command.

After the build, you can easily test it by passing one of the sample scripts inside as an argument:
`docker run -p 8080:8080 IMAGE_ID /home/node/app/examples/scripts/counter.js`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git
+.vscode
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM node:12.16.1-alpine3.11
+
+RUN apk add --no-cache \
+	gcc \
+	make \
+    python2
+
+ARG BUILD_ENV=development
+
+WORKDIR /home/node/app
+
+COPY . .
+
+RUN npm install \
+    && npm run build \
+    && if [ "${BUILD_ENV}" = "production" ]; then node_modules/.bin/lerna exec "npm prune --production"; fi \
+    && npm run link
+
+EXPOSE 8080
+
+ENTRYPOINT [ "wot-servient" ]
+CMD [ "-h" ]


wiresio:
@hyperloris Thank you!

relu91:
One note: On windows, the `docker build` command may hang for a long period without responding. 
It is probably related to [this issue](https://github.com/docker/toolbox/issues/613) and low memory allocated to the docker demon. If you experience the same problem, as I did with this repository, it is useful to call `npm run clean` before building the image. 

I found that in the case you have built node-wot repo using `npm run build && npm run link` the docker client tries to send almost 10Gb as build context to the docker daemon.  This is probably due to how sim links work on windows (I actually don't know). So using `npm run clean` solves this issue.

I am leaving this here to help others in case they found themself in the same situation.

danielpeintner:
@hyperloris Thanks!

 Somehow the Eclipse ECA (Eclipse Contributor Agreement) check failed. Can you check whether you signed that.

Sometimes you also need the "Signed-off-by footer" which is required for all commits made by a contributor.

hyperloris:
@danielpeintner Done!

joshco:
@relu91 I've experienced the delay as well.  I see it often on other projects, for similar reasons. The working directory size and number of files.
I also deleted the node_modules from the main project directory, which has tons of stuff in it. 
I was wondering if the main node_modules, (or perhaps all) should be listed in the .dockerignore file so they aren't sent over.
If one is using docker, then you dont want the node modules going back and forth between the host and container. Symlinks aside, they may be different versions.

danielpeintner:
> I was wondering if the main node_modules, (or perhaps all) should be listed in the .dockerignore file so they aren't sent over.

@hyperloris @wiresio @relu91 does this make sense for you?

egekorkan:
In [this project](https://github.com/tum-esi/shadow-thing/blob/master/Dockerfile.server) we do not bother with this and haven't experienced that there is a big overhead. However, after a bit of googling, this may not be the best solution. See here: https://dev.to/alex_barashkov/using-docker-for-nodejs-in-development-and-production-3cgp

joshco:
The central point I was making is that node_modules makes sense to include in .dockerignore.  

The Dockerfile includes npm install, so it's building inside the docker container. 
````
RUN npm install \
    && npm run build \
````
Copying it over is expensive, and given potential OS differences between the host and the container, it seems like we're begging for trouble. no?

hyperloris:
It makes perfect sense @danielpeintner @joshco. I forgot to put the folder in the .dockerignore :)
However this may not solve the delay problem, because @relu91 found it even without passing node_modules in the context.

relu91:
Yeah, exactly I do agree that the node_modules should be ignored. In fact, I tried to ignore them but as @hyperloris mentioned I still had troubles. The docker build command hanged for a while and then I gave up and use the clean method above. Maybe I made some mistake in the `.dockerignore` declaration, I'll try to reproduce with the changes of #219.

danielpeintner:
Note: feel also free to provide a PR ;-)

relu91:
It still hangs for minutes. My guess is that is something wrong with symlinks in windows+docker since I built other nodejs projects without any issues. So the dirty solution still remains (in my case):
> I found that in case you have built node-wot repo using `npm run build && npm run link` the docker client tries to send almost 10Gb as build context to the docker daemon. This is probably due to how sim links work on windows (I actually don't know). So using `npm run clean` solves this issue.

danielpeintner:
PR merged!

Thanks @relu91

PR #201, opened 2020-04-06
oAuth2.0 security schema client side implementation

relu91:
# Description

This PR introduces the support for oAuth2.0 protocol as a security schema in node-wot. I decided to focus on the client-side first while leaving the server-side for another PR (if it makes sense). 

## Changes

Currently, I have implemented basic support for `client_credentials` and `password` flow. I tried to be as conservative as possible, but I had to change the method `setSecurity` as shown in the commit ff03b31. 

The idea is that during the client initialization the HTTP client bootstraps the oAuth2 protocol and obtains a valid token. After that, every consequent call uses that token. If the token expires, the client refreshes it and perform again the HTTP operation. 

Moreover, I decide to use an OAuth client library to handle the details of the flow. This adds a new dependency to the http-binding module. I evaluated the pro and cons, the library seems well supported and adopted with just two (lightweight) dependencies.  Furthermore, I added as Dev dependency an oAuth2.0 server to test the changes. The server depends on express which is a quite big package but since it used only for dev proposes it should influence t0o much the production package of node-wot.

## Future work discussion
I am planning to add also the `code` flow but it is more challenging with respect to the other two. I have pinpointed three use cases for node-wot as a oAuth2.0 client:
1. node-wot used as a client in a web browser
2. node-wot used as a native client (ie. Node application) in a desktop environment
3. node-wot used as a client in a device or in a remote host

The `code` flow can be applied for use cases 1 and 2 while for the use case 3 IMHO `client_credentials` and `password` flows fit better. However, `code` flow requires user interaction with a web page, how can we handle such scenario inside client library? 

The naive solution is to display the "oAuth authorization" page automatically every time node-wot calls a Property/Action/Event with oAuth2.0 schema and which is not already initialization with a valid token. However, this might influence negatively the user experience and it makes quite difficult to create web applications with node-wot. On the other hand, the current implementation requests the token once for every TD, which is less invasive than the naive solution. Still, it could pause the execution of a script for a very long time (i.e. until the user authorizes the application). 

I think that a more consistent method should be developed to address this kind of flow.. which it might require a modification of Scripting API (i.e. create an Event for authentication requests (?) )

# TODOs:

- [x] Automatically refresh an expired token.
    - [x] refactor: introduce a single function to make an HTTP request so that it is easier to call again the operation after the refresh token operation is completed.
- [x] Test refresh token
- [ ] Handle multiple security schemes (?)
- [ ] Implement `code` flow
    - [ ] ??


danielpeintner:
@relu91 may I ask you to resolve the conflict with cert/key files that popped up because https://github.com/eclipse/thingweb.node-wot/pull/203 was merged

Question: do you plan to work on the TODOs in this PR?
Please let us know once you think the PR is ready for review...

relu91:
Thank you for point that out! Sure I am gonna get rid of them soon. 

> Question: do you plan to work on the TODOs in this PR?

yes, for sure I will refine the implementation of `client_credentials` and `password` flow. However, I have still some doubts about the code flow, as in the last internal call, we didn't really achieve a consensus about it. Probably, we might need to open the discussion to the WG. 

As soon I have fished to test the refresh token functionality, the PR should be ready for a first review.

Meanwhile, what do you think about this:
>Moreover, I decide to use an OAuth client library to handle the details of the flow. This adds a new dependency to the http-binding module. I evaluated the pro and cons, the library seems well supported and adopted with just two (lightweight) dependencies. Furthermore, I added as Dev dependency an oAuth2.0 server to test the changes. The server depends on express which is a quite big package but since it used only for dev proposes it should influence t0o much the production package of node-wot.

Are you ok about these two new dependencies?

danielpeintner:
> Are you ok about these two new dependencies?

I am OK with it. I even think it is better to use a promising library instead of trying to implement it ourselves.

@egekorkan @sebastiankb any other opinion?

egekorkan:
Thank you @relu91 ! I am ok with the dependencies. Some other things that caught my eye:
- Could you document somewhere, at least in the examples, which flows are supported? Currently, a user will not know which he/she can use. 
- Just a question but is it fine for the server.key and server.cert to be published to npm? Shouldn't they be in examples or test?

relu91:
>Could you document somewhere, at least in the examples, which flows are supported? Currently, a user will not know which he/she can use.

Sure! I guess the best place is binding-http README.md file. 

> Just a question but is it fine for the server.key and server.cert to be published to npm? Shouldn't they be in examples or tests?

You're right, it makes no sense to have `server.key` and `server.cert` inside the root folder of the HTTP package. They are meant only for testing purposes; I'll move them inside the test folder.

egekorkan:
> Sure! I guess the best place is binding-http README.md file.

I totally agree :) 

The keys can be also useful for the people experimenting with it. Can you mention in the documentation that they are in the test folder?

relu91:
The PR is ready for review. A couple of final comments about the implementation and future work:
- the example provided requires that node-wot uses the form provided in the produce dictionary. So in the future, I might review it to make it work at 100%.  
- the code flow is delayed until there is more consensus about how to handle this scenario
- the flows supported now are: `client credential` and `password`
- Server-side implementation could use external services for token validation and user management. I still need to research more about this topic (I.e. I have yet to find specifications about the interaction between the resource server and the authorization server) 
- This PR adds the following dependencies:
    * client-oauth2
    * node-fetch

examples/security/oauth/exposer.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    "security": [
+        "oauth2_sc"
+    ],
+    "actions": {
+        "sayOk": {
+            "forms": [
+                {
+                    "href": "http://localhost:3000/resource",
+                    "htv:methodName": "GET"
```
egekorkan, line 40: Is this supported in node-wot? So does the HTTP binding start listening to GET for actions?

relu91: See this comment above:
> the example provided requires that node-wot uses the form provided in the produce dictionary. So in the future, I might review it to make it work at 100%.

Since we don't have yet a server-side implementation of oAuth the point of the example was to use the mockup oAuth server as `invokeaction` operation. So right now the example does not validate the token. I guess when we can override href inside forms using the *produce dictionary* the problem should be solved. What do you think?

relu91:
Thank you, Ege, for your feedback! I'll answer each point below:
>We started using TS for the examples so that we can track when the API changes. Could you adapt to it?

I wasn't sure to follow the provided workflow. My doubts were that the example is not only just a script but it has some convenient npm commands and also the mockup oAuth server. I guess it deserves a separate folder to not pollute the example package with other stuff. What do you think?

>I think even examples should have the license comment block in the beginning. You have it only in exposer.js

Sorry I did not pay attention to this! I'll add it ASAP. 

> I do know that it is not your addition, but shouldn't "wot-typescript-definitions": "0.7.1-SNAPSHOT.1" be a devDependency?

Yes, I agree.

> I couldn't understand the need for the node-fetch dependency. Couldn't it be done with the built-in http? From what I see, it is a convenience feature in some of the cases here but I can imagine that it would be messy with the native http. I lack the experience in this but it also changes the readResource implementation

Mostly is because I needed to have a compact function to call again when the token is expired. I could implement it using the http module but its event-based interface made it a little bit less linear. Plus node-fetch is really lightweight and it is promise-based. Plus it has a nice symmetric usage also in browsers. If you have some more tests and it breaks something; please tell me. Anyway, if it is a problem I could spend some time to revert back to http. 

> Could you add explanation for the memory model, it is something that does not exist in other bindings and I couldn't understand it directly

The memory model is only used for testing proposes. The oAuth2.0 server implementation used in the test requires it as an interface to handle users/passwords/credentials in general. See more [here](https://oauth2-server.readthedocs.io/en/latest/model/overview.html)

> I couldn't understand the need for async await in the consumed-thing.ts

It is because of this commit ff03b31. Basically, I changed the method `setSecurity` to `initSecurity` which is asynchronous. Consequently, every other caller method needed to be changed, I could have used the `then` primitive, but, as you may have understood, I am a fan of `async/await`. Now, that I look closely, I can remove the async keyword inside the promise and make the entire method async. What do you think?

I'll change back the PR to draft and apply some of your suggestions! Meanwhile, other comments are always welcomed.

egekorkan:
> I wasn't sure to follow the provided workflow. My doubts were that the example is not only just a script but it has some convenient npm commands and also the mockup oAuth server. I guess it deserves a separate folder to not pollute the example package with other stuff. What do you think?

I think it is OK to use another folder but we simply don't want to have any JS code that seems to compile but actually never works when run after 1-2 years.

Regarding node-fetch: I also find it neater but it is question of tracking changes in another library. So it is a maintenance question. @danielpeintner what is your opinion?

> The memory model is only used for testing proposes. The oAuth2.0 server implementation used in the test requires it as an interface to handle users/passwords/credentials in general. See more here

I get it now :) Could you add this link as a comment?

> It is because of this commit ff03b31. Basically, I changed the method setSecurity to initSecurity which is asynchronous. Consequently, every other caller method needed to be changed, I could have used the then primitive, but, as you may have understood, I am a fan of async/await. Now, that I look closely, I can remove the async keyword inside the promise and make the entire method async. What do you think?

I am also a fan of async/await but since I don't trust the tests (not yours but in general tests in nodewot) I always fear that it will break something somewhere. I see that it is due to that commit but why the change to initSecurity and also making it async was needed in the first place? I guess I am missing something :worried:

examples/security/oauth/exposer.js
egekorkan: I see, sorry for missing the comment. So does this exposed Thing contain a form that doesn't belong to itself? I am pretty sure that node-wot would replace with its own.

danielpeintner:
> Regarding node-fetch: I also find it neater but it is question of tracking changes in another library. So it is a maintenance question. @danielpeintner what is your opinion?

To be honest. No strong opinion either way...

examples/security/oauth/exposer.js
relu91: Exactly, I imagined that in the future TDs could contain forms that are not served from the exposing servient. Is that so? otherwise, I'll be happy to change the example.

relu91:
> I think it is OK to use another folder but we simply don't want to have any JS code that seems to compile but actually never works when run after 1-2 years.

Great I'll refactor the example 👍 , at this point I would also try to get rid of the `"href": "http://localhost:3000/resource"` so that the example would be fully operational.

>I get it now :) Could you add this link as a comment?

sure!

>I am also a fan of async/await but since I don't trust the tests (not yours but in general tests in nodewot) I always fear that it will break something somewhere. I see that it is due to that commit but why the change to initSecurity and also making it async was needed in the first place? I guess I am missing something

See in the PR body:
>The idea is that during the client initialization the HTTP client bootstraps the oAuth2 protocol and obtains a valid token. After that, every consequent call uses that token. If the token expires, the client refreshes it and perform again the HTTP operation.

The point is that oAuth needs asynchronous calls to initialize, so every other call in the chain needed to be changed to asynchronous. I could do a lazy initialization, which means that the protocol is not started until the affordance is called. 
- Pro: no changes to the API
- Cons: less clean error handling (?) 

What do you think?

About `node-fetch` there is some discussion on the [Scripting API repo](https://github.com/w3c/wot-scripting-api/issues/201#issuecomment-623914185), we could kill two birds with one stone 😄 .

egekorkan:
> The point is that oAuth needs asynchronous calls to initialize, so every other call in the chain needed to be changed to asynchronous. I could do a lazy initialization, which means that the protocol is not started until the affordance is called.

Now I get everything! Thanks for the clarification :blush: 

> About node-fetch there is some discussion on the Scripting API repo, we could kill two birds with one stone smile . 

This is very true :smile:  Then let's go!

zolkis:
> About node-fetch there is some discussion on the Scripting API repo, we could kill two birds with one stone. 

Flies. We kill flies. :) 
But first we fetch with node-fetch :). I like this approach.
If based on this experience you have any suggestions on how could we explore this synergy in the Scripting API, the doors are open!
I'd like to explore a neat subset of Response/Body on how we expose interaction data.

relu91:
> Flies. We kill flies. :)

for sure is a less cruel image 🤣 here we don't even use kill, we say *catch two birds..*.

 > If based on this experience you have any suggestions on how could we explore this synergy in the Scripting API, the doors are open!
I'd like to explore a neat subset of Response/Body on how we expose interaction data.

I'd be really glad to join the discussion! I just need some more time to keep up with the current specification described in w3c/wot-scripting-api#201 and in relative PR. I think that node-fetch is a very elegant solution, but I still not have the whole picture clear.

relu91:
Going back to draft mode, using node-fetch seems to break the browserified package. [This](https://github.com/matthew-andrews/isomorphic-fetch) might be the solution to the problem. I'll check it tomorrow. Sorry again for the delay.

danielpeintner:
Thanks @relu91 !

Who is suited to review this rather *big* PR @egekorkan @sebastiankb ?

There are some changes in core (see `setSecurity` --> `initSecurity` with promise in `protocol-interfaces.ts`)

relu91:
Yes, unfortunately, the PR is quite big... I suggest using the commit messages for guidance they are quite explicative.  I don't know about who is the most appropriate to review; @sebastiankb asked for this feature in the first place and @egekorkan had already done some review work.

danielpeintner:
One aspect is the code size, but what makes it even more *dangerous* is the change in a core API (I understand though the need).

At least it uses a "different" name which guides TS users to the difference but I guess if people solely use JS it will break at runtime :-(

relu91:
>  but what makes it even more dangerous is the change in a core API (I understand though the need).

What about the change below? It will allow maintaining the same old API but it will defer the actual initialization of the protocol. Another concern that I have is that it might be also difficult to handle lazy initialization if we will choose to provide an API for security as @zolkis hinted in https://github.com/w3c/wot-scripting-api/issues/214#issuecomment-624131268. 

> I could do a lazy initialization, which means that the protocol is not started until the affordance is called.

relu91:
To increase the probability of the acceptance of this PR, I reverted back to the previous interface for `setSecurity`. Now the oAuth2.0 is deferred until the actual accordance is called.

packages/examples/security/oauth/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+in a different terminal 
+
+```bash
+# start the exposer
+npm run start:exposer
+```
+Finally, in other terminal
+```bash
+npm run start:consumer
```
danielpeintner, line 26: I get the following

```
...
ConsumedThing 'OAuth' setting credentials for [HttpClient]
Servient looking up credentials for 'urn:dev:wot:oauth:test'
HttpClient using security scheme 'oauth2'
ConsumedThing 'OAuth' invoking https://localhost:3000/resource
LOCALHOST FIX
HttpClient (invokeResource) sending GET  to https://127.0.0.1:3000/resource
(node:38356) UnhandledPromiseRejectionWarning: FetchError: request to https://127.0.0.1:3000/resource failed, reason: certificate has expired
    at ClientRequest.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot\packages\binding-http\node_modules\node-fetch\lib\index.js:1455:11)
```

packages/browser-bundle/index.js
```diff
@@ -2,7 +2,7 @@
 
 var Wot = {};
 Wot.Core = require("@node-wot/core");
-Wot.Http = require("@node-wot/binding-http");
+Wot.Http = require("@node-wot/binding-http/dist/http-browser");
```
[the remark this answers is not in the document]
relu91: Sorry I was not sure if browserify imported the correct module. Now, I checked and it uses the file specified in the `browser` filed of the `package.json`.

packages/examples/security/oauth/README.md
danielpeintner: The issue is that 

```
    "http":{
        "allowSelfSigned": true
    },
```

is missing in
examples\security\oauth\wot-client-servient-conf.json
different to 
packages\examples\security\oauth\wot-client-servient-conf.json

With this addition it works in both places.

May I ask you to add it to the PR...

relu91: Nice catch! solved 👍

Issue #181, opened 2020-02-27
[binding-http] Cannot explicitely configure no security in HTTP


egekorkan:
The http configuration part of the configuration file cannot accept `nosec` for the security scheme, i.e.:
```
        "http": {
            "port": 8080,
            "security": {
                "scheme": "nosec"
            }
        }
```
results in the error message `HttpServer does not allow security without TLS (HTTPS)`
This is happening due to the logic in [this line](https://github.com/eclipse/thingweb.node-wot/blob/a2f1bed088804cdc388d3db73625f06d2efbbbed/packages/binding-http/src/http-server.ts#L79).

I think that it should be possible to specify no security explicitely in the configuration file. At the moment, one has to forget the scheme to specify no security whereas TD specification is not following this logic.

fillobotto:
Moreover I'm not able to use any other security scheme expect for Basic auth.
I'm trying to use Bearer auth but the serialized TD overwrites both `security `and `securityDefinitions ` with [basic auth](https://github.com/eclipse/thingweb.node-wot/blob/f6a814b0231951ded560ba3abb13c818b8a99fff/packages/binding-http/src/http-server.ts#L259)

danielpeintner:
@fillobotto May I ask you to provide a simple test in [http-server-test.ts](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-http/test/http-server-test.ts) so that
1. We know your input
2. We can fix the issue and have a test for it

If you don't feel comfortable enough please provide some pseudo code to show what you do and what your input looks like.

@egekorkan may I ask you the same. A simple test...

Thanks!

egekorkan:
In order to reproduce,, use the following wot-servient.conf.json with the counter.js script:
```json
{
    "http": {
        "port": 8080,
        "security": {
            "scheme": "nosec"
        }
    }
}
```

egekorkan:
The following added to http-server.ts causes the test to fail:
```js
  @test async "should start and stop a server with no security"() {
    let httpServer = new HttpServer({ port: 58080, security: {scheme: "nosec"}}); //failing also with NoSec

    await httpServer.start(null);
    expect(httpServer.getSecurityScheme()).to.eq("NoSec"); // the actual test we want
    await httpServer.stop();
  }
```
which requires a function to get the security scheme in the http-server.ts

```js
  public getSecurityScheme(): Object {
    return this.securityScheme;
  }
```

[...]

PR #203, opened 2020-04-14
[binding-http] Fix security schema generation for https servers

relu91:
As exposed by @fillobotto [here](https://github.com/eclipse/thingweb.node-wot/issues/181#issuecomment-603192434) (issue #181), the https server was ignoring every security definition of an exposed thing and it overrode it with a basic security scheme.

This PR solves the issue by choosing one security definition of the exposed thing and checking if it is supported by the server. if so the exposed TD will use that definition. Otherwise, if no schema is supported it throws an error since the server cannot validate requests for the requested security schema. 

Finally, if no security schema is defined the server fills the gap with a default `nosec` schema.


danielpeintner:
Looks good from my side.

PR #215, opened 2020-04-26
feat: add new example script

fatadel:

### DIFF
--- a/examples/scripts/coffee-machine-client.js
+++ b/examples/scripts/coffee-machine-client.js
@@ -1,0 +1,75 @@
+/********************************************************************************
+ * Copyright (c) 2018 - 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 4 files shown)


danielpeintner:
@fatadel Please create the example in TypeScript (see https://github.com/eclipse/thingweb.node-wot/tree/master/packages/examples). Doing so allows us to easily detect if a an example gets out-of-sync with the scripting API, provides type feedback,  et cetera ...

Later you can copy it to https://github.com/eclipse/thingweb.node-wot/tree/master/examples as described in https://github.com/eclipse/thingweb.node-wot/tree/master/packages/examples#workflow

fatadel:
@danielpeintner But that's exactly what I did.

danielpeintner:
> @danielpeintner But that's exactly what I did.

OK great. I might have missed the ts file. Sorry!

sebastiankb:
many thanks for this new Example and Tutorial

Issue #214, opened 2020-04-23
HTTP PORT and BASE_URL deployment issues


joshco:
> FYI The wot library has a lot of depth to it, it was neat to see the event subscription using the plain http binding just work out of the box.  It's really more than a reference implementation.  kudos.

I wrote an example exposed thing that is hosted in a NodeJS application, using express for the browser UI, and wot for the exposed thing, which is controlled by browserified node-wot.  When deploying it in the cloud, I ran in to issues with HTTP PORT mapping and the BASE URL for the TD URLs, which required changes to the http-binding to get it working.  

If there's consensus on how to address these, I'm happy to create a PR.

# HTTP PORT
With the thing embedded in NodsJS, there's two http listeners i a single process.  This all worked fine on my development system, 3000 (express) and 8081 (wot)

In the cloud, I use Dokku, which is a docker based "heroku in a box" that you can run on your own ec2 instance.  It is heroku compatible and uses the heroku build-packs.   https://github.com/dokku/dokku

My first deploy got me this:
````
    > lightbar@0.0.0 start /app
       > node ./bin/www
       HttpServer starting on port 8081
       Port 5000 is already in use
       npm ERR! code ELIFECYCLE
       npm ERR! errno 1
````
(The full heroku buildpack output is here:  https://gist.github.com/joshco/e5fbf2b7167504fbc66b6ffc83b447cc)

Port 5000 is the port the build-pack sets in an environment variable that the application will use.  (this is the same as the rails and other build-packs)  When I did a curl to port 5000, i got back the wot servient response, where I expected the Node Express app to be.

8081 is the port I specified in the WOT http config, and where I expected the WOT to be.  When I did a curl to 8081, it wouldn't connect.

After troubleshooting, it turns out that WOT is fooling us. In the start method of http-server, it's logging that it is going to use the port specified in the http config (8081), but later in the code, in line 121 it gives preference to the PORT env variable if it is present.  So it's actually bound to 5000, wining the race, leaving express to get the error.

To make my project work, I changed the code from

`(+process.env.PORT || this.port, this.address);`
To
`+process.env.WOT_PORT || +process.env.PORT || _this.port;`

In http-server.ts
https://github.com/eclipse/thingweb.node-wot/blob/b9d0a3d85448d2d0652dcd3cfe5980ad5c12fd39/packages/binding-http/src/http-server.ts#L100-L123

[...]

PR #220, opened 2020-05-02
BaseURI for links in TD

joshco:
Related issues #214,  #878

TLDR:
> Implement 'base' property, and test, via
````
Assume thing.base = "http://all.your.base.are.belong.to.us:9999"

urlbase = thing.base || ( this.scheme + "://" + address + ":" + this.getPort() );
let href = base + "/" + this.ALL_DIR + "/" + encodeURIComponent(this.ALL_PROPERTIES);
console.log(href)
=> ":"http://all.your.base.are.belong.to.us:9999/WOT/all/properties" 
````

According to the language in the spec, the implied assumption is that the form, action, property URLs are relative, and relative to the base URI.

> base property: Define the base URI that is used for all relative URI references throughout a TD document. In TD instances, all relative URIs are resolved relative to the base URI using the algorithm defined in [RFC3986].
> base does not affect the URIs used in @context and the IRIs used within Linked Data [LINKED-DATA] graphs that are relevant when semantic processing is applied to TD instances.

However, node-wot has absolute URLs, and those are based on the network interfaces of the machine. See my comment https://github.com/eclipse/thingweb.node-wot/issues/214#issuecomment-621699899

This type of problem occurred in HAL as well.  In link relations, the implementations would use absolute URLs.  The challenge is that any implementations like this will probably barf if they start getting relative URLs.  They were never checking for the base, neither is node-wot. Right now, node-wot seems to be ignoring base.

This is a proposal for discussion.  In this PR, the http server expose function will check if there is a thing.base property.  If so, it will use that instead of the scheme, address and port in the HTTP config.
This keeps the links in forms, properties, actions as full URLs; it just does the "relative mapping" *squint* on the service side.  

It also behaves like a canonical URL prefix, that may be set on heroku, dokku, ec2, lambda or any front end server that is independent from node-wot.

### DIFF
--- a/packages/binding-http/README.md
+++ b/packages/binding-http/README.md
@@ -149,13 +149,14 @@ servient.start().then((WoT) => {
 The protocol binding can be configured using his constructor or trough servient config file. The `HTTPConfig` object contains a set of useful parameters: 
 ```ts
 {
-    port?: number;                  // HTTP server port
-    address?: string;               // HTTP server address
+    port?: number;                  // TCP Port to listen on
+    address?: string;               // IP address or hostname of local interface to bind to
     proxy?: HttpProxyConfig;        // proxy configuration
     allowSelfSigned?: boolean;      // Accept self signed certificates
     serverKey?: string;             // HTTPs server secret key file
     serverCert?: string;            // HTTPs server certificate file
     security?: TD.SecurityScheme;   // Security scheme of the server
+    baseUri?: string                // A Base URI to be used in the TD in cases where the client will access a different URL than the actual machine serving the thing.  [See Using BaseUri below]
 }
 ```
 When both `serverKey` and `serverCert` are defined the server is started in `https` mode. Examples of `serverKey` and `serverCert` can be found [here](../../examples/security). Moreover, when a security schema is provided the servient must be also configured with valid credentials both client and server side. See [Security](#Security) for further details.
@@ -224,7 +225,54 @@ The above configuration file, is setting up a https server with basic secure sch
 ### oAuth2.0
 Currently this binding supports only oAuth2.0 `client credential` and `Resource owner credential` flows. Other flows may be implemented in future like `code` flow. Futhermore, the oAuth2.0 protocol is only implemented for the client side.
 
-An example of a WoT oAuth2.0 enabled client can be found [here](../examples/security/oauth).
+An example of a WoT oAuth2.0 enabled client can be found [here](../examples/src/security/oauth/README.md).
+
+
+### Using baseUri
+
+Assume the example [WoT coffee machine](../examples/src/scripts/coffee-machine.ts) is in the W3C's office kitchen connected to the W3C's private network.  It allows you to start the coffee machine before you leave home so it will be ready when you get to work.
+
+Inside the W3C's private network the coffee machine can found at:
+<br/>`https://internal-host:8080/smart-coffee-machine`
+
+From your home, it can be addressed via an Internet accessible domain name: 
+<br/>`https://coffee.w3.org/things/smart-coffee-machine`
+
+
+
+__HttpServer Configuration__
+
+```js
+servient.addServer(new HttpServer({
+    port: 8080, // (default 8080)
+    baseUri: 'https://coffee.w3.org/things'
+}));
+```
+
+__External Gateway Configuration__
+
+The DNS name`coffee.w3.org` resolves to an elastic IP on a gateway using nginx, which has this rule configured. 
+
+```
+location /things/ {
+    proxy_pass https://internal-host:8080/smart-coffee-machine
+}
+```
+
+The exposed thing on the internal server will product form URLs such as:
+
+```json 
+ "actions": {
+    "makeDrink": {
+      "forms": [
+        {
+          "href": "https://wot.w3.org/things/smart-coffee-machine/actions/makeDrink"
+```
+__baseUrt vs address__
+
+> `baseUri` tells the producer to prefix URLs which may include hostnames, network interfaces, and URI prefixes which are not local to the machine exposing the Thing.
+> `address` tells the HttpServer a specific ocal network interface to bind its TCP listener. 
+
 
 ## Feature matrix
 
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -48,6 +48,7 @@ export default class HttpServer implements ProtocolServer {
 
   private readonly port: number = 8080;
   private readonly address: string = undefined;
+  private readonly baseUri: string = undefined;
   private readonly httpSecurityScheme: string = "NoSec"; // HTTP header compatible string
   private readonly validOAuthClients: RegExp = /.*/g; 
   private readonly server: http.Server | https.Server = null;
@@ -75,6 +76,9 @@ export default class HttpServer implements ProtocolServer {
 
     if (config.address !== undefined) {
       this.address = config.address;
+    }
+    if (config.baseUri !== undefined) {
+      this.baseUri = config.baseUri;
     }
 
     // TLS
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by relu91** [comment]
Hi @joshco it is great that you started working on this problem. Your solution seems acceptable but it might be also a temporary "workaround". I would like to have a more consistent approach. IMHO there are three steps to achieve "real" *base URL* handling:
1. Handle href properties in the initialization dictionary that are passed to the `produce` method.
2. Implement the client-side logic wich resolves base + href
3. Choose which form to use (is still under discussion on the scripting API)

Note also that this logic should be shared with all the other protocol binding implementations. So it might be better to have the resolution step implemented in the `core` package rather than inside a particular binding package. 

Having implemented 1 and 2 allows more complex design decisions, like in our setup when we still want to have intra-docker containers interaction but also like to expose some containers to the web.

What do you think?

**Comment by joshco** [comment]
@relu91 I agree that this is a temporary fix.  I'm wondering if this temporary fix should use an attribute with an extension prefix specific to node-wot.

> 2. Implement the client-side logic wich resolves base + href

Depending on how much adoption already exists with the current functionality of node-wot, using the standard base attribute for the temporary solution could create pain down the road.  Albeit small, that could be a friction for new implementers.

Assume A: that there are other clients that are currently paying attention to the base attribute. 
They don't see base from node-wot producer, so they must be falling back to using what's in the form link. Or maybe their code is just magically working because concatenating '' with 'http://full.url/WOT' will arrive at a useful URL

Assume B: that there are also clients who are n't bother to look. and working with node-wot producer because the URLs are absolute.

One day, when node-wot starts using the base attribute, what will category B do?  Will they blindly try to connect to an absolute URL and fail?

The group probably has a better idea of the current adoption, so this may or may not be an issue.  

What's your take?

> 3. Choose which form to use (is still under discussion on the scripting API)

Is this a discussion I could participate in as a guest?

Also, are there any group meetings for node-wot?  Seems like these implementation issues could be dealt with more efficiently as an interactive discussion.  (I've got a zoom account that could be used if helpful)

For the intra-container scenario, since someone did the work to make node-wot actually enumerating the interfaces, that could be put to use.  Since it knows what interface it is using, it could choose to only expose the URLs for that interface rather than expose all URLs on every interface.

Frankly, what's the purpose of returning all URLs, some of which an external client could never even access?

File: packages/binding-http/src/http-server.ts [hunk 1]
```diff
@@ -187,7 +187,8 @@ export default class HttpServer implements ProtocolServer {
       // fill in binding data
       for (let address of Helpers.getAddresses()) {
         for (let type of ContentSerdes.get().getOfferedMediaTypes()) {
-          let base: string = this.scheme + "://" + address + ":" + this.getPort() + "/" + encodeURIComponent(title);
+          let baseURI: string = thing.base || ( this.scheme + "://" + address + ":" + this.getPort() );
```
danielpeintner, line 190: What happens if both are present?

packages/binding-http/test/http-server-test.ts
```diff
@@ -1,15 +1,15 @@
 /********************************************************************************
  * Copyright (c) 2018 - 2020 Contributors to the Eclipse Foundation
- * 
+ *
```
danielpeintner, line 3: there are many whitespace diffs. do you think we can remove them again.

danielpeintner:
Is "staticAddress" not an option?

see wot-servient.conf.json syntax:
```
{
    "servient": {
        "clientOnly": CLIENTONLY,
        "staticAddress": STATIC,
        "scriptAction": RUNSCRIPT
    },
```

packages/binding-http/src/http-server.ts
joshco: > What happens if both are present?
What do you mean by both?  it either uses the base attribute (as a URL prefix), or generates based on the network interface address it is using in the for loop.  The network interface address is always present.  It is only overridden if base attribute is set.
Am I missing your question?

> What happens with different bindings (CoAP, HTTP, ..)
Good question.  This is my first exposure to CoAP, I'll look into that.

packages/binding-http/test/http-server-test.ts
joshco: Whoops, my bad.  I'll remove them.

packages/binding-http/src/http-server.ts
danielpeintner: Thanks!

packages/binding-http/test/http-server-test.ts
```diff
@@ -170,4 +170,46 @@ class HttpServerTest {
     await httpServer.stop();
 
   }
+
```
joshco, line 173: @danielpeintner FYI, I removed the whitespace diffs and squashed.

joshco:
@danielpeintner "What happens with different bindings (CoAP, HTTP, ..)?"

I read up on CoAP, and  played with the the Ruby coap gem/cli.  I'm new to CoAP, but here are my observations.  Please let me know if I'm off base.

CoAP can be loosely viewed as a tokenized HTTP over UDP, which can optionally be secured by DTLS, which can be loosely viewed as TLS over UDP.  The CoAP connection endpoints are URLs as well, either coap://coap.me/hello or coaps://coap.me/openSafeThing.  

@danielpeintner "Is `staticAddress` not an option?"

Are they the same thing?

The static staticAddress is defined by [/packages/cli/wot-servient.conf.json.md](https://github.com/eclipse/thingweb.node-wot/blob/2ff2bd1f817bcbc19420356397933d7e8dee4ba9/packages/cli/wot-servient.conf.json.md) defines staticAddress as:

> A string indicating a static hostname / IP address

[Thing Definition](https://www.w3.org/TR/wot-thing-description/#thing) declares base as:

> Define the base URI that is used for all relative URI references throughout a TD document. In TD instances, all relative URIs are resolved relative to the base URI using the algorithm defined in [RFC3986].


A simple use of this could be just a difference in the hostname advertised to the client.
 Instead of local hostnames/IPs like  `ip-172-20-0-121.ec2.internal` or `thingpatrol.wot.dockernet`,
The producer might try setting addresss to a publicly accessible FDQN of `thingpatrol.joshco.org`, `thingpatrol.whitehouse.gov` or `thingpatrol.sun.com`  The client would then navigate to `http://thingpatrol.sun.com/WOT` 

Attempting to use the static address attribute for this purpose will result in a socket error as node-wot tries to bind a listen socket on a non-local host/IP address.

[WoT Arch](https://w3c.github.io/wot-architecture/#hypermedia-driven) Sec 6.6.1 states:

> Protocol Bindings MUST be serialized as hypermedia controls (see § 6.5 Hypermedia Controls)

__WoT Arch Sec 6.6.2__ states:

> Hypermedia controls rely on URIs [RFC3986] to identify link and submission targets.

So is it fair to assert that regardless of protocol binding, apps should be routing based on URIs independent of lower level aspects like port, ip address, or the physical network interfaces?  

Beyond concatenation, and standardized conventions for query parameter name/values (UriVariables), can URLs be treated as opaque outside of the protocol binding libraries?


__WoT Arch Sec 6.7.2 "Indirect Communication" states:__
  
> An Intermediary is required if the Servients use different protocols or if they are on different networks that require authentication and provide access control (e.g. firewalls).

When talking to the cat-food dispenser in our home, a LAN/Wifi attached consumer can speak with different security, QOS or other parameters than one outside our home.  Therefore, it could plausibly have a different scheme (http,https,coap(s),ws), DNS name, listener port, an addition a path, or all of the above. 

Even setting aside the non-local interface issue, the address attribute only handles hostname/IP, not the rest of the parts.
What will happen if you set address to a URL and execute line 65:

https://github.com/eclipse/thingweb.node-wot/blob/251dfd3e40e0591876a1fab80b6aedf4f68ea261/packages/binding-coap/src/coap-server.ts#L65

This comment is already long enough, and I'm new here, so I'll stop and pass the mic.  What do you think?

Also,on the namespaced attribute question, do you think this is a situation where we should do that, or just stick with "base"?

egekorkan:
I am not able to answer all the questions but the static address is given to the servient first, which then generates Things with the same static address. This was a bit the reason behind my comment [here](https://github.com/eclipse/thingweb.node-wot/issues/214#issuecomment-621679897).

joshco:
@egekorkan Ok, I will research that distinction. and report back

relu91:
>Assume A: that there are other clients that are currently paying attention to the base attribute.
They don't see base from node-wot producer, so they must be falling back to using what's in the form link. Or maybe their code is just magically working because concatenating '' with 'http://full.url/WOT' will arrive at a useful URL

>Assume B: that there are also clients who are n't bother to look. and working with node-wot producer because the URLs are absolute.

>One day, when node-wot starts using the base attribute, what will category B do? Will they blindly try to connect to an absolute URL and fail?

>The group probably has a better idea of the current adoption, so this may or may not be an issue.

>What's your take?

Sorry for the late response.  My idea is that group B is not following the specs therefore they have a bug in the affordance URI resolution. It might be a too strong opinion, but, frankly, I guess it is too early to care about backward compatibility. 

>For the intra-container scenario, since someone did the work to make node-wot actually enumerating the interfaces, that could be put to use. Since it knows what interface it is using, it could choose to only expose the URLs for that interface rather than expose all URLs on every interface.

>Frankly, what's the purpose of returning all URLs, some of which an external client could never even access?

Actually, we did some tests and the best option for intra-container communication is to leverage on docker swarm name resolution. Therefore, maybe specifying the base address or using `addressStatic` could solve some troubles that we encountered. 

My guess for the address enumeration feature is that it is a reasonable default behavior if no other configuration parameter is given. Unfortunately, I do not know which was the rationale back then, maybe @danielpeintner has more insights (?)

joshco:
OK, so no need for the namespace prefix

> for intra-container communication is to leverage on docker swarm name resolution. Therefore, specifying the base address or using addressStatic could solve some troubles that we encountered.

Ok, so the intra-docker issue is hostname rather than protocol

joshco:
@egekorkan @danielpeintner I've done some more research and here's where I'm at.

egekorkan wrote:
> no base is given but static address is there. This is the current case and all the TDs of the servient will have the static address

I wrote this test:

https://github.com/joshco/thingweb.node-wot/blob/c87dee4901dede8d17d169c4c6e2490fb724a14e/packages/binding-http/test/http-server-test.ts#L174-L222

What the test demonstrates is that address cannot be used when the base links in the TD need to have an external hostname, eg a non-local IP address which does not correspond to a local network interface.

__Scenarios__
Imagine an w3c employee with a smart coffee machine on his desk.  It is connected to the w3c corporate network.  When then roll out of bed in the morning, they want to use the wot client on their mobile phone to start brewing the coffee before they get to the office.  

The same scenario could be that the coffee machine is at home and they want to start it before they get home and connect to the home network.

Their internet connected client is going to need to talk to a proxy gateway to the corporate network or their home.  The coffee machine automatically creates a tunnel to the gateway from the other side on ACME Coffee Machine Inc's cloud. (like my wot_powerpoint)

The TD that the mobile client consumes will need the base of the URLs in the TD to be either

* https://theirthing.things.w3.org/WOT
* https://theirmachine.acmecoffee.com/WOT

The Address property cannot be used because the coffee machine cannot bind to the hostname/IP in either of the above URLs.

__My recommendations__

1. Address should be used only to determine the local interface to bind the listen socket.
2. base uri should be the only thing used to compose the URLs in the TD (relative or absolute)

Another question
Why bother having the TD provide all links to all interfaces when the client will only be able to connect to one but not all hostname/IPs?  If the producer limited the link hostnames to only the interface it is bound to, or the external base URI, that would simplify the client since it doesn't need to pick one of many and try to find out it's own IP address to determine if it has a route to the interface.  That's another mingling of layers.

For inter-container communication, the TD accessible from the internal docker network would only give URLs to the interface on the internal docker network, which makes intra-docker communication simpler.

My Recommendation is either:
a. The producer only provides links on the interface it is responding to
b. The producer only provides 2 links on a given interface.  One is the local interface, the other is the base URI (presumably public/external) BUT the link has an attribute that signals to the client that it is public/canonical, so it can easily choose.


Thoughts?

joshco:
Another thought about multiple links in the TD; there's only one base attribute.  So if the client is expected to concat the base and the relative URIs in the TD links, it can only arrive at a single hostname/IP.  So the multiple links will end up essentially being the same URLs, even if the TD is rendered specific to each interface.

danielpeintner:
> there's only one base attribute

FYI: https://github.com/w3c/wot-thing-description/issues/803

*Maybe* this is going to change...

danielpeintner:
@egekorkan @relu91 ?

I fear a bit that it breaks some other use-cases I am not even aware of. Any opinion?

joshco:
@danielpeintner @egekorkan @relu91 has there been any resolution in the WG on how you would like to proceed?

danielpeintner:
No decision was made yet.

@sebastiankb  do you have any advice how to continue best?
Shall we merge and fall-back (or re-work) if we encounter issues...

joshco:
## Proposal: Add baseUri to the HttpServer config.

* Base URI's can be transport specific. An http base URI might be different than the coap one, unless it is a proxy from one to the other.
* A TD might be exposed over multiple transports, so the TD base attribute as it is, doesn't seem to cover that case
* You may also have one HttpServer with multiple TD's exposed on it.  eg http://server.com/REDWOT and BLUEWOT

So it feels like this belongs on the transport config rather than the TD, at least the way node-wot works.  It avoids collisions as each transport binding can implement the baseUri attribute as appropriate.

The way it works is if you set the baseUri attribute, then the expose method __will not enumerate all interfaces and create links for each.__   It will just output the one with the baseUri

I've also turned away from an ENV variable for the value.  At least in my experience, there tends to be a run script type of thingy
for rails and node apps that wrap the app code, take command line args, or mate with the deployment infrastructure like Heroku.  It's on the outer ring.  Putting it inside the http binding seems to hide it.  Like the PORT issue, it took a lot of time to troubleshoot because node-wot never told the user it overrode the setting.

Anyway... thoughts?

PS: this is based on master

danielpeintner:
This looks promising to me. Any other opinions?

I suggest to discuss this during our next meeting but feel free to chime in right away!

relu91:
> This looks promising to me. Any other opinions?

At first glance, it seems quite close to what we already have inside the Servient Config. Like ege mentioned in different comments: https://github.com/eclipse/thingweb.node-wot/pull/220#issuecomment-625459297  https://github.com/eclipse/thingweb.node-wot/issues/214#issuecomment-621679897. With the only difference that here we use the `base` field to force an address in the underlying protocol bindings. 

I do not see big shortcomings so I think we can also add this functionality.

danielpeintner:
We will discuss this PR shortly and I will get back to you.

FYI: Unfortunately merging https://github.com/eclipse/thingweb.node-wot/pull/317 caused conflicts...

danielpeintner:
Hi @joshco!

We discussed your PR/proposal.
We do not really have the use case you are describing but see the benefits and your points.

Having said that we are happy to merge your PR. Please
* resolve the recent conflicts
* and add a test

Thanks!

joshco:
OK, I'm on it!

danielpeintner:
@joshco FYI: the Eclipse ECA check failed with "Josh Cohen (joshco@****.com) did not include the "Signed-off-by footer" which is required for all commits made by a contributor."

Note: It did work for the previous commits.

joshco:
@danielpeintner How do you sign your commits without your email address showing up in the commit message?

danielpeintner:
> @danielpeintner How do you sign your commits without your email address showing up in the commit message?

There must be a line 
`Signed-off-by: Joe Somebody <somebody@someplace.net>`
in the commit message.

Having said that, in my case it works without. I just don't really understand why and when Eclipse requires it. Usually if your account is properly linked with your ECA it should work right away. After clicking "recheck" it did work.

Unfortunately your [PR ](https://github.com/eclipse/thingweb.node-wot/pull/322) I merged caused a conflict again. I am very sorry.

danielpeintner:
Please also remove "WIP" in the title since I think it is ready to merge.

joshco:
@danielpeintner lolz, it's actually my own two PRs that are causing conflicts.  The tests I wrote for the port and base uri are both at the bottom of the http-server-test.ts
I'll fix

packages/binding-http/README.md
```diff
@@ -226,6 +227,43 @@ Currently this binding supports only oAuth2.0 `client credential` and `Resource
...
 An example of a WoT oAuth2.0 enabled client can be found [here](../examples/security/oauth).
 
+
+### Using BaseUri
+
+> BaseUri allows the TD to use hostnames, network interfaces, and URI prefixes whicch are not local to the machine exposing the Thing
+
+Assume the coffee machine example is sitting in an office inside the W3C's private network.  It allows you to start the coffee machine before you leave home so it will be ready when you get to work.
+
+* It can be addressed from home via the internet facing domain name `https://wot.w3c.org/things/smart-coffee-machinet`
```
danielpeintner, line 237: Typo ?
smart-coffee-machinet --> smart-coffee-machine

packages/binding-http/README.md
```diff
@@ -226,6 +227,43 @@ Currently this binding supports only oAuth2.0 `client credential` and `Resource
 
 An example of a WoT oAuth2.0 enabled client can be found [here](../examples/security/oauth).
 
+
+### Using BaseUri
+
+> BaseUri allows the TD to use hostnames, network interfaces, and URI prefixes whicch are not local to the machine exposing the Thing
```
danielpeintner, line 233: Typo
whicch -> which

danielpeintner (approved):
Looks good. I think there are some minor types left.

danielpeintner:
@relu91 @egekorkan can anyone of you check as well. Thanks!

packages/binding-http/README.md
```diff
@@ -226,6 +227,43 @@ Currently this binding supports only oAuth2.0 `client credential` and `Resource
...
+```
+
+The exposed thing on the internal server will product form URLs such as:
+
+```json 
+ "actions": {
+    "makeDrink": {
+      "forms": [
+        {
+          "href": "https://wot.w3c.org/pot/things/smart-coffee-machine/actions/makeDrink"
```
relu91, line 264: Should it be: "https://coffee.w3c.org/things/smart-coffee-machine/actions/makeDrink" ?

packages/binding-http/src/http-server.ts
```diff
@@ -205,113 +209,126 @@ export default class HttpServer implements ProtocolServer {
...
+      return new Promise<void>((resolve, reject) => {
+        resolve();
+      });
+    }
+  }
+
+  public addEndpoint(thing: ExposedThing, tdTemplate: WoT.ThingDescription, base: string) {
+      for (let type of ContentSerdes.get().getOfferedMediaTypes()) {
+
+        if (true) { // make reporting of all properties optional?
```
relu91, line 240: Do we still need this `if`? 🤔 since we are refactoring I'd remove it.

packages/binding-http/src/http-server.ts
```diff
@@ -205,113 +209,126 @@ export default class HttpServer implements ProtocolServer {
...
+          let href = base + "/" + this.EVENT_DIR + "/" + eventNamePattern;
+          let form = new TD.Form(href, type);
+          ProtocolHelpers.updateEventFormWithTemplate(form, tdTemplate, eventName);
+          form.subprotocol = "longpoll";
+          form.op = ["subscribeevent", "unsubscribeevent"];
+          thing.events[eventName].forms.push(form);
+          console.debug("[binding-http]",`HttpServer on port ${this.getPort()} assigns '${href}' to Event '${eventName}'`);
+        }
+      }
+    }
```
relu91, line 330: I found it difficult to understand the changes from 242 till here. Could you briefly explain what/why did you change here? Thanks!

relu91 (commented):
In general, it seems ok. I left a couple of comments/concerns. Furthermore, I think we can improve the example adding which are the differences between `servient._staticAddress` (which forces each PB to have the same "base address") and this new parameter (the server socket is not actually binding to the provided DNS address).

packages/binding-http/src/http-server.ts
joshco: I created a new function called addEndpoint, refactoring the functionality that takes the TD template and turns it into links/href.

It used to be inline to the for loop enumerating addresses. The baseUri if/else logic would have been more complicated, since the add endpoint logic was buried in the for loop enumerating the physical addresses, (With baseUri, we don't care about the local addresses in the TD links, and dont want a client to be distracted/confused by interfaces it wont be able to connect to)

Ye Old Way:
https://github.com/eclipse/thingweb.node-wot/blob/45d8fa739cb6e796c2ee38bf156a977c4f0739bf/packages/binding-http/src/http-server.ts#L188-L278

By having it separated, and public, a developer who wants to do some other crazy thing, can just add their own endpoints.

joshco:
On it.  Addressing your code review feedback.

Also:
> In general, it seems ok. I left a couple of comments/concerns. Furthermore, I think we can improve the example adding which are the differences between `servient._staticAddress` (which forces each PB to have the same "base address") and this new parameter (the server socket is not actually binding to the provided DNS address).

Agreed.  The baseURI is a URI prefix.  staticAddress is just a hostname or IP, which becomes part of the full URL

Also, in the CLI Readme, the issue is further confused by referring to an environment variable ADDRESS, which as far as I can tell doesn't actually exist in the code, but the value is specified as a URL rather than hostname/IP.
https://github.com/eclipse/thingweb.node-wot/pull/319/files#diff-fc8b9e5b88bd2528cdf853fd8285a65eR101-R108

```
PORT=4242
ADDRESS=http://hello.com
```

I propose that we only use the term 'base' in the context of URIs  Unless I've read the spec wrong, which is always a possibility, WoT resources are addressable by URIs, not IPs, hostnames, ports, or paths.  

And technically, URIs are supposed to be opaque to the application client above the URI transport layers.

packages/binding-http/src/http-server.ts
joshco: removed

danielpeintner:
@joshco are you still interested in proceeding?

joshco:
@danielpeintner Yes!  
Apologies for not being able to attend the w3c scenario breakout today.  Is it recorded?
Did this scenario come up?

danielpeintner:
> Apologies for not being able to attend the w3c scenario breakout today. Is it recorded?

Yes, I think it has been recorded but I do not know whether it is already available ...

joshco:
Are there tasks for me to do?  Did I address all of your feedback?

relu91 (approved):
The content looks good. The only problem that I see is that your commit messages do not follow the specification described in the [CONTRIBUTING.md](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md).

joshco:
I'll work on that. standby

joshco:
I've changed the commit messages, with one exception.  After my first commit, I was a bad dev and merged master into my feature branch, so trying to edit the message triggered a rebase which was too hairy.  Instead, I made an empty "errata" commit, that adds a changelog appropriate message. It's the feat one:

https://github.com/eclipse/thingweb.node-wot/pull/220/commits/7301557082ca0774e140dc7d9c3ee8900e05cd36

Let me know if this doesn't work for you, and how I should go about fixing that.  I'll do better in the future.

joshco:
@relu91 hang on a bit, i think I was able to fix all the commit messages with a total rebase.  See the commit history.
However, before you merge, this I'd actually like to run the tests, and test my apps with it.  I want to make sure the git voodoo didn't leave out a commit or something.

joshco:
OK, it's now ready.  I've squashed the commits, and fixed a test conflict, with the commit messages conforming to the protocol.

danielpeintner:
@relu91 I take your hooray as "approved". Right?

relu91 (approved):
Yes, I think we can merge it. 👍

danielpeintner:
Great! Thank you all for your work!

PR #228, opened 2020-05-12
Update to latest wot-typescript-definitions 0.7.3

danielpeintner:
* formIndex
* <s>InteractionData<s>


danielpeintner:
I would like to merge this PR "as is" even though it just implements the `formIndex` to be able to select a *dedicated* form.

The `InteractionData` discussion goes on and there are likely more updates/revisions expected. Any opinions?

Issue #21, opened 2018-07-15
subscribe() for ConsumedThingProperty


mkovatsc:
Needs to be implemented...
`ExposedThingProperty` needs a `Subject` and `next()` calls.

mkovatsc:
For now at least a placeholder that provides a meaningful error ("not implemented").

alexbrdn:
It seems to me that `subscribe()` did [get implemented](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/core/src/consumed-thing.ts#L243) in the meantime, but it fails to work with coaps (I'm using good old Trådfri lamps).

Here's what I was able to figure out: [`ContentSerdes.contentToValue()`](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/core/src/content-serdes.ts#L114) expects the `content` to have field `body`. [`read`](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-coap/src/coaps-client.ts#L51) and [`invoke`](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-coap/src/coaps-client.ts#L79) do provide `content.body` to the callback. However, [`subscribe`](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-coap/src/coaps-client.ts#L97) calls `coaps.observe()`, which results in the response message from [`CoapClient`](https://github.com/AlCalzone/node-coap-client/blob/master/src/CoapClient.ts#L773) to be sent directly to `next` callback, bypassing the conversion as in `read` and `invoke`.

Result: Parsing error at [`codec.bytesToValue()`](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/core/src/content-serdes.ts#L114), because `content.body` is undefined.

The fix for coaps is simple enough, but I don't know how well it will play with all other bindings, hence no PR.

sebastiankb:
quick question: which node-wot version are you using?

alexbrdn:
* 7ff8cdd (HEAD, tag: v0.6.2) chore: update version tag for 0.6.2 node-wot release

danielpeintner:
Is this issue still unresolved even after the *new* Scripting API changes/updates?

danielpeintner:
HTTP is fixed by https://github.com/eclipse/thingweb.node-wot/pull/233 while CoAP has been fixed before.

Please reopen if you encounter any related issue.

Issue #151, opened 2019-11-25
[binding-http] Unsubscribing from an Event does not work


adriankast:
# Setup
- thingweb.node-wot v0.7.0-SNAPSHOT.3
- consumed Thing with http form
# Doing
- subscribe to an Event with the `subscribeEvent(name, listener, options?)` function of a consumed Thing
- trying to unsubscribe from the mentioned Event with the `unsubscribeEvent(name)` of the same Thing
# Expected Behavior
The given `listener` should not be called anymore and no further http `GET` requests being sent, for the mentioned event.
# Observed Behavior
- The function `unsubscribeEvent` calls `unlinkResource` of the protocol binding, which for http sends a `DELETE` Request to the event form. This leads to the error:
```Javascript
http-client.js?7d6f:334 Uncaught (in promise) Error: Client error: Method Not Allowed
    at HttpClient.checkResponse (http-client.js?7d6f:334)
    at IncomingMessage.eval (http-client.js?7d6f:133)
    at IncomingMessage.emit (events.js:199)
    at endReadableNT (_stream_readable.js:1141)
    at processTicksAndRejections (internal/process/task_queues.js:81)
```
while the `listener` function is still being called and the longpoll protocoll kept alive (-> further http `GET` requests are sent)

egekorkan:
This issue is also present in CoAP. In order to stop observing, the client should send a GET request with Observe option set to 1. At the moment it sends a DELETE request, like in HTTP, which is semantically incorrect since we don't want to delete the resource itself but just the observation.

danielpeintner:
fixed by https://github.com/eclipse/thingweb.node-wot/pull/233

PR #233, opened 2020-05-27
fix: properly unsubscribe by stopping HTTP long polling

danielpeintner:
see https://github.com/eclipse/thingweb.node-wot/issues/151

### DIFF
--- a/packages/binding-http/src/http-client.ts
+++ b/packages/binding-http/src/http-client.ts
@@ -172,6 +172,8 @@ export default class HttpClient implements ProtocolClient {
         if (this.activeSubscriptions.has(form.href)) {
           next({ type: result.headers.get("content-type"), body: buffer })
           polling()
+        } {
+          complete && complete()
         }
       } catch (e) {
         error && error(e)
--- a/packages/binding-http/src/http-client.ts
+++ b/packages/binding-http/src/http-client.ts
@@ -46,6 +46,8 @@ export default class HttpClient implements ProtocolClient {
 
   private credential:Credential = null;
 
+  private activeSubscriptions = new Set();
+
   constructor(config: HttpConfig = null, secure = false, oauthManager: OAuthManager = new OAuthManager()) {
 
     // config proxy by client side (not from TD)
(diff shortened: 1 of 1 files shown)


danielpeintner:
@egekorkan @relu91 May I ask you to take a look at the proposed fix for unsubscribing HTTP events/properties. I did experiment with it and it seems to work fine for me...

@relu91 specifically I think the whole code in `unlinkResource`

https://github.com/eclipse/thingweb.node-wot/blob/7652a56c8d0dbb06b7c558bd52d2c05f47ebb534/packages/binding-http/src/http-client.ts#L145-L161 (other than `this.activeSubscriptions.delete(form.href)`) can be removed. No need to send DELETE to HTTP server because it does not know what to do.. we do HTTP GET (long polling) on client side only? Am I right?

relu91:
I am a little bit puzzled as you. When I refactored the code I thought that sending a delete request was a specification of the `longpoll` protocol.  However, I agree on what @egekorkan is saying in https://github.com/eclipse/thingweb.node-wot/issues/151#issuecomment-558135107. Therefore, I would delete the HTTP request and leave only `this.activeSubscriptions.delete(form.href) `. 

Probably, we should also abort the pending HTTP GET to clean up resources immediately. See this [example on MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort)

danielpeintner:
>  However, I agree on what @egekorkan is saying in [#151 (comment)](https://github.com/eclipse/thingweb.node-wot/issues/151#issuecomment-558135107).

Mhh, not sure. CoAP is different in the way that it supports "observing" as part of the protocol. HTTP does not. @egekorkan ?

> Therefore, I would delete the HTTP request and leave only `this.activeSubscriptions.delete(form.href) `.

Will update the PR accordingly.

egekorkan:
Yes CoAP has it built and this is already fixed no?

danielpeintner:
> Yes CoAP has it built and this is already fixed no?

Yes, see https://github.com/eclipse/thingweb.node-wot/commit/00bbb0df8ba377767ca61e1175c6b5b4a7dee613

egekorkan:
Maybe this also fixed https://github.com/eclipse/thingweb.node-wot/issues/21 ?

danielpeintner:
> Maybe this also fixed #21 ?

Yes, correct. The issue pointed out CoAP which has been fixed before.

PR #234, opened 2020-05-29
fix: https://github.com/eclipse/thingweb.node-wot/issues/202

danielpeintner:
"find" forms only that match op field

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -93,7 +93,7 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
 
         // find right operation and corresponding scheme in the array form
         for (let f of forms) {
-            let fop: string | Array<string>;
+            let fop: string | Array<string> = "";
             if (f.op != undefined) {
                 fop = f.op;
             } else {
@@ -111,11 +111,10 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
                         break;
                 }
             }
-            if (fop != undefined)
-                if (fop.indexOf(op) != -1 && f.href.indexOf(schemes[idx] + ":") != -1) {
-                    form = f;
-                    break;
-                }
+            if (fop.indexOf(op) != -1 && f.href.indexOf(schemes[idx] + ":") != -1) {
+                form = f;
+                break;
+            }
         }
 
         // Note: form can be null if no appropriate op can be found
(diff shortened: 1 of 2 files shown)


danielpeintner:
@sebastiankb @egekorkan @erkann-sen @relu91 

I believe the PR is fine I would just like "someone" quickly scan over it.
I also added a new testcase w.r.t. observe only properties

packages/core/src/consumed-thing.ts
```diff
@@ -82,23 +88,37 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
...
+                        break;
+                    case Affordance.ActionAffordance:
+                        fop = "invokeaction";
+                        break;
+                    case Affordance.EventAffordance:
+                        fop = "subscribeevent";
+                        break;
+                }
+            }
+            if (fop != undefined)
```
relu91, line 114: I think this check can be avoided if we initialize fop to an empty string.

danielpeintner: Sure, fop initialized with an empty string would remove this check.
However, I am not sure if it conveys the same meaning. 
`undefined` to me means *really* undefined while an empty string is actually something else. Am I wrong?

However, either way is fine to me...

danielpeintner:
> As a general note, it might worth having a single place where the Form is validated and filled with the default values. So that from that point on it is considered valid and we do not have to check every time if a field is undefined or null. I tried to apply this logic inside the Modbus PR (see https://github.com/eclipse/thingweb.node-wot/pull/186/files#diff-828e8079bd3d0d416f08e6345ee7bdb4R132)

Makes sense to me. Shall we open a new issue and accordingly PR for that?

packages/core/src/consumed-thing.ts
relu91: In general yes `undefined` has a particular meaning, however, in this case, it is used as "default initialization value" so an empty string might work as well. Also, now that I'm looking again at this piece of code, it seems not even possible that `fop` is not assigned to a `not` `undefined` value. In the if branch is assigned to f.op (which is not undefined because of the if boolean expression). On the other hand, in the else branch is assigned to static values, unless we are passing something that is not in the case list (which should throw an Error I guess).

danielpeintner: I updated the PR accordingly (fop is initialized with empty string)

PR #235, opened 2020-06-01
Use tsconfig extends mechanism

relu91:
This PR refactors the project configuration using tsconfig `extends` keywords. Therefore, now the basic tsc configuration is contained inside the root `tsconfig` file, whereas package specific settings are described inside the relative `tsconfig.json`.

Finally, it also moves the security example inside the src folder so that it gets built automatically as the other examples.


danielpeintner:
Looks good to me.

@egekorkan @sebastiankb ?

egekorkan:
Good idea, I also find it a better approach!

PR #240, opened 2020-06-05
fix(cli): allow loading of files transpiled from typescript

ajs124:
As noted [here](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/examples#workflow) examples/code transpiled from typescript cannot be run with the wot-servient cli right now.

The described process is to manually remove these lines before copying them. Since this is quite tedious, I automated the process.

This is only one of several possible approaches to achieve this goal. Another one would be to run something from `npm` after the build (`tsc`). Which approach would you prefer?

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -26,6 +26,7 @@ const defaultFile = "wot-servient.conf.json";
 const baseDir = ".";
 
 var clientOnly: boolean = false;
+var sanitizeCode: boolean = false;
 
 var flagArgConfigfile = false;
 var confFile: string;
@@ -58,6 +59,15 @@ const readConf = function (filename: string): Promise<any> {
     });
 }
 
+const sanitizeScript = function(input: string): string {
+    let s = input.replace(/^Object\.defineProperty\(exports, "__esModule", { value: true }\);$/m, '')
+        .replace(/^require\("wot-typescript-definitions"\);$/m, '')
+        .replace(/^let WoT;$/m, '')
+        .replace(/^let WoTHelpers;$/m, '')
+        .replace(/^exports.* = .*;$/gm, '');
+    return s;
+}
+
 const runScripts =async function(servient: DefaultServient, scripts: Array<string>,debug?: DebugParams) {
     const launchScripts = (scripts : Array<string> ) => {
         scripts.forEach((fname : string) => {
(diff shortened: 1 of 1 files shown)


relu91:
My vote is to handle this sanitation at npm level. A `postbuild` hook plus a script should work quite well in my opinion. Another solution is to use [wam](https://github.com/UniBO-PRISMLab/wam) but it does not handle multi-output builds (yet).

danielpeintner:
The related issue is https://github.com/eclipse/thingweb.node-wot/issues/171 (see also PR https://github.com/eclipse/thingweb.node-wot/pull/178#issuecomment-592057917 that mentions the pending issues)

I agree with @relu91  that we should try to find a way that handles the JS conversion automatically.

Maybe it is also *wise* to step back once again and try to find another solution that avoids duplicating the *almost* same code? The reason that we introduced the workflow we are using right now was that the TS files in `/packages/examples` give us the check that code is according to the _latest_ Scripting API. JS code cannot do that for us.

Any opinions how we can improve the situation?

ajs124:
I don't think I know enough about the TS/JS ecosystem to come up with a more elegant solution than regex, which is admittedly very brute force. If the regex approach is to be kept, I'd prefer a solution like suggested here instead of at `postbuild`, because rewriting code without any semantic understanding of it (which is what regex does) is inherently fragile and can break legitimate code, which is less of a problem when it can be turned off at runtime.

relu91:
>Maybe it is also wise to step back once again and try to find another solution that avoids duplicating the almost same code? The reason that we introduced the workflow we are using right now was that the TS files in /packages/examples give us the check that code is according to the latest Scripting API. JS code cannot do that for us.

I am afraid that IF we want to maintain BOTH js and ts examples we are constrained to have this kind of *workflow*. However, we can open a discussion on how we are going to maintain examples in the future. For example, sometime the example could be an entire project rather than just a script (see the oauth example).

> I don't think I know enough about the TS/JS ecosystem to come up with a more elegant solution than regex, which is admittedly very brute force. If the regex approach is to be kept, I'd prefer a solution like suggested here instead of at postbuild, because rewriting code without any semantic understanding of it (which is what regex does) is inherently fragile and can break legitimate code, which is less of a problem when it can be turned off at runtime.

About this, the proper way to handle the sanitation process that you implemented is to manipulate the AST. In JS AST manipulation is almost every time done using  [babel](https://babeljs.io/). The steps would be:
1. Write a babel plugin that does the sanitation. (ie. remove unwanted nodes fro AST). Check [Babel Plugin Handbook](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-asts)
2. in a `postbuild` call babel configured with the plugin form point one

I see this solution a little bit overkilling, but it assures code compilation after the "sanitation".

danielpeintner:
we can close this, right?
@relu91 @ajs124 due to https://github.com/eclipse/thingweb.node-wot/pull/321

Issue #229, opened 2020-05-20
[proposal] slightly improve logs


relu91:
In #186 I tried to use a simple patter for logging. I think it might be useful if it is adopted for the whole library. it does not require external libraries but it gives slightly better control over what is logged and what is not. 

The pattern is: `console.debug("[package-name]", "log message)`. Following this rather simple pattern can provide two positive outcomes:
1. Quickly identify where the log message is generated 
2. Allows filtering procedures to just shut down package level messages. 

Considering the second point, an example could be this js script:
```javascript
// turn off messages from core package
const debug = console.debug
console.debug = (package,...args) => {
 if(package !== "[core]"){
    debug(package,...args)
 }
}
```

In general, I would prefer to use just `debug`, `warn`, and `error` leaving `info` and `log` reserved to applications that use node-wot. 

Notice that with this suggestion this statement from README remains valid: 

> We used to have a node-wot-logger package to allow fine-grained logging (by means of Winston). It turned out though that depending on the actual use-case other logging libraries might be better suited. Hence we do not want to prescribe which logging library to use. Having said that, we use console statements which can be easily overriden to use the prefered logging library if needed (see here).

danielpeintner:
I like the suggested approach. We would need to

* document that approach in the README (as *best* practice)
* walk over the code and update accordingly (i.e., adding "[package]" and replace any info/log level messages with debug)

sebastiankb:
issue is kind of related to each other #193

relu91:
@sebastiankb @danielpeintner I am working on this.

relu91:
I think we can close this issue after #250

PR #250, opened 2020-06-16
Improve logging

relu91:
This PR implements and fixes what described in #229. I did not modify log messages generated inside examples and tests as I think them as different from the one created inside the library itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,12 @@ Note: More mediaTyes can be easily added by implementing `ContentCodec` interfac
 
 We used to have a node-wot-logger package to allow fine-grained logging (by means of Winston). It turned out though that depending on the actual use-case other logging libraries might be better suited. Hence we do not want to prescribe which logging library to use. Having said that, we use console statements which can be easily overriden to use the prefered logging library if needed (see [here](https://gist.github.com/spmason/1670196)).
 
+The logs in the library follows those best practice rules (see [here](https://github.com/eclipse/thingweb.node-wot/issues/229)):
+1. Tag log messages with the package as following: `console.debug("[package-name]", "log message)`. This is useful to identify which package generated the log.
+2. Avoid to use `info` and `log` in packages other than the cli package.
+
+Please follows these rules if you are going to contribute to node-wot library.
+
 ### Install new/different versions of NodeJS
 
 Using NPM, you can install NodeJS independent from the usually outdated package managers such as apt. This is nicely done by n:
--- a/packages/binding-http/src/http-client.ts
+++ b/packages/binding-http/src/http-client.ts
@@ -66,7 +66,7 @@ export default class HttpClient implements ProtocolClient {
         secure = true;
       }
       
-      console.info("[binding-http]",`HttpClient using ${secure ? "secure " : ""}proxy ${this.proxyRequest.hostname}:${this.proxyRequest.port}`);
+      console.debug("[binding-http]",`HttpClient using ${secure ? "secure " : ""}proxy ${this.proxyRequest.hostname}:${this.proxyRequest.port}`);
     }
 
     // config certificate checks
@@ -90,7 +90,7 @@ export default class HttpClient implements ProtocolClient {
 
   public async readResource(form: HttpForm): Promise<Content> {
     const request = await this.generateFetchRequest(form, "GET")
-    console.info("[binding-http]",`HttpClient (readResource) sending ${request.method} to ${request.url}`);
+    console.debug("[binding-http]",`HttpClient (readResource) sending ${request.method} to ${request.url}`);
 
     let result = await this.fetch(request)
     
@@ -114,7 +114,7 @@ export default class HttpClient implements ProtocolClient {
     
     let result = await this.fetch(request)
 
-    console.info("[binding-http]",`HttpClient received ${result.status} from ${result.url}`);
+    console.debug("[binding-http]",`HttpClient received ${result.status} from ${result.url}`);
 
     this.checkFetchResponse(result)
     
@@ -130,11 +130,11 @@ export default class HttpClient implements ProtocolClient {
       body : content?.body
     })
 
-    console.info("[binding-http]",`HttpClient (invokeResource) sending ${request.method} ${content ? "with '" + request.headers.get("Content-Type") + "' " : " "}to ${request.url}`);
+    console.debug("[binding-http]",`HttpClient (invokeResource) sending ${request.method} ${content ? "with '" + request.headers.get("Content-Type") + "' " : " "}to ${request.url}`);
 
     let result = await this.fetch(request)
 
-    console.info("[binding-http]",`HttpClient received ${result.status} from ${request.url}`);
+    console.debug("[binding-http]",`HttpClient received ${result.status} from ${request.url}`);
     console.debug("[binding-http]",`HttpClient received Content-Type: ${result.headers.get("content-type")}`);
     
     const buffer = await result.buffer()
@@ -143,7 +143,7 @@ export default class HttpClient implements ProtocolClient {
   }
 
   public async unlinkResource(form: HttpForm): Promise<any> {
-    console.info("[binding-http]",`HttpClient (unlinkResource) ${form.href}`);
+    console.debug("[binding-http]",`HttpClient (unlinkResource) ${form.href}`);
 
     this.activeSubscriptions.delete(form.href)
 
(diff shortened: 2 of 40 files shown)


egekorkan:
Just for the clarifyin something for me, when you use console.debug, it will log only in the debug level and below right? Since this is a modification that some might be relying on ( I do for some small device implementations), it should be explained somewhere how to enable logging in general. The cli config does this already but maybe already in the readme? I say this since everything was logged even if the developer was not specifying anything, it breaks their expectation.

README.md
```diff
@@ -274,6 +274,12 @@ Note: More mediaTyes can be easily added by implementing `ContentCodec` interfac
 
 We used to have a node-wot-logger package to allow fine-grained logging (by means of Winston). It turned out though that depending on the actual use-case other logging libraries might be better suited. Hence we do not want to prescribe which logging library to use. Having said that, we use console statements which can be easily overriden to use the prefered logging library if needed (see [here](https://gist.github.com/spmason/1670196)).
 
+The logs in the library follows those best practice rules (see [here](https://github.com/eclipse/thingweb.node-wot/issues/229)):
```
egekorkan, line 277: My comment at https://github.com/eclipse/thingweb.node-wot/pull/250#issuecomment-644993618 can be added somewhere around this paragraph

packages/binding-http/src/http-client.ts
```diff
@@ -90,16 +90,16 @@ export default class HttpClient implements ProtocolClient {
 
   public async readResource(form: HttpForm): Promise<Content> {
     const request = await this.generateFetchRequest(form, "GET")
-    console.info(`HttpClient (readResource) sending ${request.method} to ${request.url}`);
+    console.debug("[binding-http]",`HttpClient (readResource) sending ${request.method} to ${request.url}`);
```
egekorkan, line 93: mapping info to debug, is it intended?

egekorkan:
So my general comment is to document the proposed mapping. Please change if the following is not correct:
- log -> debug
- info -> debug
- error -> error
- warn -> warn

egekorkan:
It would be also good to differentiate between cli and cli-default-servient since both can be used at the same time whereas in the other packages they are either client or server and are less likely to be used together

egekorkan:
Please don't take my comments badly, I think the improvement is more than a slight improvement! Thank you @relu91 :)

relu91:
Thank you Ege for your comments, here my thoughts. 

>Just for the clarifyin something for me, when you use console.debug, it will log only in the debug level and below right? Since this is a modification that some might be relying on ( I do for some small device implementations), it should be explained somewhere how to enable logging in general. The cli config does this already but maybe already in the readme? I say this since everything was logged even if the developer was not specifying anything, it breaks their expectation.

In my understanding Node console does not really have the concept of *log levels*. In fact, most of the functions supported by console are just aliases for `console.log` and `console.error` (see [here](https://nodejs.org/docs/latest-v10.x/api/console.html#console_console_debug_data_args) for example). So when we use `console.debug` we are just using `console.log`. So why bother? Well using that function we can emulate *log levels*. For example, consider the following as the default level order:

0. Off
1. Error
2. Warning
3. Info
4. Debug

Setting a level the log to 2 means that `console.info = () => {}` and `console.debug=() =>{}` . This PR is not about adding this functionality. @fatadel is working on this, because I see it as a different feature (i.e. implement log levels). 

>log -> debug
info -> debug
error -> error
warn -> warn

I was unsure to map info to debug. However, most of the info messages seemed more suited in the debug messages. We can discuss which one may make more sense as an info message even if I think that the lesser level we use the easier is to manage them. 

On the other hand about `console.log` I think that most of the users just do not care about a complex logging system and in the end just use `console.log`. Using this function in node-wot will make it harder for such users to shut down the logs completely (i.e. we cannot do  `console.log = () => {}` because it shuts down also application messages).  

So the rule is not a mapping, it is more do not use log at all and think twice before use `info`. 

>It would be also good to differentiate between cli and cli-default-servient since both can be used at the same time whereas in the other packages they are either client or server and are less likely to be used together.
Similarly for core which has other subpackages inside

This could improve even more the detection of the log messages. I thought a pattern like the following: `[packageName:ScriptName] Some message`. What do you think? too long?

egekorkan:
- Thanks for the clarification regarding upcoming PR by @fatadel :)
- Regarding not having simple console.log, it also makes a lot of sense.
- Agree on having less levels. In the end how should I make a decision whether I want warn or info :)
- `[packageName:ScriptName]` makes sense, e.g. `[core-codec/json]`. I think this would be needed only for the core packages since the protocol bindings differentiate themselves by being Consumer or Server anyways but it wouldn't hurt to add the suffix there as well :)

relu91:
> [packageName:ScriptName] makes sense, e.g. [core-codec/json]. I think this would be needed only for the core packages since the protocol bindings differentiate themselves by being Consumer or Server anyways but it wouldn't hurt to add the suffix there as well :) 

Ok, I'll change the PR to draft while I'm adding that change.

danielpeintner:
@relu91 Thank your very much for all the work.

I looked at the PR and it looks very good.

Some comments/questions below.

### I found `console.info` in one file

I guess we might find more but we can tackle them once we encounter them.

packages\binding-coap\src\coaps-client.ts, line 149 --> console.info(`HttpClient overriding client-side proxy with security proxyURI '${security.proxyURI}`);

### Id/Ids in one file per package?
One question I would like to raise is whether it makes sense to have a *single* place in each package for the identifier like "[binding-http]". This would make it simple to change it in the future (IF we ever want to do so).
In some cases (other than binding) we have multiple identifiers like "[core/helpers]", "[core/exposed-thing]", .. but even there it might make sense to have all const values in a single place?

Does that make sense or is it not worth it?
I am myself unclear what is best...

### Consistent identifier
Question: in in all bindings we use the package name except "[binding-ws]" where it might be "[binding-websockets]" to be consistent?


### Another nit-picking comment is about formatting.

When using the VisualStudio code formatting it adds spaces between , et cetera.

For example
`console.debug("[binding-http]",`HttpClient (readResource) sending ${request.method} to ${request.url}`);`
becomes
`console.debug("[binding-http]", `HttpClient (readResource) sending ${request.method} to ${request.url}`);`

Does it make sense to do this clean-up also in this PR? Or shall we do that formatting clean-up in another PR we plan to make anyway...

### Merge conflict 
We also need to fix the merge conflict that was raised by a recent PR merge...

relu91:
Thank you @danielpeintner for the great review! here they are my thoughts:
> I found console.info in one file
I guess we might find more but we can tackle them once we encounter them.

packages\binding-coap\src\coaps-client.ts, line 149 --> console.info(HttpClient overriding client-side proxy with security proxyURI '${security.proxyURI});

Nice catch! it escaped my review process thank you for point it out. 

> Does that make sense or is it not worth it?
I am myself unclear what is best...

It does make sense. However, I am afraid that going that path will lead us to have a full-blown logging framework... with this PR, I wanted to make a micro-step toward better log management. I mean, if we gonna use "global/local" `const` tags variables why not make a dedicated logger class for each package --> why don't we factorize it a stand-alone package? --> now we have a logging framework. As you said I am not certain that we do need/not need this. 

>Question: in in all bindings we use the package name except "[binding-ws]" where it might be "[binding-websockets]" to be consistent?

I agree.

> Another nit-picking comment is about formatting.

I am not sure if I can do this without formatting the entire file. One question: this means that my vscode instance is not using the `.editorconfig` rules?

> We also need to fix the merge conflict that was raised by a recent PR merge...

I'll do it 👍 

Changing back to draft while I'll make the changes requested above

relu91:
Just find out that `packages\binding-coap\src\coaps-client.ts, line 149` is a commented line. I'll leave the comment as it is.

danielpeintner:
> Just find out that `packages\binding-coap\src\coaps-client.ts, line 149` is a commented line. I'll leave the comment as it is.

Sorry, my bad!

> > Another nit-picking comment is about formatting.
> 
> I am not sure if I can do this without formatting the entire file. One question: this means that my vscode instance is not using the `.editorconfig` rules?

I have to explicitly say "Right mouse click" --> "Format selection" ...
It is *not* done automatically.

> > Does that make sense or is it not worth it?
> > I am myself unclear what is best...
> 
> It does make sense. However, I am afraid that going that path will lead us to have a full-blown logging framework... with this PR, I wanted to make a micro-step toward better log management. I mean, if we gonna use "global/local" `const` tags variables why not make a dedicated logger class for each package --> why don't we factorize it a stand-alone package? --> now we have a logging framework. As you said I am not certain that we do need/not need this.

Maybe you are right and we better stick with the *simple* approach.
Any other opinions?

If not, I am OK with merging!

PR #253, opened 2020-06-19
fix(binding-mqtt): pass along mqtt payload properly

### LINKED ISSUE #224: [binding-mqtt] MQTT message payload could not be accessed in the ActionHandler
### Description

I am exposing a Thing via MQTT and when I try to access the payload of an MQTT message in setActionHandler, I get an error. More details to reproduce it are posted below, but it is likely due to the mqtt package (dependency).
I am using the binding-mqtt-0.7.0-SNAPSHOT.6 and one of its dependancies is the mqtt package version 3.0.0. (As it is in GitHub.) 
Another device in my network publishes a message periodically on the relevant topic (setNumber1) and as long as I don't want to access the payload of the MQTT message everything works as intended in setActionHandler.

### To reproduce

My code:

```js
Servient = require("@node-wot/core").Servient
MqttBrokerServer = require("@node-wot/binding-mqtt").MqttBrokerServer

servient.addServer(new MqttBrokerServer("mqtt://192.168.178.38")); //change for another broker, like test.mosquitto.org

servient.start().then((WoT) => {

    WoT.produce({
        "@context": "https://www.w3.org/2019/wot/td/v1",
        "id": "de.tum:ei:esi",
        "title": "example_td",
        "securityDefinitions": { "nosec_sc": {"scheme": "nosec"}
        },
        "security": ["nosec_sc"],
        "actions":{
            "setNumber1":{
                "input":{
                    "type": "integer"
                }
            }
        }
    })
    .then((thing) => {
    console.log("Produced " + thing.getThingDescription().title);

    // set action handlers
    thing.setActionHandler("setNumber1", (msg) => {
        
         return new Promise ((resolve, reject) => {
             console.log(msg);
             resolve();
         });
    });

    // expose the thing
    thing.expose().then(() => {
        console.info(thing.getThingDescription().title + " ready");
    }).catch((e) => {
        console.log(e);
    });
});
```

I am unable to access the MQTT message payload over msg and console.log(msg) returns :
```bash
function () { … }

    [[FunctionLocation]]:internal#location
    [[Scopes]]:Scopes[4]
    arguments:TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
    caller:TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
    length:0
    name:""
    prototype:Object {constructor: }
    __proto__:function () { … }
```
### Hint

A looked at the mqtt-broker-server.ts file and found out that in line 176 a problem occurs with packet.properties.
When looking up mqtt in npm I saw that there were some breaking changes in v2.0.0.

ajs124:
The library we use changed some stuff around, now everything is a `Buffer`, so we need to handle that.

Additionally 32f787951c5ad29f444e1ab41fc4b46e3f592a74 contained a partial fix, which lead to the behavior observed in #224.

Fixes #224.

### DIFF
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -34,11 +34,11 @@ export default class MqttBrokerServer implements ProtocolServer {
   private user: string = undefined; // in the case usesername is required to connect the broker
 
   private psw: string = undefined; // in the case password is required to connect the broker
-  
+
   private clientId: string = undefined; // in the case clientId can be used to identify the device
 
   private protocolVersion: number = undefined;
-  
+
   private brokerURI: string = undefined;
 
   private readonly things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
@@ -155,10 +155,16 @@ export default class MqttBrokerServer implements ProtocolServer {
       }
 
       // connect incoming messages to Thing
-      this.broker.on("message", (receivedTopic: string, payload: string, packet: IPublishPacket) => {
+      this.broker.on("message", (receivedTopic: string, rawPayload: Buffer | string, packet: IPublishPacket) => {
 
         // route request
         let segments = receivedTopic.split("/");
+        let payload: Buffer;
+        if (rawPayload instanceof Buffer) {
+            payload = rawPayload;
+        } else if (typeof rawPayload === "string") {
+            payload = Buffer.from(rawPayload);
+        }
 
         if (segments.length === 4 ) {
           // connecting to the actions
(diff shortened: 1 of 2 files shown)


ajs124:
cc @egekorkan

egekorkan:
I have tested this and can verify that the payload is passed and parsed correctly if it is JSON or not. However, I am not able to test whether the properties of the packet are checked since I cannot verify whether they are received by the Thing. I have tested with `mosquitto_pub -h test.mosquitto.org -p 1883 -t /example_td/actions/setNumber1 -m "123" -D PUBLISH content-type "application/json"` (tried with different " combinations) for the Thing in #224. It never goes into the branch with content serdes. Doing a console.log of the packet gives the following:
```json
{"cmd":"publish","retain":false,"qos":0,"dup":false,"length":35,"topic":"/example_td/actions/setNumber1","payload":{"type":"Buffer","data":[49,50,51]}}
```

@ajs124 how did you test the intended behavior and could you also add tests for this while also addressing https://github.com/eclipse/thingweb.node-wot/issues/161 ?

ajs124:
Seems like that branch can only be reached once [this](https://github.com/mqttjs/MQTT.js/pull/1103) is merged upstream.
To be honest, I just took the code that was already there, because it looked reasonable enough. I didn't actually test that part.

We can either drop it or leave it, because it will probably work once upstream fixes this.
I'll look into add some tests and fixing #161.

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -167,32 +173,38 @@ export default class MqttBrokerServer implements ProtocolServer {
...
-                    }
-                  })
-                  .catch(err => {
-                    console.error(`MqttBrokerServer at ${this.brokerURI} got error on invoking '${segments[3]}': ${err.message}`);
-                  });
-                // topic found and message processed
-                return;
-              } // Action exists?
-            }
+                if ('properties' in packet && 'contentType' in packet.properties) {
```
egekorkan, line 179: A note should be added here since it is not testable atm

egekorkan:
In this case, I would recommend to merge this PR since it fixes an issue first. However, a note in the first check/branch whether contentType exists in the payload. See my comment above

egekorkan:
@danielpeintner what do you think?

danielpeintner:
> @danielpeintner what do you think?

Makes sense.
I will wait till the the PR is updated with the requested note...

ajs124:
Rebased to fix merge conflict and added comment.

PR #258, opened 2020-06-23
[http-binding] OAuth server side token validation 

relu91:
This PR introduces token validation for OAuth tokens in HTTP server implementation. OAuth token validation can be carried out using different methods. This PR uses a standard protocol to request token validation to a `introspection-endpoint`. Specifically, the protocol is defined in [RFC7662](https://tools.ietf.org/html/rfc7662) and it defines how a resource server interacts with the authorization server to validate a token. The protocol is quite flexible but it has some security implications and therefore it is advised to be used with caution. For example, the `introspection-endpoint` must be not visible to other clients or it must be protected using an authentication method (basic or even OAuth again).

This PR is still a draft because I want to complete the OAuth example with also the server-side validation.

### DIFF
--- a/examples/security/oauth/README.md
+++ b/examples/security/oauth/README.md
@@ -1,13 +1,11 @@
-# oAuth example
-The example is composed by two actors: one Thing that expose an action with oAuth2.0 client credential security constraint and a client who want to use that function. In the example, we are using an utility server to generate tokens and validate client credentials. Therefore, the **wot-consumer** force the action `href` to be the same as the utility server with the goal to validate the obtained oAuth2.0 token. In the feature the exposing servient could also play this role. 
+# oAuth JavaScript example
+This is example is composed by two actors: one Thing that expose an action with oAuth2.0 client credential security constraint and a client who want to use that function. In the example, we are using an utility server to generate tokens and validate client credentials. Therefore, the **wot-consumer** force the action `href` to be the same as the utility server with the goal to validate the obtained oAuth2.0 token. In the feature the exposing servient could also play this role. 
 
 ## run the example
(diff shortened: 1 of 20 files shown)


danielpeintner:
FYI: PR https://github.com/eclipse/thingweb.node-wot/pull/262 is merged

danielpeintner:
Sorry for no progress on this...

Who is able to review this PR?
@egekorkan @sebastiankb 
I will try to look at it as well..

examples/security/oauth/README.md
```diff
@@ -0,0 +1,35 @@
...
+```bash
+npm run start:consumer
+```
+
+you should see the following line at the end of consumer log:
+```bash
+oAuth token was Ok!
+```
+This confirm that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
+to `./memory-model.js` and try to remove the string `"limited"` from the grants. Run again the example and you will see that the action is not executed and an error is returned by the client.
```
danielpeintner, line 32: There is no `"limited" in grants.
There is however `"limited" in exposer.js scopes.

I think this needs to be updated.

relu91: In the readme I am referring to this line https://github.com/eclipse/thingweb.node-wot/pull/258/files#diff-2d09f3d34f25fa2a6cd5e3712c91f8e4R24 . Should I specifically point the reader to the `memory-model.js` constructor?

danielpeintner: Strange, I checked the PR locally and I did see "password" instead of "limited" ... I might have opened an *older* file when looking at it.

Hence, fine as is. Sorry!

PR #261, opened 2020-06-24
Formatting based on .editorconfig

danielpeintner:
run `eclint fix **/*` for applying .editorconfig to entire project


packages/binding-coap/src/coap-client.ts
```diff
@@ -53,83 +53,83 @@ export default class CoapClient implements ProtocolClient {
...
+    }
+    req.end();
     });
   }
 
   public unlinkResource(form: CoapForm): Promise<any> {
     return new Promise<void>((resolve, reject) => {
 
-      let req = this.generateRequest(form, "GET", false);
+    let req = this.generateRequest(form, "GET", false);
```
danielpeintner, line 122: not sure why but eclint does strange nesting.
Code in promise should be nested...

danielpeintner:
> I agree that it is pretty massive, but I do not see any other better solution to this problem. We only need to be sure to do this once... So I think once we figure out why `eclint` does not nest correctly promises IMHO we can merge this.

If eclint does not work as expected we can do it somewhat manually with the VSC extension...

> Also, last time that I bring this up 😄 , using other linters we might enforce more strict rules like: do not use `console.log`.

Do you want to work on a proposal? 
I am fine either way...

danielpeintner:
lots of conflicting files and out-dated

PR #264, opened 2020-06-25
chore: use info logging level for server starts

danielpeintner:
allows users to see address and port


danielpeintner:
see https://github.com/eclipse/thingweb.node-wot/issues/254#issuecomment-648876734

PR #265, opened 2020-06-26
fix: issue 244, avoid logging serialized TD every time

danielpeintner:
Note: causes lots of noise in HTTP observe

caller may still to do logging if required/wanted


danielpeintner:
> Perfect! However I'd still open an issue to keep track of why `getThingDescription` method calls `serializeTD` just to parse it again.

I moved your comment to https://github.com/eclipse/thingweb.node-wot/issues/244#issuecomment-658004268

Let's merge this PR and continue discussion there

PR #268, opened 2020-06-29
Fixed feature matrix in ReadMe

erkann-sen:
A PR which adds support of SSE in HTTP binding is already in master, what I have forgot to change is, the feature matrix. 
It is done now.


sebastiankb:
thanks

danielpeintner:
Did you look at the *rendered* version, see https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-http#feature-matrix

It seems there is an issue

PR #273, opened 2020-07-03
feat: merge config with defaults

### LINKED ISSUE #254: [cli] cli-default-servient assumes that there is always a log level
Even though the log level features added by @fatadel  have been tested, we have overlooked something. In the config file, the log level was optional (as far as I remember) but now in https://github.com/eclipse/thingweb.node-wot/blob/6b207b0029654055f170f83dc94443cc8c94f956/packages/cli/src/cli-default-servient.ts#L69 the `log.level` is trying to be read. So if the config file does not have it, an error like the following is thrown:
`WoT-Servient config file error: Cannot read property 'level' of undefined`

As mentioned at https://github.com/eclipse/thingweb.node-wot/issues/177 , I think we really need a JSON Schema (and TS definitions) for the configuration file that validates it before being used in the runtime. This would also serve as a guide for those who are writing a configuration file.

fatadel:
Closes #254

### DIFF
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -59,7 +59,6 @@ export default class DefaultServient extends Servient {
         this.config = (typeof config === "object") ?
             mergeConfigs(DefaultServient.defaultConfig, config) :
             DefaultServient.defaultConfig;
-        if (!this.config.servient) this.config.servient = DefaultServient.defaultConfig.servient;
 
         // apply flags
         if (clientOnly) {
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -56,7 +56,9 @@ export default class DefaultServient extends Servient {
         super();
 
         // init config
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli-default-servient.ts
```diff
@@ -56,7 +56,9 @@ export default class DefaultServient extends Servient {
         super();
 
         // init config
-        this.config = (typeof config === "object") ? config : DefaultServient.defaultConfig;
+        this.config = (typeof config === "object") ?
+            mergeConfigs(DefaultServient.defaultConfig, config) :
+            DefaultServient.defaultConfig;
         if (!this.config.servient) this.config.servient = DefaultServient.defaultConfig.servient;
```
danielpeintner, line 62: I believe the line about `this.config.servient` can be removed since it should be handled in the line before. Am I wrong?

fatadel: No, you are right. I have overlooked that, I'd change the PR now.
Thank you!

fatadel:
@danielpeintner done

PR #283, opened 2020-07-15
Issue 277

danielpeintner:


egekorkan:
So how does the error look like now for the developer?

danielpeintner:
> So how does the error look like now for the developer?

that depends on the binding implementation and belongs to it.

The only thing this PR does is returning null (meaning no credentials set for this id) and the binding should report "forbidden" or anything else that seems appropriate if credentials are necessary for a given interaction

danielpeintner:
@relu91 you are right with your comment and that this topic relates to each binding. 
getCredentials() is called only once when setting [client security](https://github.com/eclipse/thingweb.node-wot/blob/0626e89b967e5c7d114a1a1ab8668bb5e1c13710/packages/core/src/consumed-thing.ts#L136).
`setSecurity(metadata: Array<TD.SecurityScheme>, credentials?: any): boolean;`

The assumption was since the input `credentials` is optional anyway *every* binding should deal with `undefined`/`null` already.
Not sure what we should do. Check each binding I guess?

BTW, this PR fixes one more issue w.r.t. `getThing()` call which should have been fixed in a separate PR ....

relu91:
>@relu91 you are right with your comment and that this topic relates to each binding.
>getCredentials() is called only once when setting client security.
>setSecurity(metadata: Array<TD.SecurityScheme>, credentials?: any): boolean;

>The assumption was since the input credentials is optional anyway every binding should deal with undefined/null already.
>Not sure what we should do. Check each binding I guess?

Sorry, I was thinking about server-side binding implementations. Like HTTP server that uses getCredentials(id); methods and handles correctly the undefined case. I wondered if it was better to just fail there so that we are sure to throw the correct error (i.e. credentials not found). 

https://github.com/eclipse/thingweb.node-wot/blob/70c31a4a41ed324b9b33147e170722f4219b281a/packages/binding-http/src/http-server.ts#L295-L317

Also, if the credential argument is *optional* it can either be the argument type or `undefined`. Therefore, usually, devs might check if the optional parameter is `undefined` but not if it is `null`.  TS should not even allow putting nullable values in optional function arguments.  Probably, since the `credential` is `any` TS is not complaining. 

My point is checking if the value is `null` for an optional argument is a little bit odd.  I propose to just return `undefined` and check what's going on in each binding.

danielpeintner:
> My point is checking if the value is `null` for an optional argument is a little bit odd. I propose to just return `undefined` and check what's going on in each binding.

So we would keep the code that is there right now, right?
@egekorkan is this fine by you since you created https://github.com/eclipse/thingweb.node-wot/issues/277 ?

egekorkan:
> My point is checking if the value is null for an optional argument is a little bit odd. I propose to just return undefined and check what's going on in each binding.

I think that this approach would be the best, even for now

danielpeintner:
Ok, I will withdraw  and close this PR and fix the other about `getThing()` problem in a separate PR/commit.

PR #287, opened 2020-07-16
fix: improve error handling of unknown interaction name

danielpeintner:
see https://github.com/eclipse/thingweb.node-wot/issues/285

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -201,27 +201,31 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
         return new Promise<any>((resolve, reject) => {
             // TODO pass expected form op to getClientFor()
             let tp: TD.ThingProperty = this.properties[propertyName];
-            let { client, form } = this.getClientFor(tp.forms, "readproperty", Affordance.PropertyAffordance, options);
-            if (!client) {
-                reject(new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`));
-            } else if (!form) {
-                reject(new Error(`ConsumedThing '${this.title}' did not get suitable form`));
+            if (!tp) {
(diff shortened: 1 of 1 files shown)


danielpeintner:
Note: the PR has a lot of whitespace diffs due to indentation but the actual change is purely the initial check like

```
            if (!tp) {
                reject(new Error(`ConsumedThing '${this.title}' does not have property ${propertyName}`));
            } else {
```

egekorkan:
:+1: from my side. Since this kind of code is called for every possible interaction, wouldn' t a try/catch be faster? Not sure if it matters and also if based checking is more clear in the code

danielpeintner:
> 👍 from my side. Since this kind of code is called for every possible interaction, wouldn' t a try/catch be faster? Not sure if it matters

Mhh, not sure but I guess try/catch will fail later. So I don't think it is faster

> and also if based checking is more clear in the code

Can you elaborate. I do not really understand what you mean by that...

egekorkan:
> Can you elaborate. I do not really understand what you mean by that...

I just wanted to say to that with if else, it is obvious what error or case is being tested. With try/catch it would be catching anything and assuming it is the error we are interested in.

danielpeintner:
> > Can you elaborate. I do not really understand what you mean by that...
> 
> I just wanted to say to that with if else, it is obvious what error or case is being tested. With try/catch it would be catching anything and assuming it is the error we are interested in.

Either way is fine but the error message in the PR 
`new Error(`ConsumedThing '${this.title}' does not have property ${propertyName}`)`
explicitly mentions the reason that is tested in the `if` statement .... a given interaction name is not available. 

@egekorkan how should we proceed best?

egekorkan:
We can proceed the way it is now, it was just a confusion from my side

danielpeintner:
Great!

PR #288, opened 2020-07-17
feat: improve API.md

### LINKED ISSUE #209: Updating API.md file
The API.md file that is in the root directory is not up to date with the latest API. (https://github.com/eclipse/thingweb.node-wot/blob/master/API.md).

It also does not show some features like adding security, uriVariables.

ajs124:
Fixes #209

Signed-off-by: ajs124 <eclipse@ajs124.de>

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,59 +1,112 @@
 ### What to do with the library
-
-Two main functionalities of node-wot is to creating a WoT Thing and interacting with another WoT Thing. These can be also combined to have a Thing interacts with other Things.
+The two main functionalities of node-wot are creating WoT Things and interacting with other WoT Things.
+These can be combined into a Thing that interacts with other Things.
 
 #### Creating a WoT Thing
-Creating a WoT Thing is called exposing a Thing. Exposing a Thing creates a Thing Description that can be used to by others to interact with this Thing.
+Creating a WoT Thing is called exposing a Thing.
+Exposing a Thing creates a Thing Description that can be used to by others to interact with this Thing.
 
-* Exposing a Thing:
+##### Starting a Servient
+```javascript
+WotCore = require("@node-wot/core")
+let servient = new WotCore.Servient();
+let WoT = await servient.start();
+```
 
+##### In Client mode, add factories
+```javascript
+WotCore = require("@node-wot/core")
+BindingHttp = require("@node-wot/binding-http")
+BindingCoap = require("@node-wot/binding-coap")
+let servient = new NodeWoTCore.Servient();
+servient.addClientFactory(new BindingHttp.HttpClientFactory())
+servient.addClientFactory(new BindingCoap.CoapClientFactory())
+```
+The different bindings offer client factories.
+These need to be added in order to be able to access devices through this protocol.
+
+For more details on bindings, e.g. configuration options for a specific `*ClientFactory`, look at the `README.md` files in their respective directories.
+
+##### In Server mode, add servers
+```javascript
+WotCore = require("@node-wot/core")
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -65,89 +118,88 @@ thing.setPropertyReadHandler(
...
+  });
+});
   ```
 Here, you see also how to access the properties of a Thing you are creating.
 
-* Add an Event definition to the Thing.
-
-The Event Interaction Pattern describes event sources that asynchronously push messages. Here not state, but state transitions (events) are communicated (e.g., "clicked"). Events may be triggered by internal state changes that are not exposed as Properties. Events usually follow strong consistency, where messages need to be queued to ensure eventual delivery of all occured events.
+##### Add an Event definition to the Thing
+The Event Interaction Pattern describes event sources that asynchronously push messages.
```
egekorkan, line 175: Affordance instead of pattern

API.md
```diff
@@ -171,59 +223,64 @@ setInterval( async () => {
...
-  
-  You can read the property values with the `read` function. It is an asynchronous function that will take some time to complete. So you should handle it explicitely. Here we use the await functionality of node.js.
 
-  ```javascript
-  let read1 = await thing.properties.count.read();
-  console.info("count value is", read1);
-  ```
-  * Set the value of a Property or a set of properties.
+##### Read the value of a Property or set of properties
+You can read the property values with the `read` function.
```
egekorkan, line 249: with the readProperty() function

API.md
```diff
@@ -171,59 +223,64 @@ setInterval( async () => {
...
+It is an asynchronous function that will take some time to complete.
+So you should handle it explicitely.
+Here we use the await functionality of node.js.
```
egekorkan, line 252: Node.js

API.md
```diff
@@ -171,59 +223,64 @@ setInterval( async () => {
...
+let read1 = await thing.readProperty("count");
+console.info("count value is", read1);
+```
 
-  ```javascript
-thing.properties.color.write({ r: 255, g: 255, b: 0 } );
-  ```
-  <!-- * Observe value changes of a Property. -->
+##### Set the value of a Property or a set of properties
+You can write to a property by using the `write` function.
```
egekorkan, line 260: writeProperty

API.md
```diff
@@ -65,89 +118,88 @@ thing.setPropertyReadHandler(
...
-  "brightness",
-  (value : any) => {
-    return new Promise((resolve, reject) => {
-      setBrightness(value);
-      resolve(value);
-    });
+thing.setPropertyWriteHandler('brightness', (value) => {
+  return new Promise((resolve, reject) => {
+    setBrightness(value);
+    resolve(value);
```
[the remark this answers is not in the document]
ajs124: I used modulo 2 as an example and added a sentence explaining it.

egekorkan:
So mostly good, there was an incorrect review from my side, otherwise there is the `write()` method that is still remaining

ajs124:
> incorrect review from my side

What? Where? Did I incorrect the thing as well or just ignore and forget it?

> there is the write() method that is still remaining

Oops, looks like I only changed the text but not the code.

egekorkan:
> What? Where? Did I incorrect the thing as well or just ignore and forget it?

I just wanted to say that I corrected name to title. However, that name was for the SW object and not the Thing. It can be reverted back to name. See https://github.com/eclipse/thingweb.node-wot/pull/288#discussion_r461754147

egekorkan:
Looks good! Could be merged from my side :)

PR #292, opened 2020-07-21
[binding-http] fix 290 basic authentication

relu91:
Basic credentials now correctly set authorization header. I also provided a basic (no pun intended) JUnit test to capture

### DIFF
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -40,7 +40,7 @@ export class BasicCredential extends Credential{
     }
     async sign(request:Request){
         let result = request.clone()
-        result.headers.set("authorization",Buffer.from(this.username + ":" + this.password).toString('base64'))
+        result.headers.set("authorization","Basic "+Buffer.from(this.username + ":" + this.password).toString('base64'))
         return result
     }
 }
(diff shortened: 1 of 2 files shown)


danielpeintner:
Thanks, looks good. Test is very useful.

@relu91 do you think we can add another test that fails with *invalid* credentials

relu91:
> @relu91 do you think we can add another test that fails with invalid credentials

Right, just pushed 😃

ajs124:
Any ideas what travis-ci doesn't like about this? The tests run just fine for me, locally.

danielpeintner:
> Any ideas what travis-ci doesn't like about this? The tests run just fine for me, locally.

Mhh, all instances fail with `FetchError: request to https://127.0.0.1:3001/ failed, reason: connect ECONNREFUSED 127.0.0.1:3001` for test "HTTP auth basic client implementation should authorize client with basic"

I am quite sure it worked fine before the last commit.
There used to be a re-run button in Travis but it does not show it any longer. I think we need to add yet another *ping* commit to rerun travis ..

relu91:
I noticed that in some test servers are not correctly closed after the test suite is executed. It might make random failures like this one. I'll add a fix commit to this PR. @danielpeintner is it ok?

danielpeintner:
Mhh, 2 out of 4 Travis builds failed again :-( 
I just fear that IF we simply merge it will not get any better... since it is always the same test that fails there must be anything...

Somehow the port is still in use

PR #294, opened 2020-07-21
#289 update README

### LINKED ISSUE #289: [documentation] Dependency management in package.json
I am trying to use ``@node-wot/core`` to run an "ordinary" Node.js module, in which the dependencies are managed in a package.json file.
However, it seems that when running ``wot-servient`` the script does not pick up locally installed dependencies, but requires me to install all dependencies of my project globally.
Presumably, the solution can be to install ``@node-wot/core`` locally and invoke ``wot-servient`` locally as well.
How can this be achieved?
IMHO, it is important to document this as the standard way of working (instead of having to clone this GitHub repo as a starting point, which is a rather "hacky" workflow).

TimKam:
To document how to install node-wot tools as local dependencies that in turn can run scripts using other local dependencies.
Closes #289

@relu91 please check if this is according with how node-wot is intended to be used.
If not, changes may be necessary to properly support the management of local dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,33 +63,74 @@ Using a browser with only ES5 support (eg. IE 11) might be possible if you add p
 ## How to get the library
 ### As a Node.js dependency
 
+You can install node-wot in the following ways:
+
+1. As a normal dependency (i.e., like loadsh). In this case, you are embedding a servient inside your application.
+2. As a CLI to run scripts. In this case, your application is running inside
+the default servient.
+
+#### Normal Dependency
+
 If you want to use node-wot as a library in your Node.js application, you can use npm to install the node-wot packages that you need. To do so, `cd` inside you application folder, and run:
 
 ```
-npm i @node-wot/core @node-wot/binding-coap @node-wot/cli --save
+npm i @node-wot/core @node-wot/binding-coap --save
 ```
 
-This requires that you have already initiated a project with `npm init`.
-Now, add the following script to your `package.json` so that you can later execute your scripts using the
-`npm run start` command:
+Now, you can implement your node-wot entry point, e.g., `main.js` as follows:
 
+```JavaScript
+// Required steps to create a servient
+const Servient = require('@node-wot/core').Servient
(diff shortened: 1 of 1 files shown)


egekorkan:
Does the `wot-servient` command become available on the terminal if nothing is installed globally?

TimKam:
I tried it out and uninstalled all wot-node dependencies globally. I cannot successfully call `wot-servient` or `node wot-servient` from the command line, but calling `wot-servient` through the package's scripts property (via `npm run start`) works. This makes sense after I've read the documentation at https://docs.npmjs.com/misc/scripts: _"If you depend on modules that define executable scripts, like test suites, then those executables will be added to the PATH for executing the scripts"_.

relu91:
> I tried it out and uninstalled all wot-node dependencies globally. I cannot successfully call wot-servient or node wot-servient from the command line, but calling wot-servient through the package's scripts property (via npm run start) works. This makes sense after I've read the documentation at https://docs.npmjs.com/misc/scripts: "If you depend on modules that define executable scripts, like test suites, then those executables will be added to the PATH for executing the scripts".

Yes, it is the intended behavior. However, I think that it should work even when the CLI is installed globally. My guesses are that when installed globally the module requires paths points to the global directory and not to the local folder (maybe security reasons?). I'll open an issue and start to work on a possible fix.

TimKam:
Okay, I can adjust the PR, but then I'd like to add that only the _local_ approach implies that the package's dependencies are available; otherwise, all dependencies have to be installed globally. I'd expect this is an important differentiator, because it means the the standard dependency management workflow does not apply when having the CLI globally installed. Okay?

relu91:
yes, sure! You may also add that it is a workaround and in the future, node-wot should support local dependencies.

TimKam:
I gave it another try/another round of edits!

TimKam:
Fixed a typo (changed one word)

danielpeintner:
Thanks @TimKam and @relu91

PR #186, opened 2020-03-06
Modbus client protocol binding - First Candidate

relu91:
This PR introduces another protocol binding to thigweb.node-wot. I made some experience with Modbus and I thought that this repo could benefit from this. I followed the guidelines of the other implementations and I made some tests to verify the correct behavior of the library.

Modbus is a pretty old protocol but it is quite well spread in the industry, I am using this URL syntax as form href: "modbus://address:port". I saw that it already used around but I cannot find a formal definition of modbus:// prefix. Anyway, it should work for this purpose.

Issue #267, opened 2020-06-29
Suggestion about adding new protocol bindings/consumers


erkann-sen:
Hi everyone,
I am currently writing my master thesis, which is about interoperability of protocols/systems with wot.
Till now, I made [Mozilla webthings](https://github.com/erkann-sen/wot-mozilla) and [openHAB](https://github.com/erkann-sen/wot-openhab/) consumable with node-wot with different methods.
What I wish is to somehow bring this functionality to node-wot library. 
But I know that, adding a binding or helper for every platform makes the main library fatter and harder to maintain and also creates an illusion that for every platform there is another binding needed, which makes node-wot library seem hard to use and there is a discussion about what to make with Oracle and Fujitsu bindings.

When I am working with openHAB, I found their solution to this system, can be also useful for node-wot.
They have two different but connecting repos. A core repo, which has the main functionality of the system and ana add-ons repo which has several bindings or helpers for other systems.

I suggest we should have a main node-wot library, which has only the main bindings like MQTT, HTTP etc.
In another repo, we can have vendor specific bindings and helpers, all has their own readme file, which explains how to use. In this way it is also encouraged to write people more bindings to make other protocols work easily.

Advantages that I can think of are:

1. Easily readable and less complex main code
2. Easily maintainable main repo.
3. Adding another bindings are easier because it does not mess up with the main repo, which can increase the adoption of the system 
4. The discussion about what to do with vendor specific bindings will end

relu91:
I totally support this idea. I have been tinkering for a while about the possibility to move node-wot to a full plug-in based architecture. In my mind, not only protocol bindings should be provided as add-ons but also security schemas implementations, content-encoders/decoders, and (maybe) system APIs. As you mentioned there are quite a few benefits but also, I want to add, that this pattern is well-known in the node/JS community. Most of the big libraries are plug-in based like Vue.js, Jquery, Rollup... (just to mention a few). My personal opinion is that this open plugin architecture contributed a lot to the wide success of this open-source JS libraries.  

Furthermore, this open plug-in architecture will cope with the inherent complexity of Thing Descriptors. As @mlagally mentioned during different calls, it might even be impossible to build a client that is able to correctly understand ALL the possible TDs. That's why WoT Profiles are been explored and if they are adopted we could even tag different plug-ins to be compliant to one or more WoT profile.  

[...]

PR #298, opened 2020-07-31
Optimize docker build

relu91:
This PR provides two contributions:
1. fixes the docker build file. After #186 docker builds failed due to missing system dependencies. Now it correctly install also g++, linux-headers and udev (serial-port requires them for building).   
2. it reduces the size of the final image from about 650 Mb to 140 Mb.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.16.1-alpine3.11
+FROM node:12.16.1-alpine3.11 as BUILD
 
 RUN apk add --no-cache \
 	gcc \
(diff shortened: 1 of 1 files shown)


danielpeintner:
Thanks @relu91!

>     2\. it reduces the size of the final image from about 650 Mb to 140 Mb.

That sounds great!

So far I did not really use the docker containers. Who can quickly cross-check the update?

relu91:
I think @hyperloris is a good candidate for cross-checking.

danielpeintner:
@hyperloris may I ask you to do so... thanks!

hyperloris:
I tried the new Dockerfile and it works perfectly. You did a great job @relu91.

I noticed you're not copying all the bindings, just the "main ones". I was wondering if it wouldn't be better to add an ARG in order to specify a list of bindings you want to add to the image (maybe with a default). What do you think?

relu91:
> I noticed you're not copying all the bindings, just the "main ones". I was wondering if it wouldn't be better to add an ARG in order to specify a list of bindings you want to add to the image (maybe with a default). What do you think?

You noticed right. I think it is a great idea. Unfortunately, the default servient actually requires those dependencies and it cannot be customized. After we'll address #267, we can then change the docker build to have some ARG parameter.

danielpeintner:
@hyperloris thanks for taking a look!
@relu91 thanks for the work!

PR #303, opened 2020-08-11
feat: add readme to file binding

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/284


danielpeintner:
@egekorkan do you expect more information?

egekorkan:
I think another feature to show would be reading (fetch with helpers) a TD that is stored as a file but otherwise this is very good :)

danielpeintner:
> I think another feature to show would be reading (fetch with helpers) a TD that is stored as a file but otherwise this is very good :)

I added a second example showing how to read a local TD file....

PR #309, opened 2020-08-14
Correct websocket implementation status

relu91:
Currently, the WebSocket client does not have any functionality implemented. Therefore, I changed the main readme to reflect the real status.


egekorkan:
I would even say that no subprotocol is implemented, thus all messages are piped on different sockets. I think if we say that ws is implemented, we need to specify how it is implemented, otherwise it is like saying that we implemented TCP :)

danielpeintner:
I agree

Issue #190, opened 2020-03-19
How to use environment variables in WoT script?


philiptran:
Our WoT script needs to access some configuration via environment variables. When using the _process.env_ to get the environment variable, the following error is thrown:

```
const API_URL = process.env.API_URL
ReferenceError: process is not defined
```

danielpeintner:
Question: How do you use node-wot?
* via the CLI interface (calling `wot-servient` / `node packages\cli\dist\cli.js`) OR
* via NPM package dependency (`npm install @node-wot/core` etc.)

philiptran:
I use it via NPM package dependency and run the script using `node @node-wot/cli/dist/cli.js counter.js`. The process.env is being used in the `counter.js` script.

danielpeintner:
> I use it via NPM package dependency and run the script using `node @node-wot/cli/dist/cli.js counter.js`. The process.env is being used in the `counter.js` script.

Mhh, this reads a bit contradictory to me.

You should either
1. use CLI without *explicit* NPM package dependency (CLI brings it own dependencies with it) and run `node packages\cli\dist\cli.js examples\scripts\counter.js`
 OR
2. Use NPM package dependency and run simply run `node sample-script.js` (e.g., [HTTP](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-http#http-protocol-binding-of-node-wot))

relu91:
Anyway, I've just checked and it seems that environment variables are not exported to CLI scripts. I am opening an issue. https://github.com/eclipse/thingweb.node-wot/blob/b428faae483b3a8dce937a02ce8255365e8b71b3/packages/core/src/servient.ts#L104-L141

philiptran:
Thanks to today's webex call that we have a better idea on using node-wot as npm dependency. However, we are using the CLI in an independent project without cloning the node-wot code repository. The root cause of this issue is the `process` is not exported to CLI script as @relu91 has rightly pointed out.

Attached is a sample project that demonstrates our usage:

[test-wot.zip](https://github.com/eclipse/thingweb.node-wot/files/4354645/test-wot.zip)

relu91:
For the record, I don't know if it is the case to export the entire process object to scripts...  Are you using `process` for something else? 😀

philiptran:
[...]

Issue #275, opened 2020-07-07
[core] Default servient does not pass CLI arguments to scripts


relu91:
This issue is related to #190 and #192. Right now it is not possible to read command-line arguments from WoT scripts since the `process` object is not forwarded to VM context. In principle, we could preprocess `process.args` and extract only arguments after the script names. For example in the following line:
```bash
wot-servient  consumer.js producer.js -- arg1 arg2 -option1 
```
we can pass `arg1`, `arg2`, and `-option1` to the VM context of `consumer.js` and `producer.js`. IMHO, this can be done also for *unprivileged* scripts.

Issue #192, opened 2020-03-19
Export environment variables to WoT scripts


relu91:
As mentioned in #190 environment variables are currently not exported to the WoT script context. I guess they might be exported only to privilege scripts since there could be system-specific information there.

Issue #191, opened 2020-03-19
How to import utils script from the same folder of WoT script?


philiptran:
We need to import some utility script from the same folder of WoT script using the `require` function. But the local script file can not be resolved correctly.

## To reproduce

The counter.js is copied to a separate folder and to be run using @node-wot CLI as `As a Node.js dependency` mode.

1. Create config.js file in the same folder of the WoT script with the following content:

```
const config = {
    API_URL: 'https://localhost/api/',
    API_AUTHORIZATION: 'xyz'
};
module.exports = config;
```

1. Edit counter.js to add the following:

```
const config = require("./config");
```

1. Run the `counter.js` script using @node-wot CLI, the following error is thrown.

```
Servient caught error in privileged 'C:\wot\counter.js' and halted with internal/modules/cjs/loader.js:796
    throw err;
    ^

Error: Cannot find module './config'
Require stack:
...
```

philiptran:
A workaround is to import the `config.js` as follow.

```
const path = require('path');
const config = require(path.resolve('./config'));
```

egekorkan:
For similar issues I had before, this is due to the fact that CLI runs the script but the directory it uses is actually its own. You can run process.cwd() to understand where it is running from.

Another workaround would be specifying the full path but this is a dirty solution.
Could you try to use the scriptDir option? (I am not very sure about it though)
```json
{
    "servient": {
        "scriptDir": AUTORUN,
        "scriptAction": RUNSCRIPT
    }
}
```
  AUTORUN is a path string for the directory to load at startup
  RUNSCRIPT is a boolean indicating whether to provide the 'runScript' Action

philiptran:
Yes, it was trying to run `process.cwd()` to figure out the current working directory but the `process` object is not exported to WoT script hence I did the workaround using `path.resolve` as above. 

Please see issue #190 for issue with `process` object.

danielpeintner:
@relu91 solved by https://github.com/eclipse/thingweb.node-wot/pull/315 ?

relu91:
Just tested, yes solved!

PR #315, opened 2020-08-27
Upgrade to vm2

### LINKED ISSUE #314: [core][cli] Upgrade to vm2
Currently, we are using the node `vm` module to load and execute WoT applications. However, `vm` has different problems:
1. it is not completely "safe". [Here](https://github.com/patriksimek/vm2#vm2-----) is explained how a malevolent script can access the main context and kill the servient. 
2. it is difficult to make `require` works. As discussed in different issues/PRs is not possible to require local modules or scripts without heavy modification to the current code. Modifications that will end up very similar to what is already implemented in vm2.

So my suggestion is to upgrade to [vm2](https://github.com/patriksimek/vm2) which has interesting features  (from the readme):
* Runs untrusted code securely in a single process with your code side by side
* Full control over sandbox's console output
* Sandbox has limited access to process's methods
* Sandbox can require modules (builtin and external)
* You can limit access to certain (or all) built-in modules (*currently is not working in our servient CLI*)
* You can securely call methods and exchange data and callbacks between sandboxes
* Is immune to all known methods of attacks
* Transpilers support (*so we may support TS from the cli without the building step*)

What do you think? I can volunteer to make this happen.

relu91:
This PR closes #314. As described in the issue vm2 has a couple of benefits. In particular, it will solve problems for `require` calls in WoT application scripts ( #191). With subsequent PRs we could close issues #190 #192 #275, without major modifications. Probably also #185 might be resolved thanks to the `transpilers` support of vm2.

### DIFF
--- a/packages/browser-bundle/package.json
+++ b/packages/browser-bundle/package.json
@@ -22,6 +22,6 @@
   "dependencies": {
   },
   "scripts": {
-    "build": "browserify index.js --plugin tinyify -o dist/wot-bundle.min.js"
+    "build": "browserify index.js --plugin tinyify --external coffee-script -o dist/wot-bundle.min.js"
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,13 +24,13 @@
     "typescript-standard": "0.3.36"
   },
   "dependencies": {
+    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.2",
-    "@types/uritemplate": "0.3.4",
+    "uuid": "3.4.0",
     "rxjs": "5.5.11",
     "uritemplate": "0.3.4",
-    "uuid": "3.4.0",
-    "vm2": "^3.9.2",
-    "wot-typescript-definitions": "0.7.3"
+    "@types/uritemplate": "0.3.4",
+    "vm2": "^3.9.2"
   },
   "scripts": {
     "build": "tsc",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,12 +24,13 @@
     "typescript-standard": "0.3.36"
   },
   "dependencies": {
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.2",
-    "uuid": "3.4.0",
+    "@types/uritemplate": "0.3.4",
     "rxjs": "5.5.11",
     "uritemplate": "0.3.4",
-    "@types/uritemplate": "0.3.4"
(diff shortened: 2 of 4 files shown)


packages/core/package.json
```diff
@@ -24,12 +24,13 @@
...
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.2",
-    "uuid": "3.4.0",
+    "@types/uritemplate": "0.3.4",
     "rxjs": "5.5.11",
     "uritemplate": "0.3.4",
-    "@types/uritemplate": "0.3.4"
+    "uuid": "3.4.0",
+    "vm2": "^3.9.2",
+    "wot-typescript-definitions": "0.7.3"
```
danielpeintner, line 33: May I ask you to introduce back the original order since most of the entries are just re-shuffled and not *new* expect "vm2": "^3.9.2"

The reason I am asking for it is that if we merge THIS work into https://github.com/eclipse/thingweb.node-wot/tree/v0.8.x we will get merge problems since this will be detected as a CHANGE.

packages/browser-bundle/package.json
```diff
@@ -22,6 +22,6 @@
   "dependencies": {
   },
   "scripts": {
-    "build": "browserify index.js --plugin tinyify -o dist/wot-bundle.min.js"
+    "build": "browserify index.js --plugin tinyify --external coffee-script -o dist/wot-bundle.min.js"
```
danielpeintner, line 25: Is this change needed?

danielpeintner (changes requested):
Did some sanity checks/runs and everything seems to work just fine. Great work, thanks!

I added 2 comments/requests....

packages/core/package.json
relu91: yes sure I'll try to put everything back in order. It was npm that changed the order.

packages/browser-bundle/package.json
relu91: Yes, vm2 has as *optional* dependency coffee-script. However, broswerify does not detect it as optional and tries to find the module. The --external argument just tells to broswerify to ignore that module since it will be added at runtime by users of the library (although I do not think this will ever happen in our case).

danielpeintner: I see. Let's keep `--external coffee-script`. I was just not sure *why* it is there

packages/core/package.json
relu91: It should be fine now 👍

danielpeintner (approved):

PR #316, opened 2020-08-28
Introduce WoT scripts command line args

### LINKED ISSUE #275: [core] Default servient does not pass CLI arguments to scripts

relu91:
This PR is based on #315 so I put this as a draft until #315 is merged. It closes #275 introducing the ability to pass command-line arguments to WoT Scripts. I choose the same syntax used by npm and other CLIs. Basically, to recognize which arguments should be passed and which argument should be used by the CLIs I used `--` as a separator. For example in the following command line:
```bash
wot-servient -c config.js script1.js script2.js -- -option arg1 arg2
```
`-c`, `config.js`, `script1.js`, and script2.js are interpreted by our CLI package whereas `-option`, `arg1`,and arg2 can be read by **both** script1.js and script2.js.

This means that scripts can have the same argument without conflicts: 
```bash
wot-servient script1.js script2.js -- -c config.js
```
Node-wot cli does not read the config.js file.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -28,6 +28,8 @@ const baseDir = ".";
 var clientOnly: boolean = false;
 
 var flagArgConfigfile = false;
+var flagScriptArgs = false;
+var scriptArgs:Array<string> = [];
 var confFile: string;
 
 interface DebugParams {
@@ -69,7 +71,7 @@ const runScripts =async function(servient: DefaultServient, scripts: Array<strin
                     // limit printout to first line
                     console.info("[cli]",`WoT-Servient running script '${data.substr(0, data.indexOf("\n")).replace("\r", "")}'... (${data.split(/\r\n|\r|\n/).length} lines)`);
                     fname = path.resolve(fname)
-                    servient.runPrivilegedScript(data, fname);
+                    servient.runPrivilegedScript(data, fname,{argv:scriptArgs});
                 }
             });
         });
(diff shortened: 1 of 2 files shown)


packages/core/package.json
```diff
@@ -24,12 +24,13 @@
...
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.2",
-    "uuid": "3.4.0",
+    "@types/uritemplate": "0.3.4",
     "rxjs": "5.5.11",
     "uritemplate": "0.3.4",
-    "@types/uritemplate": "0.3.4"
+    "uuid": "3.4.0",
+    "vm2": "^3.9.2",
+    "wot-typescript-definitions": "0.7.3"
```
danielpeintner, line 33: FYI: This PR has the same changes as https://github.com/eclipse/thingweb.node-wot/pull/315

relu91: thanks, I'll fix it and then this PR is ready to be merged

danielpeintner: Great, thanks! Please let us know once you think it is ready

relu91:
It is ready to be merged now 👍

danielpeintner (approved):
Looks good!

There are some minor formatting issues but there is an outstanding issue about formatting anyway...

Issue #313, opened 2020-08-23
Slugify title in URLs


fatadel:
From Telegram discussion:
If we use slagify then the thing titles containing spaces will be automatically converted to the ones using dashes not `%20`.
So that, `Smart Coffee Machine` becomes `smart-coffee-machine` (note lowercase as well) and we are not obliged to use dashes in titles for readability in URLs.

danielpeintner:
I like the proposal to use more *readable* URLs.

The only downside that comes to my mind is that at the moment there is a 1-to-1 mapping from title to URL. This is going to fail with the proposed change:
`"title": "Smart Coffee Machine"`
and
`"title": "Smart-Coffee-Machine"`
and many more variante lead to the same url
`smart-coffee-machine`

fatadel:
I agree with you, @danielpeintner.
But to overcome the downside you've mentioned we can simply log the generated URL.
If we do it on `console.log` level we can be sure that this message never gets suppressed and the user knows where to look for the exposed Thing.

egekorkan:
Yes but log messages get lost easily among other messages. Not objecting this need but only 90% convinced on the solution.

danielpeintner:
An alternative is to make sure that by means of slagify we never get duplicates.

Having said that, a second *thing* getting the same URL would need to 
a) get another URL (and we need to store that URL internally) or
b) being refused to be exposed due do conflicts

fatadel:
I think keeping the list of URLs and appending some suffix to conflicting ones is better than refusing to expose and making the user change the Thing title.

relu91:
> I think keeping the list of URLs and appending some suffix to conflicting ones is better than refusing to expose and making the user change the Thing title.

IMHO this behavior can be easily implemented by the developer. I think failing should be the best option. It is better than falling silently and change the title.

 > But to overcome the downside you've mentioned we can simply log the generated URL.
If we do it on console.log level we can be sure that this message never gets suppressed and the user knows where to look for the exposed Thing.

A link to the generated TD might be still nice. However, no `console.log` this message is clearly a debug message to help developers test their generated Web Things. I see it hardly useful in production.

danielpeintner:
> > I think keeping the list of URLs and appending some suffix to conflicting ones is better than refusing to expose and making the user change the Thing title.
> 
> IMHO this behavior can be easily implemented by the developer. I think failing should be the best option. It is better than falling silently and change the title.

[...]

PR #317, opened 2020-09-03
Slugify url

fatadel:
This PR closes issue #313.


packages/binding-coap/package.json
```diff
@@ -25,11 +25,12 @@
...
   "dependencies": {
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/core": "0.7.2",
-    "rxjs": "5.5.11",
     "coap": "0.23.1",
-    "node-coap-client": "1.0.2"
+    "node-coap-client": "1.0.2",
+    "rxjs": "5.5.11",
+    "slugify": "^1.4.5",
+    "wot-typescript-definitions": "0.7.3"
```
danielpeintner, line 33: Please keep the original entries as is....

packages/binding-websockets/package.json
```diff
@@ -26,9 +26,10 @@
     "typescript-standard": "0.3.36"
   },
   "dependencies": {
-    "@node-wot/td-tools": "0.7.2",
-    "@node-wot/core": "0.7.2",
     "@node-wot/binding-http": "0.7.2",
+    "@node-wot/core": "0.7.2",
+    "@node-wot/td-tools": "0.7.2",
+    "slugify": "^1.4.5",
```
danielpeintner, line 32: Please keep the original entries as is....

packages/binding-coap/package.json
fatadel: Sorry, I just did `npm i slugify`, and apparently npm automatically sorted the dependencies.

packages/binding-websockets/package.json
fatadel: Sorry, I just did `npm i slugify`, and apparently npm automatically sorted the dependencies.

packages/binding-coap/package.json
relu91: yeah, that's quite annoying... It happened also to me when I added vm2. I wonder if we can mitigate this automatic behavior by simply asking npm to order the dependencies and make a dedicated PR.

danielpeintner: Yes, @relu91 run into the same issue and automatic sorting makes sense. The reason why I am asking is that such changes make it difficult to merge future updates that moved on to another version and cause conflicts ...

fatadel: IMHO we could also sort the dependencies once and then just add to the contribution instructions to keep the alphabetical order in case anyone adds a new dependency.

fatadel:
I think the PR is ready for review now.
Please also see the PR https://github.com/eclipse/thingweb.website/pull/22 when this one is merged.

danielpeintner:
I think the changes itself look OK.

@sebastiankb @egekorkan @relu91 what do you think. This may break some examples "out there". Shall we do such a change?
It seems the *right* thing to do but also quite late in the game now.. still not 100% sure

packages/examples/src/security/oauth/consumer.ts
```diff
@@ -18,7 +18,7 @@ import { Helpers } from "@node-wot/core";
 let WoT: WoT.WoT;
 let WoTHelpers: Helpers;
 
-WoTHelpers.fetch("https://localhost:8080/OAuth").then(td => {
+WoTHelpers.fetch("https://localhost:8080/oauth").then(td => {
```
relu91, line 21: Probably this should be changed also in the root example folder.

relu91 (commented):
In general, I think this is needed. However, as Daniel said we should pay extra-attention to update all the examples and make sure that they work with these new URLs. Since this is a breaking change in my opinion we should bump the version once it is merged. Other than this I think it is fine.

packages/binding-http/test/http-server-test.ts
```diff
@@ -160,6 +161,7 @@ class HttpServerTest {
     await httpServer.start(null);
     
     let testThing = new ExposedThing(null);
+    testThing.title = "Test";
```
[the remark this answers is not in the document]
fatadel: Because otherwise, you end up with `slugify(undefined)` which will throw an exception. And also, according to https://www.w3.org/TR/wot-thing-description/#thing, the `title` field is mandatory.
Do you think there should be a different solution for that?

packages/examples/src/security/oauth/consumer.ts
fatadel: I think I already did it if you mean `examples/security/oauth/consumer.js` file.

fatadel:
> I think there are many more updates needed once this is merged.

I've covered all the URLs in thingweb.node-wot and thingweb.website repositories. If anyone gives me a list of projects where these changes might affect, I can provide appropriate PRs there as well.

fatadel:
> > I think there are many more updates needed once this is merged.
> 
> I've tried covering all the URLs in thingweb.node-wot and thingweb.website repositories. If anyone gives me a list of projects where these changes might affect, I can provide appropriate PRs there as well.

Apparently, I've missed some hardcoded URLs but I've fixed them now in the latest commit.
Now, I think that must be indeed *all* URLs.

packages/binding-http/test/http-server-test.ts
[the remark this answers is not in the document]
relu91: Ah ok, there was a time where the title was not mandatory. So yes I think is fine as it is 👍 .

relu91 (approved):

egekorkan (approved):

danielpeintner (approved):

PR #319, opened 2020-09-07
[CLI/Core] introduce environment variable support

### LINKED ISSUE #192: Export environment variables to WoT scripts

relu91:
Closes #192. Notice that this PR is based on  #318. So it will stay in the draft mode until the other is merged.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -85,11 +85,26 @@ wot-servient.conf.json fields:
   TOKEN           : string for providing a Bearer token
   USERNAME        : string for providing a Basic Auth username
   PASSWORD        : string for providing a Basic Auth password
+  ---------------------------------------------------------------------------
+ 
+Environment variables must be provided in a .env file in the current working directory. 
+
+Example:
+VAR1=Value1
+VAR2=Value2
 
 ```
 
 ### Prerequisites
 See instructions [how to build node-wot as a standalone application](https://github.com/eclipse/thingweb.node-wot/#as-a-standalone-application).
+
+### Environment variables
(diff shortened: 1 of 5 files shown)


danielpeintner (approved):

danielpeintner:
@egekorkan can you take a look also?

egekorkan:
I have tried and the doing the following did not work:
1. I took the counter example and added the following line after line 127 that says ` Produced ...`
`console.log(process.env.PORT);`
2. I ran the script like the following
`PORT=9000 node ../../packages/cli/dist/cli.js counter.js`
3. I get an output of undefined. However, I can log process.env and prints and empty object, so something definitely works.

relu91:
Environment variables should be defined inside a `.env` file. Just put `PORT=9000` in your current directory `.env` file and it should work.  Looking around it seems a common practice and it has the benefit to not expose EVERY single variable to servient scripts.

egekorkan:
It does :heart_eyes: Can be merged

danielpeintner:
> Environment variables should be defined inside a `.env` file. Just put `PORT=9000` in your current directory `.env` file and it should work. Looking around it seems a common practice and it has the benefit to not expose EVERY single variable to servient scripts.

I wonder whether we should add a description/note saying so...
@relu91 what do you think?

Having said that, I am fine with merging "as is" also.

relu91:
> I wonder whether we should add a description/note saying so...
> @relu91 what do you think?
> Having said that, I am fine with merging "as is" also.

Yeah, I agree. I'm adding some documentation to describe this.

relu91:
Pushed 👍 check if it is sufficient.

PR #321, opened 2020-09-09
[CLI/core] custom compilers

relu91:
This PR add supports for custom compilers in the Servient runtime. Currently, I tested only `ts-node` but in principle also others might work. I managed to run this simple script inside `node-wot`:
```ts
/// <reference types="../types/wot" />

WoT.produce({}).then( (thing:WoT.ExposedThing) => {
    console.log(thing.getThingDescription().id)
})
```
to run it: `npm i ts-node && wot-servient --compiler ts-node main.ts` 

**Notice**: I used wam typescript project template (i.e. custom type definitions) and I needed to use `/// <reference types="../types/wot" />` (see point 2 below). 

.However, there still a couple of open points:
1. `cli.ts` is getting pretty messy. I think It wasn't originally designed to be a simple interface, now it is evolved in a quite complex command-line tool. Probably it needs a refactor. 
2. typescript examples still not work. The problem is how ts-node handles types (see [here](https://www.npmjs.com/package/ts-node#help-my-types-are-missing)) and how we defined our API type definitions. There still an open issue on the [scripting-api repo](https://github.com/w3c/wot-scripting-api/issues/215) so I did not go ahead and I'll wait until we make a final decision there.
3. if this PR will be merged, PR #240 is obsolete.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -23,6 +23,9 @@
     "typescript": "3.7.5",
     "typescript-standard": "0.3.36"
   },
+  "optionalDependencies": {
+    "ts-node": "^8.6.2"
+  },
   "dependencies": {
     "@node-wot/core": "0.7.2",
     "@node-wot/binding-coap": "0.7.2",
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -28,6 +28,8 @@ const baseDir = ".";
 var clientOnly: boolean = false;
 
 var flagArgConfigfile = false;
+var flagArgCompilerModule = false;
+var compilerModule:string;
 var flagScriptArgs = false;
 var scriptArgs:Array<string> = [];
 var confFile: string;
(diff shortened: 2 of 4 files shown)


danielpeintner:
@relu91 thank you for your work!

Some thoughts below

> * typescript examples still not work. The problem is how ts-node handles types (see [here](https://www.npmjs.com/package/ts-node#help-my-types-are-missing)) and how we defined our API type definitions. There still an open issue on the [scripting-api repo](https://github.com/w3c/wot-scripting-api/issues/215) so I did not go ahead and I'll wait until we make a final decision there.

This is sad :-(

> * if this PR will be merged, PR #240 is obsolete.

I think the overall issue https://github.com/eclipse/thingweb.node-wot/issues/171  is just half-solved. I think there are more aspects:
1. allow loading transpiled TS files directly  -> solved
2. Unfortunately the _original   TS file is not supported (which would be the ultimate/ideal solution)
3. (related to 2.) we still should have an automated approach to transpile TS to JavaScript

Anyway I am OK with merging this improvement.
@sebastiankb @egekorkan  do you have any concerns doing so?

danielpeintner (approved):

relu91:
>This is sad :-(

I know 😢 , I did my best to solve the issue but, finally, I didn't have the time to find a good solution. 

>> * if this PR will be merged, PR #240 is obsolete.
>
>I think the overall issue https://github.com/eclipse/thingweb.node-wot/issues/171 is just half-solved. I think there are more >aspects:
>
>1. allow loading transpiled TS files directly -> solved
>
>2. Unfortunately the _original TS file is not supported (which would be the ultimate/ideal solution)
>
>3. (related to 2.) we still should have an automated approach to transpile TS to JavaScript
>
>allow loading transpiled TS files directly -> solved
>Unfortunately the _original TS file is not supported (which would be the ultimate/ideal solution)
>(related to 2.) we still should have an automated approach to transpile TS to JavaScript

Right, the issue is still opened. I think that fixing the typings might help. If we do not need to import wot-typescript-definitions then we don't have in the output file those lines:
``` js
Object.defineProperty(exports, "__esModule", { value: true }); // maybe this will stay
require("wot-typescript-definitions");
let WoT;
let WoTHelpers;
``` 
Then in ts-node, we could use `typesRoot` option which removes the requirement for /// <reference types="../types/wot" /> .

> Anyway I am OK with merging this improvement.

Thanks, one step at a time and we'll get there 👍

PR #322, opened 2020-09-10
improve http PORT env var situation

joshco:
and when we override with an environment variable, let's actually tell the user!

This is the  PORT side of the issue I raised in  #214   The BaseURI is in #220 
The current implementation silently overrides the port.  This change shows a clear message that it has been overridden.

it also adds an additional environment variable WOT_PORT which will take precedence over PORT if present. This is to allow node-wot implementations that coexist with another web server in the same container, such as nodejs, ruby on rails, nginx/apache for static assets etc.  

Test and documentation included

In #214 I also mentioned running into this, where both servers were trying to bind to the PORT value specified by the heroku build pack.

@danielpeintner I saw your recent env-var PR #319 was merged, so I figured the PORT issue was important to fix.

Signed-off-by: Josh Cohen <joshco@gmail.com>

### DIFF
--- a/packages/binding-http/README.md
+++ b/packages/binding-http/README.md
@@ -160,6 +160,17 @@ The protocol binding can be configured using his constructor or trough servient 
 ```
 When both `serverKey` and `serverCert` are defined the server is started in `https` mode. Examples of `serverKey` and `serverCert` can be found [here](../../examples/security). Moreover, when a security schema is provided the servient must be also configured with valid credentials both client and server side. See [Security](#Security) for further details.
 
+### Environment Variable Overrides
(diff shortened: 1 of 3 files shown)


relu91 (approved):
Looks good! In the future, we might support also setting other config values using ENV variables. For example:
```
WOT.HTTP.PORT=8000
WOT.HTTP.ADDR=http://domain.me/
```

joshco:
I think that's great.  These two PORT vars are meant to be compatible with heroku et al, so they will probably need to stay (unless you can convince Heroku to align with node-wot :) )

Also, you might want to consider underscores, or double underscores rather than periods as separators.  WOT.HTTP isn't a valid shell variable in bash.

danielpeintner (approved):

PR #328, opened 2020-09-18
feat: check whether property is readOnly or writeOnly on exposed-thing

danielpeintner:
* adds tests also

fixes https://github.com/eclipse/thingweb.node-wot/issues/326


danielpeintner:
@relu91 @egekorkan may I ask you to check what I did ...

on writeProperty() call I also added a check whether the property at all exists which makes it a bigger diff but the actual change is minor..

egekorkan (approved):

PR #330, opened 2020-09-22
feat: align all files to use "smart-coffee-machine"

danielpeintner:
Note: used to be partly "coffee-machine" only


danielpeintner:
@fatadel may I ask you to cross check the changes

fatadel:
Seems to be all correct from my side.
The only note is that we will also need to update the tutorial file - https://github.com/fatadel/thingweb.website/blob/master/content/smart-coffee-machine.md.
I can do it myself since anyway I was going to update that tutorial with the new information about the static smart coffee machine page we have added.

fatadel:
@danielpeintner Please take a look at the PR - https://github.com/eclipse/thingweb.node-wot/pull/336.
I've found some naming differences that we haven't spotted in this PR.

fatadel:
I've also updated the tutorial as I said before - https://github.com/eclipse/thingweb.website/pull/23.

PR #335, opened 2020-09-24
feat: uses internal variables for reading/writing count

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/333

### DIFF
--- a/examples/scripts/counter.js
+++ b/examples/scripts/counter.js
@@ -130,83 +130,59 @@ WoT.produce({
     // init property values
     count = 0;
     lastChange = (new Date()).toISOString();
-    // set property handlers
-    thing.setPropertyReadHandler("count", () => {
-        return new Promise((resolve, reject) => {
-            resolve(count);
-        });
-    });
-    thing.setPropertyReadHandler("lastChange", () => {
-        return new Promise((resolve, reject) => {
-            resolve(lastChange);
-        });
-    });
-    thing.setPropertyReadHandler("countAsImage", (options) => {
-        return new Promise((resolve, reject) => {
-            let fill = "black";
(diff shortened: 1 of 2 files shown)


danielpeintner:
@egekorkan @relu91 May I ask one of you to check and approve. Thanks!

examples/scripts/counter.js
```diff
@@ -126,23 +128,32 @@ WoT.produce({
...
+    thing.setPropertyReadHandler("count", () => {
+        return new Promise((resolve, reject) => {
+            resolve(count);
+        });
+    });
+    thing.setPropertyReadHandler("lastChange", () => {
+        return new Promise((resolve, reject) => {
+            resolve(lastChange);
+        });
+    });
```
relu91, line 143: Using async it can be written like this:
```js
thing.setPropertyReadHandler("lastChange", async () => lastChange });
```

relu91 (changes requested):
In general, I think it will look easier to grasp using async-await. I left an example below.  However, it looks correct in the handling of private properties ( no writes allowed).

examples/scripts/counter.js
danielpeintner: My personal preference would be staying with promise return but I do not have a strong opinion. Either way is fine.

To get yet another opinion....  @egekorkan what do you think?

egekorkan: I also prefer the version @relu91 wrote above :)

danielpeintner: Agreed and updated accordingly

danielpeintner:
@egekorkan @relu91 pls take a look once again

relu91 (approved):

PR #336, opened 2020-09-24
feat: align coffee machine namings

fatadel:
This PR resolves naming differences for the smart coffee machine that were not spotted in https://github.com/eclipse/thingweb.node-wot/pull/330.


danielpeintner (approved):
Looks good

Issue #331, opened 2020-09-23
Smart Coffee Machine - Fixes/Updates/Proposal


danielpeintner:
I would like to split the issue in 3 sections about the smart-coffee-machine samples in https://github.com/eclipse/thingweb.node-wot/tree/master/packages/examples/src/scripts

### 1. Bugs

The code uses promises in the wrong way (see [here](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/examples/src/scripts/smart-coffee-machine.ts#L251-L267) or [here](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/examples/src/scripts/smart-coffee-machine.ts#L271-L288), ..). 

There should be *one* promise and errors should return `reject` and not resolve.

```
    thing.setPropertyWriteHandler('availableResourceLevel', (val, options) => {
        return new Promise((resolve, reject) => {
			// Check if uriVariables are provided
			if (options && typeof options === 'object' && 'uriVariables' in options) {
				const uriVariables: any = options['uriVariables'];
				if ('id' in uriVariables) {
					return thing.readProperty('allAvailableResources').then((resources) => {
						const id = uriVariables['id'];
						resources[id] = val;
						return thing.writeProperty('allAvailableResources', resources);
					});
				} else {
					reject('...');
				}
			} else {
				reject('Please specify id variable as uriVariables.');
			}            
        });
    });
```

Note: I did not check the code for validity but I think you get the problem (difference to your code).

### 2. Best Practices

Should we actually propose using uriVariables. I think this is something WoT needs to support given it exists but I suggest using variables in the body.

### 3. Improvements

It would be nice IF the `makeDrink` action is a longer running process. Hence a coffee should not be ready immediately. Moreover... in reality `makeDrink` quickly one after the other should fail given that the coffee machine is not ready yet.

Hence I also propose to remove the `maintenanceNeeded` property and use a more generic `status` that could be an enumeration like ["maintenanceNeeded", "inUse/brewing..", ...]

Doing so would also allow modelling the brewing process that takes some time and would refuse new drinks till the *old* on is ready or the machine is maintained...

@fatadel can you take a look (I think 1. should be fixed. I don't feel strongly about the rest)
@sebastiankb @egekorkan @relu91  others, please comment if you have another opinion

danielpeintner:
relates to https://github.com/eclipse/thingweb.website/issues/19

egekorkan:
Some other things I have seen:
[...]

PR #337, opened 2020-09-24
feat: refine promise usages in smart coffee machine scripts

fatadel:
Part of #331

### DIFF
--- a/packages/core/test/ServerTest.ts
+++ b/packages/core/test/ServerTest.ts
@@ -612,48 +612,6 @@ class WoTServerTest {
         expect(thing).to.have.property("@context").to.deep.include({"@language": "xx"});
     }
 
-    @test async "should reject reading property without uriVariables when they are mandatory"() {
-        let thing = await WoTServerTest.WoT.produce({
-            title: "UriVariablesThing",
-            properties: {
-                uriProp: {
-                    type: 'number',
-                    uriVariables: {
-                        id: {
-                            type: 'string', 
-                            enum: ['water', 'milk', 'chocolate', 'coffeeBeans']
-                        }
-                    }
-                }
-            }
-        });
-
-        thing.setPropertyReadHandler('uriProp', (options) => {
-            return new Promise((resolve, reject) => {
-                // Check if uriVariables are provided
-                if (options && typeof options === 'object' && 'uriVariables' in options) {
-                    const uriVariables: any = options['uriVariables'];
-                    if ('id' in uriVariables) {
-                        return resolve(100);
-                    }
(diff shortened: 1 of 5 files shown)


danielpeintner (changes requested):
I am more in favor of nesting code with if/else so that I can be sure that a `resolve` or `reject` is the last possible command executed.

However if you want to keep your code style I suggest adding a `return` statement after `resolve`/`reject` if yet another `resolve`/`reject` can follow.

see https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject

fatadel:
> I am more in favor of nesting code with if/else so that I can be sure that a `resolve` or `reject` is the last possible command executed.
> 
> However if you want to keep your code style I suggest adding a `return` statement after `resolve`/`reject` if yet another `resolve`/`reject` can follow.
> 
> see https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject

Done!

packages/core/test/ServerTest.ts
```diff
@@ -612,6 +612,48 @@ class WoTServerTest {
         expect(thing).to.have.property("@context").to.deep.include({"@language": "xx"});
     }
 
+    @test async "should reject reading property without uriVariables when they are mandatory"() {
```
danielpeintner, line 615: Thank you @fatadel . You are right this is really an issue.
I will fix it in another PR to better keep track.

May I ask you to remove the test case in this pr. Sorry indeed and thanks!

fatadel: Sure, no worries!

danielpeintner (approved):
Looks good to me.

Whether you want to use `async` and `await` as mentioned by @relu91 [here](https://github.com/eclipse/thingweb.node-wot/issues/331#issuecomment-699088498) is up to you.

Just let me know whether you want to make these changes or you prefer keeping it as is.

fatadel:
> Looks good to me.
> 
> Whether you want to use `async` and `await` as mentioned by @relu91 [here](https://github.com/eclipse/thingweb.node-wot/issues/331#issuecomment-699088498) is up to you.
> 
> Just let me know whether you want to make these changes or you prefer keeping it as is.

I would prefer to keep it as it is. `async/await` might make the code more understandable but for now it is consistent with other promise usages in this example (and also in others I think).

danielpeintner:
Makes sense to me since we now have both possibilities in the examples directory (the counter uses async-await on request, see https://github.com/eclipse/thingweb.node-wot/pull/335)

Issue #239, opened 2020-06-05
Inconsistent code style


ajs124:
I'm taking a look at several parts of this project right now and noticed that there are lots of whitespace errors, e.g. indented empty lines and spaces at the end of lines and inconsistencies in code style, e.g. tabs and spaces randomly mixed, even in the same file.

While this isn't a big problem, it means at least two things (to me)
- it's unprofessional. As harsh at it sounds, if a project cannot get these surface level things right, why would it get the actual code right?
- it's confusing. If I'm contributing code, what should it look like? How should it be formatted, structured, etc.

Assuming you agree to this in principle, the question is how to solve this.
Reformatting old code is annoying, because it leads to huge "useless" commits which make version control history harder to use.
For new code or code that's being touched anyways, it's worth thinking about trying to enforce some kind of standard. There are projects like [editorconfig](https://editorconfig.org/) which aim to address this accross different IDEs and editors while also having CI integration with [GitHub actions](https://github.com/marketplace/actions/editorconfig-action) or other systems.

Therefore my proposal would be to leave the old code as-is and introduce an editorconfig and have CI validate against it on PRs.

relu91:
I do agree that we need a proper code style check for `node-wot`. 

>Reformatting old code is annoying, because it leads to huge "useless" commits which make version control history harder to use.

Well, why not make one commit and refactor the whole repository? I think it does not hurt too much the history and the dev experience. 

About enforcing code style my favorite solution is to use [Husky](https://www.npmjs.com/package/husky). It adds a pre-commit hook in your package.json where you can run your favorite code prettier.  Like so, it is *almost* impossible to commit not formatted code. The downside is that you add a dev dependency. 

footnote: In my PRs I used the `codestyle` script that is provided inside of some binding-* module. There it uses `typescript-standard` as a code formatter. We could continue using that one.

danielpeintner:
> Therefore my proposal would be to leave the old code as-is and introduce an editorconfig and have CI validate against it on PRs.

While I agree with most of your technical arguments I *disagree* with the way it was communicated.
Either way, let's look at the technical arguments:
* we do have editorconfig 
* wo do *NOT* have PR check

> Well, why not make one commit and refactor the whole repository? I think it does not hurt too much the history and the dev experience.

On the hand I agree (since it solves all issues at once) on the other hand I disagree (given that it makes code parts coming from person X while being from Y).

Other than that I see 2 next steps forward now
[...]

PR #339, opened 2020-09-25
Fix typescript defintions in 0.8.x branch

relu91:
This PR fixes the errors provided by the new typescript definition file of the Scripting API. Notice that now classes cannot implement directly `WoT` type since it is a namespace.  This PR works around the issue using `typeof WoT` when returning an object that should implement the functions defined inside the `WoT` namespace.

### DIFF
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -118,9 +118,9 @@ export default class DefaultServient extends Servient {
     /**
      * start
      */
-    public start(): Promise<WoT.WoT> {
+    public start(): Promise<typeof WoT> {
 
-        return new Promise<WoT.WoT>((resolve, reject) => {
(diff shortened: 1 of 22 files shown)


relu91:
I reviewed the examples. My proposal is to put them aside for now and finishing the new API implementation. Fixing them right now means that later we have to re-do the work (since the new API is different).  I suggest the following plan for branch 0.8.x:
- Chose a code style (see #261,#239). I think we could integrate also this step since there'll be major modifications in the codebase... 
- Introduce the new APIs
- Define unit tests for those APIs
- Fixing the examples

Any comments?

relu91:
Argh, we have a failing test in windows for sse. Seems again some misconfiguration or disclosing of connections. I'll check that soon so that we can close this and proceed.

danielpeintner:
Good to merge from my side and go on from that.

Do you want to remove WIP label ?

PR #341, opened 2020-09-25
fix: pass along rejects for handlers

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/340

### DIFF
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -165,16 +165,20 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
                     reject(new Error(`ExposedThing '${this.title}', property '${propertyName}' is writeOnly`));
                 }
 
+                let ps: PropertyState = this.properties[propertyName].getState();
                 // call read handler (if any)
-                if (this.properties[propertyName].getState().readHandler != null) {
+                if (ps.readHandler != null) {
                     console.debug("[core/exposed-thing]",`ExposedThing '${this.title}' calls registered readHandler for Property '${propertyName}'`);
-                    let ps: PropertyState = this.properties[propertyName].getState();
-                    ps.readHandler(options).then((customValue) => {
-                        resolve(customValue);
-                    });
+                    ps.readHandler(options)
+                        .then((customValue) => {
+                            resolve(customValue);
+                        })
+                        .catch((err) => {
+                            reject(err);
+                        });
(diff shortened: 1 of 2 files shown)


danielpeintner:
@egekorkan @relu91 May I ask one of you to check and approve. Thanks!

packages/core/src/exposed-thing.ts
```diff
@@ -165,16 +165,20 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
-                    ps.readHandler(options).then((customValue) => {
-                        resolve(customValue);
-                    });
+                    ps.readHandler(options)
+                        .then((customValue) => {
+                            resolve(customValue);
+                        })
+                        .catch((err) => {
+                            reject(err);
+                        });
```
danielpeintner, line 178: Note: This is the actual fix

danielpeintner: FYI: the changes are very minimal (see https://github.com/eclipse/thingweb.node-wot/pull/341/files?diff=split&w=1)

* introduced variables like ps to avoid writing this.properties[propertyName].getState() over an over again (I should/could have done that in another PR)
* the actual fix is in the 3 lines (see comment above)
* added test-cases (just one fails before this fix "should reject if property read handler fails")

packages/core/test/ServerTest.ts
```diff
@@ -612,6 +612,94 @@ class WoTServerTest {
...
+
+        let readingPossible = false;
+        try {
+            await thing.readProperty("uriProp");
+            readingPossible = true;
+        } catch (e) {
+            // as expected
+        }
+
+        if (readingPossible) {
```
relu91, line 640: What about not using this boolean?
```js
try {
            await thing.readProperty("uriProp");
            fail("reading property 'uriProp' should throw error")
        } catch (e) {
            // as expected
            assert.ok(true);
        }
```

relu91 (approved):

packages/core/test/ServerTest.ts
danielpeintner: This looks a bit strange I agree but throwing fail within try {} does not work.

Anyhow.. its a test only ;-)

relu91: >This looks a bit strange I agree but throwing fail within try {} does not work
that's odd, maybe because fail actually throw an error? 🤔 .

PR #342, opened 2020-09-29
Fix WoT scripts cannot require builtin modules

relu91:
In #315 we updated node-wot to use [vm2](https://www.npmjs.com/package/vm2) module. Since [vm2](https://www.npmjs.com/package/vm2) is more restrictive, it needs to be manually set to support builtin modules require.


danielpeintner (approved):
Looks good.

Verified: test fails without fix and passes with it

PR #347, opened 2020-10-27
fix: fix longpoll client

fatadel:
The longpoll client doesn't work giving an `UnhandledPromiseRejectionWarning: ReferenceError: closed is not defined` exception.
This PR fixes the error.


danielpeintner:
@fatadel can you provide a test-case in your PR that fails before your fix and resolves the issue. Thanks!

PR #352, opened 2020-11-03
fix: fix the way used protocols are checked

### LINKED ISSUE #351: Fix protocol checks
Say, I want to start a Thing (like the smart coffee machine or the counter) but would like to disable some default bindings (say, coap). What I do is create a configuration file and put `{ "coap": null }` there. Otherwise, it will use the coap settings from the default config.
Now, my set will fail in this block https://github.com/eclipse/thingweb.node-wot/blob/d420cfb259dc01d23177fe40ff5c4dea4799cdbc/packages/cli/src/cli-default-servient.ts#L95 because `null` is not `undefined`. So, it will simply terminate with `Cannot read property 'port' of null` error.

fatadel:
Closes #351

### DIFF
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -85,19 +85,19 @@ export default class DefaultServient extends Servient {
         }
         if (!this.config.servient.clientOnly) {
 
-            if (this.config.http !== undefined) {
+            if (this.config.http) {
                 let httpServer = new HttpServer(this.config.http);
                 this.addServer(httpServer);
 
                 // re-use httpServer (same port)
                 this.addServer(new WebSocketServer(httpServer));
             }
-            if (this.config.coap !== undefined) {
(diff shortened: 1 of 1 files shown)


relu91 (approved):

egekorkan:
I'm ok for this but we simply need the validation of this file (there's an issue) and not find every little thing as time goes on. There can be a schema file that simply says that coap is an object with some properties like port and similar. It doesn't have to be required but allow validation in case it is given.

egekorkan:
This config file can also have default values and the validator can insert them (https://ajv.js.org/#assigning-defaults). This way, the mergeConfigs function would not be needed anymore

danielpeintner (approved):
I am ok but I think we should tackle the issue more generally as @egekorkan mentioned. I opened another issue for that, see https://github.com/eclipse/thingweb.node-wot/issues/353

PR #355, opened 2020-11-09
Fix opcua tests

### LINKED ISSUE #349: [test] OPCUA tests fail on Windows
The OPCUA tests fail on Windows (depending on setup?)
_**it used to work on my Windows box but does not any longer**_

Hence we should try to fix the problems by
1. Fix the actual issues (linux and windows treat differently unreachable hosts)
2. Don't run OPCUA tests on Windows (e.g., by using https://github.com/charlesguse/run-script-os or so )

"1." is preferred

relu91:
This PR fixes #349.  Thank you @lukesmolo for the guidance. 

P.s. The issue says that the tests didn't work on windows but they actually failed also in Linux. It should be fixed now on both systems.

### DIFF
--- a/packages/binding-opcua/test/opcua-client-test.ts
+++ b/packages/binding-opcua/test/opcua-client-test.ts
@@ -27,7 +27,7 @@ import { OpcuaServer } from './opcua-server';
 
 
 
-describe.skip('OPCUA client test', function () {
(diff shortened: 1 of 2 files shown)


danielpeintner (approved):
Thanks!

PR #356, opened 2020-11-09
Fix netconf tests

relu91:
This PR fixes `netconf` tests that failed on my windows machine. It was basically just a time issue. The tests are verifying the correct behavior when the `netconf` server is not online. In my windows version, the operating systems give up just after 2000 ms which is the default mocha time out. Increasing the timeout solved the issue.  

As you can see from the commit logs the PR also adds windows as a testing platform in travis. I thought it might worth it since now it seems that all the tests are green.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 os:
   - linux
   - osx
+  - windows
 node_js:
   - "node"
   - "10"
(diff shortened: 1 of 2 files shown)


danielpeintner (approved):
Travis tests pass just fine!

PR #365, opened 2020-12-11
test: GitHub actions PR should fail

danielpeintner:


danielpeintner:
Close, works as expected.

PR #378, opened 2021-01-18
fix: write value before notifying observers

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/376

+ minor OPCUA fixes

### DIFF
--- a/packages/binding-opcua/src/opcua-client.ts
+++ b/packages/binding-opcua/src/opcua-client.ts
@@ -216,7 +216,7 @@ export default class OpcuaClient implements ProtocolClient {
 
 		return new Promise<any>((resolve, reject) => {
 			if (res) {
-				resolve(null);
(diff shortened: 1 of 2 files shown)


danielpeintner:
@egekorkan @relu91 may I ask one of you to make a _sanity_ check ?

Note: OPCUA issue popped up.. as in v8 branch? Not sure why...

relu91 (approved):
Looks good. I bet the problem with opcua is that we fix only the major version in package.json (see [here](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004)). Even though minor or patch releases shouldn't break retro compatibility in the npm world is not rare to found libraries that do not follow those rules. From now on I think it is safer to fix the version, even more in those "less" matained packages like `opcua` or `netconf`.

PR #381, opened 2021-01-22
Fix v0.8.x

danielpeintner:
remove ConsumedThing hierarchy from ExposedThing


danielpeintner:
TODOs
* any -> InteractionInput / Output in core
  *  fix also reading InteractionOutput in all examples (once it is properly returned)
* subscriptions changes (a lot to do!
  * update TS definitions also?

danielpeintner:
@relu91 shall we merge again "early" so that we can try to move on step by step ?

PR #391, opened 2021-03-01
catch decodeURI errors gracefully

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/389


relu91 (approved):
Right. We cannot trust the client sending us well-formed URI parameters.

Issue #255, opened 2020-06-21
[binding-mqtt] MQTT Topics should not start with `/`


egekorkan:
As pointed out by @benjaminbraun , the MQTT topics that node-wot subscribes and publishes to always start with a forwards slash, i.e. `/`. However, this is something that should never be done according to best practices. See more [here](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/).

This also shows something we cannot easily describe with a TD produced by node-wot since the base URI that can be used to identify the broker is not produced and all hrefs are absolute. Thus, a Consumer cannot tell if the topics starts with a `/` or not. For example, `"mqtt://test.mosquitto.org/example_td/actions/setNumber1"` can correspond to:
- Broker: `"mqtt://test.mosquitto.org"` and topic `"example_td/actions/setNumber1"` OR
- Broker: `"mqtt://test.mosquitto.org"` and topic `"/example_td/actions/setNumber1"`

Related issue in W3C WoT TD repo: https://github.com/w3c/wot-thing-description/issues/878

JKRhb:
I just ran into the same issue using `node-red-contrib-web-of-things` – there is always a `/` added to topics as a prefix making it, unfortunately, impossible to subscribe or publish to topics that don't start with one. 

Have there been any discussions on this in the meantime? Maybe it could be a workaround to simply remove the leading `/` for now in all cases? (And maybe add a specific vocabulary term to the binding templates if a leading `/ ` should be used?)

egekorkan:
I think that the `/` should be simply removed and for the consumption of MQTT TDs, a `base` with an MQTT should be present since that could be the only way of doing it while having `href` in forms. 

Another possibility would be to say that the `href` only defines the location of the broker and an additional `mqv:topic` clearly states the topic and can have `/` or not. Also, important to note that MQTT does not have a URI scheme that is standardized.

JKRhb:
Sounds good! I guess the simple `/` removal would be an easy fix for `node-wot` while the addition of `mqv:topic` (with the assumption that `href` refers to a broker) could be a good medium-term solution?

It is probably not ideal, however, that the broker address has to be stated then in every `form` if the broker is defined using `href` and not `base` but I suppose the current TD structure does not allow for a different solution.

sebastiankb:
I think, a clear URL guideline with MQTT broker and topic information should be present at the (future) MQTT binding document. Maybe we should also think about a RFC.

[...]

PR #398, opened 2021-03-08
fix: Remove leading "/" from MQTT topics

JKRhb:
This PR is a very simple fix/workaround for #255: The leading `/` that is currently derived from the URI path is being removed. As most MQTT topics should not feature a leading `/` (among other things, because it is not best practice to so, see #255) this fix should not cause much harm while making it possible to actually use `node-wot` with the majority of MQTT devices.

### DIFF
--- a/packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
+++ b/packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
@@ -17,7 +17,6 @@
  * Protocol test suite to test protocol implementations
  */
 
-
 import { suite, test, slow, timeout, skip, only } from "mocha-typescript";
 import { expect, should, assert } from "chai";
 // should must be called to augment all variables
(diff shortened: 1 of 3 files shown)


relu91:
Just a reminder you have to sign your commits as described in [Contributing.md](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md)

JKRhb:
> 
> 
> Just a reminder you have to sign your commits as described in [Contributing.md](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md)

Thank you! I force-pushed a new signed commit (unfortunately, my account request at eclipse.org is still pending at the moment :/).

egekorkan:
Also another point here: The MQTT tests are simply commented out and this change only on the Consumer will actually break the Exposed Thing that is implemented for this. Did you do some tests?

JKRhb:
> 
> 
> Also another point here: The MQTT tests are simply commented out and this change only on the Consumer will actually break the Exposed Thing that is implemented for this. Did you do some tests?

Thank you for pointing this out! I will do some (more) testing and have a look on the Exposed Thing.

egekorkan:
By the way, I of course support removing it from Exposed Thing as well. If we don't, it would be actually a breaking change and need to be postponed to 0.8.x :(

JKRhb:
I did some testing now and removed the leading `/` from topics in the `mqtt-client` (I somehow missed two spots in my first commit).

In a second `fixup` commit I rearranged the slash so that the exposed `href`s should stay the same while the topics are aligned with the client. This change, however, is still untested by me.

danielpeintner:
Note: The issues about failing test are w.r.t. OPCUA and not related to this PR.
I hope re-running the workflow helps...

danielpeintner:
Grr, it does NOT help.

Can you merge master into your PR and push...

JKRhb:
> 
> 
> Grr, it does NOT help.
> 
> Can you merge master into your PR and push...

Thanks for the fixes! The CI seems to pass now :)

Do you think this PR is ready to be merged? Or should some test cases (or something else) be added before?

danielpeintner:
> Do you think this PR is ready to be merged? Or should some test cases (or something else) be added before?

@egekorkan ?

relu91 (approved):
Looks good to me. Not sure how this change could be possibly tested, but if you have something in mind I'd rather have one simple legitimate test than no one.

JKRhb:
Sorry for all those fixup commits and the stress for the CI. I now added some code and an action to perform the MQTT tests with a broker running under `localhost`. When not running in the CI pipeline `mqtt://test.mosquitto.org:1883` should be used as before.

 I'm not entirely sure if it works yet but if it does maybe this might be a good solution. If not I would suggest skipping this test in the CI pipeline as before.

packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
```diff
@@ -28,66 +27,71 @@ import { Servient, ExposedThing } from "@node-wot/core";
...
         try {
-
             let servient = new Servient();
+            var brokerUri;
+
+            if (process.env.matrix_os == "ubuntu-latest") {
+                brokerUri = "mqtt://localhost:1883";
+            } else {
+                brokerUri = "mqtt://test.mosquitto.org:1883"
+            }
```
relu91, line 41: As I said in my other comment here it would be easier if you just do:
```js
brokerUri = process.env.broker || "mqtt://test.mosquitto.org:1883"
```

relu91 (changes requested):

.github/workflows/ci.yaml
```diff
@@ -26,6 +26,18 @@ jobs:
...
+          version: '1.6'
+          ports: '1883:1883 8883:8883'
+          certificates: ${{ github.workspace }}/.ci/tls-certificates
+          config: ${{ github.workspace }}/.ci/mosquitto.conf
+          container-name: 'mqtt'
       
       - name: Test
         run: npm run test
+        env:
+          matrix_os: ${{ matrix.os }}
```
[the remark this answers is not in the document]
JKRhb: Thank you, passing the broker address is a much more elegant solution! I will integrate that in moment.

One thing I was struggling with was the fact that docker images can only be run on Linux. I would therefore probably still pass the OS type as an environment variable and skip the test (or let it automatically pass) if it is performed on Windows or Mac OS. Do you think this is the right direction or is there maybe a better way?

packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
```diff
@@ -28,66 +27,71 @@ import { Servient, ExposedThing } from "@node-wot/core";
...
-    @test.skip "should expose via broker"(done: Function) {
-
+    @test "should expose via broker"(done: Function) {
         try {
-
             let servient = new Servient();
 
-            let brokerServer = new MqttBrokerServer("mqtt://test.mosquitto.org:1883");
+            if (osToSkip.includes(process.env.matrix_os)) {
+                done(); // TODO: should be replaced with a skip()
```
JKRhb, line 39: I did not find a way yet in `mocha-typescript` to skip a test on a certain condition (while this is possible in the "regular" `mocha` library).

JKRhb:
After a lot of trial and error I think I got the tests to work now without adding anything new to the CI pipeline :) It seems to have been an issue with the delay between publishing to topics and the time limit defined for the test. Increasing both seems to have solved the problem. 

I also removed another `/` that remained in front of the topic referring to the thing itself (or rather its title) that is published to when the thing is exposed. All in all I think the PR could be merged now – although I was also wondering if the changes to the exposed thing actually make this a breaking change.

(Also I noticed that in my commit messages the initial letters after the colons are uppercase which is against the contribution guidelines – should I edit this?)

JKRhb:
With the updated test this PR should now also resolve issue #161.

danielpeintner:
Thanks @JKRhb!
LGTM.

relu91:
I fought well. Great job!

JKRhb:
Great! Thank you for your help and your patience :)

PR #401, opened 2021-03-17
V0.8.x next2

danielpeintner:


danielpeintner:
@relu91 I fixed some obvious issues like filename typo and method syntax.

Having said that, I have still issues building since `InteractionOutput` is not assignable `WoT.InteractionOutput` 

```
src/interaction-output.ts:36:5 - error TS2416: Property 'data' in type 'InteractionOutput' is not assignable to the same property in base type 'InteractionOutput'.
  Type 'import("C:/Projects/WoT/thingweb/thingweb.node-wot-danielpeintner/packages/core/node_modules/web-streams-polyfill/dist/types/polyfill").ReadableStream<any>' is not assignable to type 'ReadableStream<any>'.
    Types of property 'getReader' are incompatible.
      Type '{ ({ mode }: { mode: "byob"; }): ReadableStreamBYOBReader; (): ReadableStreamDefaultReader<any>; }' is not assignable to type '() => ReadableStreamDefaultReader<any>'.

36     data?: ReadableStream; // <any>;
```

Did this work on your side? I think that the polyfill is not correctly matched?
I am still unsure about this polyfill approach...

PR #403, opened 2021-03-25
[V0.8.x] Fixing tests and package-lock

relu91:
This PR is based on the current work in #401. It fixes some problems in the tests (first commit) and it tries to fix all the package-lock files that went out sync after merging with the master. Also now bootstrap uses the `ci` option which should help us to keep locked the versions of the dependencies. 

By the way, we should encourage committers to install new packages using `lerna add <package> --scope @node-wot/<subpackage>` it should cause fewer troubles.


danielpeintner:
Thanks @relu91

PR #406, opened 2021-03-30
fix: use correct types

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/404


danielpeintner:
@relu91 @egekorkan do you know why the types were like this?

relu91 (approved):

relu91:
I have no clue why there were designed like that... anyhow the change seems reasonable.

PR #415, opened 2021-05-14
fix: examples/scripts/counter-client.js formIndex

iomz:
When I run the example for the counter thing and its consumer `examples/scripts/counter-client.js` I get the error `ConsumedThing 'counter' missing formIndex '2'`:
```
vagrant@ubuntu-focal:/vagrant/thingweb.node-wot$ node packages/cli/dist/cli.js examples/scripts/counter-client.js -
-clientonly
[cli] WoT-Servient using defaults as 'wot-servient.conf.json' does not exist
{
  servient: { clientOnly: true, scriptAction: false },
  http: { port: 8080, selfSigned: false },
  coap: { port: 5683 },
  log: { level: 'info' }
}
[cli/default-servient] DefaultServient started
[core/servient] Servient generating ID for 'servient': 'urn:uuid:8b2cc2c6-2354-49b2-8b1c-fee5f9fd1351'
[core/servient] Servient has no servers to expose Things
[cli] WoT-Servient loading 1 command line script
[cli] WoT-Servient reading script examples/scripts/counter-client.js
[cli] WoT-Servient running script '/********************************************************************************'... (43 lines)
=== TD ===
{
...
}
==========
count value is 0
count value after increment #1 is 1
count value after increment #2 (with step 3) is 4
Script error: Error: ConsumedThing 'counter' missing formIndex '2'
    at Object.ConsumedThing.getClientFor (/vagrant/thingweb.node-wot/packages/core/dist/consumed-thing.js:139:23)
    at /vagrant/thingweb.node-wot/packages/core/dist/consumed-thing.js:288:32
    at new Promise (<anonymous>)
    at Object.ConsumedThing.invokeAction (/vagrant/thingweb.node-wot/packages/core/dist/consumed-thing.js:282:16)
    at Object.base.apply (/vagrant/thingweb.node-wot/packages/core/node_modules/vm2/lib/contextify.js:620:32)
    at /vagrant/thingweb.node-wot/examples/scripts/counter-client.js:35:21
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

It seems the client is invoking the "decrement" action with `formIndex` _InteractionOption_ as part of #228 but I suppose the corresponding form should be the one with CoAP (which is at the index 1), so the correct value should be 1 instead.

cf. in my environment the `forms` was:
```
[
  {
    href: 'http://10.0.2.15:8080/counter/actions/decrement{?step}',
    contentType: 'application/json',
    op: [ 'invokeaction' ],
    'htv:methodName': 'POST'
  },
  {
    href: 'coap://10.0.2.15:5683/counter/actions/decrement',
    contentType: 'application/json',
    op: 'invokeaction'
  }
]
```

### DIFF
--- a/examples/scripts/counter-client.js
+++ b/examples/scripts/counter-client.js
@@ -43,13 +43,12 @@ WoTHelpers.fetch("coap://localhost:5683/counter").then(async (td) => {
 }).catch((err) => { console.error("Fetch error:", err); });
 
 function getFormIndexForDecrementWithCoAP(thing) {
-    // return formIndex: 0 if no CoAP target IRI found
-    let fi = 0;
     thing.getThingDescription()['actions']['decrement']['forms']
         .forEach((form, index) => {
         if (/^coaps?:\/\/.*/.test(form.href)) {
-            fi = index;
+            return index;
         }
     });
-    return fi;
+    // return formIndex: 0 if no CoAP target IRI found
+    return 0;
 }
--- a/examples/scripts/counter-client.js
+++ b/examples/scripts/counter-client.js
@@ -43,12 +43,12 @@ WoTHelpers.fetch("coap://localhost:5683/counter").then(async (td) => {
 }).catch((err) => { console.error("Fetch error:", err); });
 
 function getFormIndexForDecrementWithCoAP(thing) {
-    thing.getThingDescription()['actions']['decrement']['forms']
-        .forEach((form, index) => {
-        if (/^coaps?:\/\/.*/.test(form.href)) {
-            return index;
+    let forms = thing.getThingDescription()['actions']['decrement']['forms'];
+    for (let i = 0; i < forms.length; i++) {
+        if (/^coaps?:\/\/.*/.test(forms[i].href)) {
+            return i;
         }
-    });
+    }
     // return formIndex: 0 if no CoAP target IRI found
     return 0;
 }
--- a/examples/scripts/counter-client.js
+++ b/examples/scripts/counter-client.js
@@ -12,7 +12,6 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
-
 WoTHelpers.fetch("coap://localhost:5683/counter").then(async (td) => {
     // using await for serial execution (note 'async' in then() of fetch())
     try {
@@ -31,8 +30,16 @@ WoTHelpers.fetch("coap://localhost:5683/counter").then(async (td) => {
         await thing.invokeAction("increment", undefined, { uriVariables: { 'step': 3 } });
         let inc2 = await thing.readProperty("count");
         console.info("count value after increment #2 (with step 3) is", inc2);
-        // decrement property with formIndex == 1 (via CoAP binding)
-        await thing.invokeAction("decrement", undefined, { formIndex: 1 });
+        // look for the first form for decrement with CoAP binding
+        // formIndex: 0 if not found
+        let fi = 0;
+        thing.getThingDescription()['actions']['decrement']['forms']
+            .forEach((form, index) => {
+            if (/^coaps?:\/\/.*/.test(form.href)) {
+                fi = index;
+            }
+        });
+        await thing.invokeAction("decrement", undefined, { formIndex: fi });
         let dec1 = await thing.readProperty("count");
         console.info("count value after decrement is", dec1);
     }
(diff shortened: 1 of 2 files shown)


danielpeintner:
The `formIndex: 2` is there because in my local setup I do have two http forms first (due to 2 IP addresses). Hence CoAP is on index 2 🙈

Having said that, I think the current example is meant to _show_ what one can do but it fails to work across different users. One would need to look into the TD forms to actually know which `formIndex` to use.

Hence, I suggest we remove the `formIndex` and just mention it in a comment... this way it becomes more stable.

Does this makes sense? What do you think?
@iomz @relu91 @egekorkan

egekorkan:
So if we want to put it in the code, I would suggest something that looks for a form with coap first. A regex can simply do that and it would "simulate" a human looking for one .

iomz:
@danielpeintner Aha, thanks. As I'm new to Typescript and was not sure how it picks up the IP address of the host, but now I see it's done inside [this loop](https://github.com/eclipse/thingweb.node-wot/blob/ed18712144b5ac1d78a2cb717946fe261d8020b7/packages/binding-http/src/http-server.ts#L218). I was doing the Thingweb hands-on and just wanted to make the first example work out of the box 😄

> I would suggest something that looks for a form with coap first. A regex can simply do that and it would "simulate" a human looking for one

@egekorkan That'll be a great example – in fact I learned _InteractionOptions_ through this.

danielpeintner:
> So if we want to put it in the code, I would suggest something that looks for a form with coap first. A regex can simply do that and it would "simulate" a human looking for one .

Mhh, the question for me is whether we should do so ... or do we confuse people by doing so.

iomz:
I attempted to look up the form for the formIndex – I'm not entirely sure if this's the right way to do this, though.
Perhaps it's a bit confusing for some, but I also want to know a correct way to introspect the properties of the TD.

egekorkan:
@iomz to me that looks correct but I haven't checked out your PR to test it

danielpeintner:
I will check the PR later...

Before that I would like to make sure that the majority thinks having this extended code makes sense in this sample. I am bit biased but if others think this is good to have I am fine with it! Thanks @iomz !

@egekorkan @relu91 @sebastiankb others any opinion?

relu91 (changes requested):
I'd approve the changes but maybe I would just introduce a small function that extracts the right form index. It will hide the complexity for newcomers and make the code more readable. 

As a side note: in the future it would be better to have a dedicated example to formIndex readings and have another one with just the default operations ( no formIndex ).

iomz:
>I'd approve the changes but maybe I would just introduce a small function that extracts the right form index. It will hide the complexity for newcomers and make the code more readable.

@relu91 Do you mean something like this micro-purpose function?
```
function getFormIndexForDecrementWithCoAP(thing: WoT.ConsumedThing): number {
	// return formIndex: 0 if no such pattern found
	let fi = 0;
	thing.getThingDescription()['actions']['decrement']['forms']
		.forEach((form: TD.Form, index: number) => {
			if (/^coaps?:\/\/.*/.test(form.href)) {
				fi = index;
			}
		});
	return fi;
}
```
There should be a number of ways (e.g., passing the action as well/implementing this as an ad-hoc class method for ConsumedThing), but as an example I suppose it ends up looking rather complex than having the "simulating lookup" logic just before making the request.

relu91:
yes exactly that would work. It is just an educational example and it can contain just the simple logic that you already using. So after adding that I'd merge this PR. 🎉

iomz:
Thanks @relu91 I've replaced the part with a function.

danielpeintner:
Looking at the code seemed fine by me. However, I tested the code locally and I noticed that the function `getFormIndexForDecrementWithCoAP()` *always* returns `0`.

I was puzzled myself but apparently a `return` statement in `foreach` does not abort the loop as you and I would expect ...
https://stackoverflow.com/questions/34653612/what-does-return-keyword-mean-inside-foreach-function

@iomz  may I ask you to update the code accordingly. Otherwise it looks good to me! Thanks!

iomz:
@danielpeintner Thanks for catching the bug🙈
I've pushed the fixed version.

danielpeintner (approved):
LGTM

danielpeintner:
@relu91 any more comments?

relu91 (approved):
Well done!

iomz:
Thank you all!

Issue #236, opened 2020-06-03
[question] Naming convention for devices built with node-wot


egekorkan:
This is mostly to gather opinions:

We have a couple of devices built with node-wot that are for now on GitHub but we will publish them on npm soon. In a similar way to Node-RED packages always named `nodered-contrib-mypackage` (example https://www.npmjs.com/package/node-red-contrib-io-key), should we also set up a naming convention for node-wot? Here is one simple infrared sensor example: https://github.com/tum-esi/wot-sys/tree/master/Devices/nodewot-infrared-sensor-dobot

I have some ideas below but would like to hear the opinions of others:
1. `@node-wot/contrib/mypackage`: Not sure since `/contrib` after `@node-wot` makes me think that it is part of the node-wot main package and maintained by node-wot developers, which should not be the case
2. `@node-wot-contrib/mypackage`
3. `@node-wot-contrib-mypackage`
4. `node-wot-contrib-mypackage`: I find this one to be the closest to the way Node-RED does, since an `@node-wot` associates it with node-wot developers.

Looking forward to ideas/opinions :)

danielpeintner:
I think the idea is to start with a *best practice* on how _others_ can publish their contribution.

At the moment *our* packages look as follows:
* `@node-wot/core`
* `@node-wot/td-tools`
* `@node-wot/binding-http`
* ...

This makes me think the following template makes most sense but I do not have a strong opinion either ;-)
* `@node-wot/contrib-<contribution-name>`
* à la 
* `@node-wot/contrib-foo`
* `@node-wot/contrib-bla`
* ...

relu91:
`@node-wot/contrib-<contribution-name>`

I think that using this pattern people would need the write access for `@node-wot` organization. Rollup went for this direction too and create a dedicated repository for the most famous [supported plugins](https://github.com/rollup/plugins). However, I think that it creates confusion and it is quite difficult to manage issues/PR.  

I would vote for `node-wot-contrib-mypackage`, but I do not like too much this one either.

erkann-sen:
`node-wot-contrib-mypackage` this convention gives an expression like  it is a node library extension not a node-wot extension, so I think it would be better to some how divide node-wot from the ending and what I can offer is putting an underscore like `node-wot_contrib-mypackage` or `@node-wot_contrib-mypackage`

egekorkan:
[...]

PR #417, opened 2021-05-30
binding-wotfirestore

hidetak:
I have developed a binding that uses Firestore.
How to use it is described in [README](https://github.com/hidetak/thingweb.node-wot/blob/master/packages/binding-wotfirestore/README.md).

I apologize for any inconvenience this may cause, as I am unfamiliar with English and the manner of creating PR for this repository.

I hope you will be interested and give me some suggestions for the merge.


packages/binding-wotfirestore/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@hidetak/binding-wotfirestore",
+  "version": "0.0.3",
+  "description": "Firestore binding for node-wot",
+  "repository": "https://github.com/hidetak/thingweb.node-wot",
```
egekorkan, line 5: This will need to be updated after merge

packages/binding-wotfirestore/src/wotfirestore-client.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  public stop(): boolean {
+    return true
+  }
+
+  public setSecurity(
+    metadata: Array<TD.SecurityScheme>,
+    credentials?: any
+  ): boolean {
+    // Firestoreにより通信路のセキュリティは確保
```
egekorkan, line 210: what is this part?

packages/binding-wotfirestore/src/wotfirestore-handler.ts
```diff
@@ -0,0 +1,296 @@
...
+          err,
+          ' topic: ',
+          topic
+        )
+        reject(err)
+      })
+  })
+}
+
+export const subscribeFromFirestore = async (
```
egekorkan, line 138: Should it be called `subscribeToFirestore` ?

egekorkan (changes requested):
I have done a quick review and did not try it locally since I don't have a firestore account. Some general remarks:
- There are some Japanese comments that are worth translating into English
- The server side binding uses `ConsumedThing` methods but this will become obsolete in the near future
- I think it is worth mentioning that the "Exposed Thing" data is actually stored in firebase, which is unique compared to other bindings we have.
- It is good to document how WoT operations map to firebase API. You are already mentioning the URL but the rest are in the code. It would be nice if you can put a section in the Readme explaining how the mapping happens.

examples/scripts/wot-servient.conf.json
```diff
@@ -28,6 +28,18 @@
 			"priority": 10
 		}
 	},
+    "wotfirestore": {
```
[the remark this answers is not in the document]
hidetak: So this is where you describe the configuration for Tutorial.
I thought it was a place to put the JSON format of the configuration.
I'll delete it.

packages/binding-wotfirestore-browser-bundle/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Supported bindings 
+HTTP / HTTPS / WebSockets
+
+## Embedding Firestore Binding library in HTML 
+
+Include the following script tag in your html
+
+```js
+<script src="https://cdn.jsdelivr.net/npm/@hidetak/binding-wotfirestore-browser-bundle@latest/dist/binding-wotfirestore-bundle.min.js"></script>
```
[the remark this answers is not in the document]
hidetak: Okay.

packages/binding-wotfirestore-browser-bundle/README.md
```diff
@@ -0,0 +1,28 @@
...
+var WoTFirestoreClientFactory = BindingWoTFirestore.WoTFirestoreClientFactory;
+var WoTFirestoreCodec = BindingWoTFirestore.WoTFirestoreCodec;
+var WoTFirestoreServer = BindingWoTFirestore.WoTFirestoreServer;
+```
+
+## Using binding-wotfirestore browser bundle library in web frameworks
+
+Install browser-bundle in your project by running
+
+* `npm install @hidetak/binding-wotfirestore-browser-bundle`
```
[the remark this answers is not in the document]
hidetak: Okay.

packages/binding-wotfirestore-browser-bundle/index.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const BindingWoTFirestore = require("@hidetak/binding-wotfirestore");
```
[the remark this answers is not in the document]
hidetak: Okay.

packages/binding-wotfirestore-browser-bundle/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "license": "EPL-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist/"
+  ],
+  "main": "dist/binding-wotfirestore-bundle.min.js",
+  "devDependencies": {
+    "@hidetak/binding-wotfirestore": "0.0.1",
```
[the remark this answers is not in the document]
hidetak: Okay.

packages/binding-wotfirestore/src/wotfirestore-client.ts
hidetak: Now it always returns `true` because it does not use security. Is it correct to return `false` if it is not used?

packages/binding-wotfirestore/src/wotfirestore-handler.ts
hidetak: Thank you. I will modify this method name.

hidetak:
Thank you for reviewing my PR !

> There are some Japanese comments that are worth translating into English

I will translate Japanese comment to English.

> The server side binding uses ConsumedThing methods but this will become obsolete in the near future

I'm sorry, I'm not sure which part of your point you were referring to.

> I think it is worth mentioning that the "Exposed Thing" data is actually stored in firebase, which is unique compared to other bindings we have.

Okay, I will try to explain in README.

> It is good to document how WoT operations map to firebase API. You are already mentioning the URL but the rest are in the code. It would be nice if you can put a section in the Readme explaining how the mapping happens.

Okay, I will try to explain above too.

I'll push it when I update it again, but it may take a while.

egekorkan:
@hidetak you are welcome! Thank you for providing a binding in the first place :) I also think that your binding would be very interesting for discovery since firestore can easily store TDs. 

> The server side binding uses ConsumedThing methods but this will become obsolete in the near future

What I had meant by that is that in the server side you use methods such as `readproperty`, `observeproperty`. These will become obsolete in the near future so you have to implement variables.

hidetak:
@egekorkan 
Thank you for letting me know.
Sorry for the inconvenience, but I would appreciate it if you could let me know if there is any code that is implemented without using the obsolete method.

examples/scripts/wot-servient.conf.json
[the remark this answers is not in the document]
egekorkan: So it should be deleted from here (in my opinion) but it can and should be put at the package's README and also https://github.com/eclipse/thingweb.node-wot/blob/5ffd659a2502657151ad4927dacab3849a09a060/packages/cli/src/cli.ts#L220

packages/binding-wotfirestore/src/wotfirestore-client.ts
egekorkan: Ah I see. Then I am not sure. @danielpeintner what do you think?

egekorkan:
Sure, just for reference what you have is this: https://github.com/eclipse/thingweb.node-wot/blob/05f114608cd3cafb4fd84d026bc11f0eff6b27e3/packages/binding-wotfirestore/src/wotfirestore-server.ts#L136

However, I could not find an example of this, also in the 0.8.x branch. @danielpeintner am I wrong somehow?

danielpeintner:
> Sure, just for reference what you have is this:
> 
> https://github.com/eclipse/thingweb.node-wot/blob/05f114608cd3cafb4fd84d026bc11f0eff6b27e3/packages/binding-wotfirestore/src/wotfirestore-server.ts#L136

Yes, `observeProperty()` goes away for ExposedThing in the future.
All writes/reads etc are to be handled internally in "your" code.

Since it is you creating the exposed thing you can react once a write is executed... Am I wrong?

Please take a look at Slide 10 of https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-03-online-f2f/2021-03-25-WoT-F2F-Scripting%20API%20status.pdf

danielpeintner (commented):
First of all thanks for your contribution. 

Note: I do not have firestore account either and might not have enough time to test it properly.
Question: Do you plan to provide tests? Is this somehow possible with a generic account?


I was wondering whether it makes sense to change all `@hidetak/binding-wotfirestore` et cetera that need to be changed later should be changed before merging.

We are also thinking about allowing externals to create their packages so that we do not have to keep up with possible changes, see https://github.com/eclipse/thingweb.node-wot/issues/236

Do you have any opinion on that. Would you prefer to host the bindings yourself or rather as part of node-wot.

hidetak:
@danielpeintner 
Thank you for your comment !

> Note: I do not have firestore account either and might not have enough time to test it properly.
> Question: Do you plan to provide tests? Is this somehow possible with a generic account?

I think the test is to create a Firestore mock, but to be honest, it is difficult to prepare it right away.
First, I'm going to add a test account to my Firesbase account and prepare a sample program using that account. I'll create it in a private repository and share it with you.

> I was wondering whether it makes sense to change all @hidetak/binding-wotfirestore et cetera that need to be changed later should be changed before merging.
>
> We are also thinking about allowing externals to create their packages so that we do not have to keep up with possible changes, see #236
>
> Do you have any opinion on that. Would you prefer to host the bindings yourself or rather as part of node-wot.

I think WoT is a good standard for the development of IoT and I would like to see it become more popular.
I think it is important to have a naming convention so that many developers can participate in the development of Binding and Script and share their artifacts.
Node-RED has a [flows](https://flows.nodered.org/) page where you can find Flows, Nodes, etc. I think it would be even better if we could provide such a mechanism as well.

Nevertheless, there is no problem in changing the package name this time.

hidetak:
@danielpeintner 

Thank you for teaching.

> Yes, observeProperty() goes away for ExposedThing in the future.
> All writes/reads etc are to be handled internally in "your" code.
>
> Since it is you creating the exposed thing you can react once a write is executed... Am I wrong?

I think it will be difficult to implement Binding if the ConsumedThing method is removed.
Because it is the Script, not the Binding, that changes properties and fires events.
I think there should be a standard way for the Binding to know what the Script is doing.

danielpeintner:
Maybe I misunderstood the problem _OR_ I was not very clear with my statements...

In your `observeProperty()` case you are listening to changes and report them back to firestore (e.g., `writeDataToFirestore()`). Correct?

I think what you can do is the is the following.
You can add handlers to the ExposedThing (e.g., `setPropertyWriteHandler("name", handler)`)

Doesn't that work for your use-case?

hidetak:
@danielpeintner 

> In your observeProperty() case you are listening to changes and report them back to firestore (e.g., writeDataToFirestore()). Correct?

Correct.

> You can add handlers to the ExposedThing (e.g., setPropertyWriteHandler("name", handler))

I will try to use `setPropertyWriteHandler` instead of `observeProperty`.
I understood that this method(like `setPropertyWriteHandler`) is the standard way.

Thank you letting me know !

egekorkan:
Also regarding the firestore account for testing, we can add secrets to this repository and GitHub actions would use these credentials. However, they would be used by anyone doing a PR and can consume your allowed credits and charge you money. Maybe you can only allow a small number of requests for the test account.

hidetak:
@egekorkan 
Thank you, I will care of  cost for firestore when prepare test.

hidetak:
@egekorkan @danielpeintner 

I have prepared a sample for binding-wotfirestore.  
I invited you to the repository which is private.
It includes my Firebase account so you can try it right away.  
Please check README.md for details.
I hope you will give it a try when you have time.

maxim-s-barabash:
The firebase has an emulator.  It is very easy to set up.https://firebase.google.com/docs/rules/emulator-setup

danielpeintner:
> I have prepared a sample for binding-wotfirestore.
> I invited you to the repository which is private.

Thanks. I cannot promise to take a look soon... will be offline shortly till to the end of the month...

hidetak:
@maxim-s-barabash 
Thanks for letting me know there is an emulator for firestore.
I've only tried it lightly, but this Binding worked on the emulator.

hidetak:
@danielpeintner 

> Thanks. I cannot promise to take a look soon... will be offline shortly till to the end of the month...

OKay.
I will try to address the points raised against README or  prepare for testing using the emulator.

danielpeintner:
_I am back from vacation..._

@hidetak I am wondering whether you have had success in using the emulator and setting up tests?

hidetak:
@danielpeintner 
I envy you that you took a vacation (^^)

Thank you for caring about this PR.
I added test using the firestore emulator, which you can run with `npm run test`.

hidetak:
@egekorkan 

Sorry for the delay in responding to your suggestions.
I have added descriptions to the README for the following points that you pointed out.

> I think it is worth mentioning that the "Exposed Thing" data is actually stored in firebase, which is unique compared to other bindings we have.

I've added a description to Overview.

> It is good to document how WoT operations map to firebase API. You are already mentioning the URL but the rest are in the code. It would be nice if you can put a section in the Readme explaining how the mapping happens.

I've added a section on `How to use Firestore` and added some explanations.

packages/binding-wotfirestore/src/codecs/wotfirestore-codec.ts
```diff
@@ -0,0 +1,57 @@
+import * as TD from '@node-wot/td-tools'
```
relu91, line 1: Why do need a decoder for the firestore? the logic seems nothing special just normal text encoding and decoding.

danielpeintner:
@hidetak did you check the tests locally?

The build/test checks fail with
> The job running on runner GitHub Actions 20 has exceeded the maximum execution time of 360 minutes.

Do the tests require setting up the emulator first? (might be more difficult for GitHub actions)

packages/binding-wotfirestore/src/codecs/wotfirestore-codec.ts
hidetak: @relu91 
Thank you for giving comment.
I am now storing data in Firestore in Buffer format, so encoder/decoder is needed.
I think you are suggesting that since Firestore can store JSON normally, encoder/decoder is unnecessary, and that may be true.
When I was making it, I thought it would be better to design it this way, thinking that it would be able to handle a variety of data types via Buffer.
Was this an unnecessary consideration...

hidetak:
@danielpeintner 
I apologize for the inconvenience.
When I run `npm run test`, it automatically starts the Emulator. However, the Emulator is not stopped after the test is run, so this may be the cause.
I will try to find out how to stop Emulator automatically after running the test.

packages/binding-wotfirestore/src/codecs/wotfirestore-codec.ts
egekorkan: If this allows for more data types, then why not. It also shows that we can actually handle such data types in the WoT (and with node-wot). I know that I asked for extra documentation-like stuff a lot but if you can provide code+TD example for such data types, it would be very interesting for node-wot but also for the standardization :)

packages/binding-wotfirestore/README.md
```diff
@@ -0,0 +1,221 @@
...
+- things/\<hostName\>%2F\<title of Thing Description\>%events%2F\<name of event\>  
+  Stores the emitted events.
+
+Data will always be overwritten and no history will be retained.
+
+## Support functions for Client and Server implementations
+
+The functions that can be implemented for Client and Server differ as follows.
+
+| functions          | Client | Server |
```
egekorkan, line 206: Nitpicky comments:
- function should be called WoT operation
- read property should be readproperty, same for all others
- Having - for some operations gives the impression that they are not implemented however a client cannot emit an event so I think it is better to say n/a for that.

Other comments:
- Are meta operations like `readallproperties` implemented ? I am guessing not but it would be good to document it here.

packages/binding-wotfirestore/src/codecs/wotfirestore-codec.ts
hidetak: To be honest, currently I haven't come up with an example of such a data type.

packages/binding-wotfirestore/README.md
hidetak: @egekorkan 
Thank you for giving comment !

To begin with, I may not yet understand the changes in the API that ExposedThing has.
Does this mean that the only methods available when implementing a Thing will be the following methods?
https://www.w3.org/TR/wot-scripting-api/#the-exposedthing-interface

If so, I don't think this table I've written is appropriate.

I think that there is a method that has not been implemented among the methods defined here, but when is it going to be implemented ?

hidetak: @egekorkan 
I deleted the table once because it might have been misleading.

hidetak:
@danielpeintner 
I modified package.json so that the Firestore Emulator exits after the test is finished.
When I, on my PC, run `npm run test` in the root directory, I get an error with binding-mqtt, but I was able to confirm that the error does not occur when I remove it and run it.
I would appreciate it if you could tell me about the test, if you are still having problems with it.

packages/binding-wotfirestore/README.md
egekorkan: Sorry about checking everything this late. I think that this table is needed to give a quick overview of what different types of operations are supported by this binding. A Thing, regardless of its protocol, might need to offer events and it should be clear that I can use this binding to achieve that behavior I wanted. You can find the list of all the current operations at https://w3c.github.io/wot-architecture/#sec-hypermedia-forms

egekorkan:
@hidetak  I remember seeing an email (or something else) about the mqtt failures. I am guessing that this is being fixed

danielpeintner:
> When I, on my PC, run `npm run test` in the root directory, I get an error with binding-mqtt, but I was able to confirm that the error does not occur when I remove it and run it.

I do not see MQTT error on my local machine / setup but recently we do see it on GitHub actions.
Hence, please don't worry about that... this is unrelated and tracked in https://github.com/eclipse/thingweb.node-wot/issues/434

packages/binding-wotfirestore/README.md
hidetak: @egekorkan 
Thank you for your comment.
I understood that you advised that I should describe the support methods about Consumed Thing.
In the table that I made, the method that Consumed Thing and Exposed Thing have was described in the same table, but because there is a difference in the method that Consumed Thing and Exposed Thing can use as shown below, I thought that it was not an appropriate table.
https://www.w3.org/TR/wot-scripting-api/#the-exposedthing-interface

I'm going to describe the methods that Consumed Thing supports.

egekorkan: @hidetak thank you for putting these. That is exactly what I had meant! :)

hidetak: @egekorkan 
I added explanation supported method for Client(Consumed Thing).
https://github.com/eclipse/thingweb.node-wot/pull/417/files#diff-a4441d0ef49c86a2f95d61efe5a084b05169a1339cd419f0df1295733e436890R202
The explanation about the Server has been removed. This is because I don't understand clearly what methods should be supported by Exposed Thing.

egekorkan: So I would say for an exposed thing, it is about whether your implementation can produce an Exposed Thing that has, let's say, that has writeable properties, i.e. a form that has op value of `writeproperty`.

hidetak: @egekorkan 
Thank you for giving comment.
If that is the case, I believe that I can express it in this table.

danielpeintner:
I wanted to give it a try but I noticed a minor issue (caused by us). Could you add a `destroy()` method call...

```
Error: src/wotfirestore-server.ts(30,22): error TS2420: Class 'WoTFirestoreServer' incorrectly implements interface 'ProtocolServer'.
  Property 'destroy' is missing in type 'WoTFirestoreServer' but required in type 'ProtocolServer'.
```

We moved on and require a destroy to be able to remove a thing again from a binding. See for example HTTP https://github.com/eclipse/thingweb.node-wot/blob/1de334c65b3cfc4248b69fd65593eb14d5ca059a/packages/binding-http/src/http-server.ts#L236-L254

hidetak:
@danielpeintner 
Yes, I will try adding destroy.

hidetak:
@danielpeintner @egekorkan 
I am trying to review the implementation of property operations after checking the following sequence diagram.
https://www.w3.org/TR/wot-binding-templates/#appendix-example-sequences

In the case of `A.1.1 Read property`, the Handler registered in setPropertyReadHandler of ExposedThing will be called, and in the case of `A.1.2 Write property`, the Handler registered in setPropertyWriteHandler will be called. 
In the case of `A.1.3 Observe property`, I believe that the Observer is registered with setPropertyObserveHandler, and the change in the property is communicated with the emitPropertyChange method. However, this method is currently not implemented, so I believe that it is currently not possible to communicate property changes.
When are these methods for ExposedThing you told me about before going to be implemented?
https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-03-online-f2f/2021-03-25-WoT-F2F-Scripting%20API%20status.pdf

packages/binding-wotfirestore/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "build": "tsc",
     "rebuild": "tsc & npm rebuild --unsafe-perm --build-from-source",
-    "test": "firebase emulators:start -c test/firebase.json & node -e 'setTimeout(()=>{}, 5000)' && mocha --require ts-node/register test/**/*.ts --timeout 10000",
+    "test": "firebase emulators:start -c test/firebase.json & node -e 'setTimeout(()=>{}, 5000)' && mocha --require ts-node/register test/**/*.ts --timeout 10000 --exit && kill-port 4000 8088 9099",
```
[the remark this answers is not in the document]
hidetak: @maxim-s-barabash
Thank you for your advice.
I have changed it to use `firebase emulators:exec`.
https://github.com/eclipse/thingweb.node-wot/pull/417/commits/48d23d3b53e1c0658d69bf7a4af8f4d67dae1cf5#diff-96c636ab9ce506f7246db2f467c9ff93fd2ef526044f9150391862cba2a2e420R10

danielpeintner:
In the [`v0.8.x` branch](https://github.com/eclipse/thingweb.node-wot/tree/v0.8.x) new parallel work is going on.

Once we feel comfortable we plan to move **v0.8.x** work to master. Having said that, many alignments and new features are planned for `v0.8.x`.

Does this answer you question.

Anyhow, I suggest to finish this binding in the current `master` and continue future work w.r.t. the new API later.

Strangely enough binding-wotfirestore tests succeeded for some settings but fail for ubuntu 10.x, see https://github.com/eclipse/thingweb.node-wot/pull/417/checks?check_run_id=3281420984
`ReferenceError: TextDecoder is not defined`

hidetak:
@danielpeintner 
Thank you for your response.

> Does this answer you question.

I looked at the v0.8.x branch, but I thought that currently some of the following ExposedThing methods are not implemented(e.g. emitPropertyChange and setPropertyObserverHandler,,).
https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-03-online-f2f/2021-03-25-WoT-F2F-Scripting%20API%20status.pdf
Will v0.8.x eventually have all of the above ExposedThing methods?

> Anyhow, I suggest to finish this binding in the current master and continue future work w.r.t. the new API later.

Thank you very much.
I don't think the implementation of Property's Observer and Property's ReadHandler are in line with the specification, but I'd like to aim to release them if you like as issues to be resolved in the future.

> ReferenceError: TextDecoder is not defined

I took a quick look and it seems that NodeJS v10 causes this problem.
I'll see if there is a workaround.

egekorkan:
Regarding emitting property changes, writing to the property internally triggers a property update in the current version.

hidetak:
@egekorkan 
Thank you for your comment.

> Regarding emitting property changes, writing to the property internally triggers a property update in the current version.

As for the current behavior, I believe you are right.
In the future, I think writeProperty for ExposedThing will be deprecated, so emitPropertyChange will be necessary.

hidetak:
I have modified it so that the test will also work with nodejs v10.

hidetak:
@danielpeintner 
Thank you for merging !
I haven't changed the package name and versions yet, I will make new PR.

PR #418, opened 2021-06-01
Modbus alligment

relu91:
This is a joint work of @fillobotto and me. The goal was to align the implementation of the Modbus binding to what we are describing in https://github.com/w3c/wot-binding-templates/pull/109 . Currently we choose to handle special number types (i.e., short, unsigned bytes) using the `@type` keyword, a method that was found not correct during the last TD call. I'll open an issue for that and we'll tackle it in the next PR. 

Relevant changes:
- use length and offset instead of range
- handling endianness using a content-type parameter
- handling all number types using `@type`


examples/scripts/modbus-thing.jsonld
```diff
@@ -27,7 +27,8 @@
             "writeproperty"
           ],
           "modbus:function": 5,
-          "modbus:range": [0, 1],
+          "modbus:offset": 0,
+          "modbus:length": 1,
           "modbus:unitID": 1
```
sebastiankb, line 32: I think, a good best practice would be that the unitID would be a good place in the `href` or better in the `base` since it provides information about the generic location of the device.

relu91: I moved everything in href. The base is another good place, but maybe for the example, it is better to explicitly show the URL in the form.

.vscode/tasks.json
```diff
@@ -16,4 +16,4 @@
             "taskName": "test"
         }
     ]
-} 
\ No newline at end of file
+}
```
danielpeintner: Nit: since I struggled over it

It is often considered _good practice_ to have a newline at end of a file.
Historically, it caused some issues not doing so.

relu91: Thank you but the file is not among the changes anymore. 795417e should have reverted it. Right?

danielpeintner: As said, I do not feel strong about 👍 
Either way is fine

danielpeintner:
@sebastiankb good to go?

danielpeintner (approved):
LGTM as far as I can tell but let's wait till @sebastiankb has time to look at it

sebastiankb:
looks quite good. I think, the README needs to be updated with the latest features. E.g., I  miss the definitions and explaintion of 'BIG_ENDIAN', 'LITTLE_ENDIAN', 'BIG_ENDIAN_BYTE_SWAP', and 'LITTLE_ENDIAN_BYTE_SWAP'.

sebastiankb (approved):

danielpeintner:
> looks quite good. I think, the README needs to be updated with the latest features. E.g., I miss the definitions and explaintion of 'BIG_ENDIAN', 'LITTLE_ENDIAN', 'BIG_ENDIAN_BYTE_SWAP', and 'LITTLE_ENDIAN_BYTE_SWAP'.

@relu91 do you want to push updates w.r.t. README to this PR or a follow-up PR?

relu91:
> @relu91 do you want to push updates w.r.t. README to this PR or a follow-up PR?

Let's do this PR I think they are related changes.

sebastiankb:
>Let's do this PR I think they are related changes.

I do not have any preference. Maybe its easy to do this here directly.

sebastiankb:
@relu91 many thanks

PR #423, opened 2021-06-08
fix: ExposedThingInit type and HTTP tests

danielpeintner:


packages/core/src/protocol-helpers.ts
```diff
@@ -20,7 +20,7 @@ import { ReadableStream } from 'web-streams-polyfill/ponyfill/es2018';
 export default class ProtocolHelpers {
 
   // set contentType (extend with more?)
-  public static updatePropertyFormWithTemplate(form: TD.Form, tdTemplate: WoT.ThingDescription, propertyName: string) {
+  public static updatePropertyFormWithTemplate(form: TD.Form, tdTemplate: WoT.ExposedThingInit, propertyName: string) {
```
relu91, line 23: shouldn't be `tdTemplate?: WoT.ExposedThingInit` I'am wondering why is working right now. Cause other parts of the code base, we are passing something that might be undefined...

packages/core/src/protocol-interfaces.ts
```diff
@@ -55,7 +55,7 @@ export interface ProtocolClientFactory {
 
 export interface ProtocolServer {
   readonly scheme: string;
-  expose(thing: ExposedThing, tdTemplate?: WoT.ThingDescription): Promise<void>;
+  expose(thing: ExposedThing, tdTemplate?: WoT.ExposedThingInit): Promise<void>;
```
relu91, line 58: Maybe we can use the default pattern instead of undefined: `tdTemplate?: WoT.ExposedThingInit = {}`

relu91 (approved):
Looks good! just some minor comments 👍🏻

PR #425, opened 2021-06-09
V0.8.x NextSteps 7

danielpeintner:


packages/core/test/ServerTest.ts
```diff
@@ -211,17 +220,26 @@ class WoTServerTest {
...
         expect(thing).to.have.property("properties").to.have.property("my number").to.have.property("readOnly").that.equals(false);
         expect(thing).to.have.property("properties").to.have.property("my number").to.have.property("observable").that.equals(false);
 
-        let value1 = await thing.readProperty("my number");
-        expect(value1).to.equal(1);
+        // XXX How to to check handlers properly with some type-safety?
+        let expThing: any = thing;
+        let propertyState = expThing.properties["my number"].getState();
+        let ff = await propertyState.readHandler();
+        expect(ff).to.equal(1);
```
danielpeintner, line 238: @relu91 @egekorkan 
Since `ExposedThing` does no longer extend `ConsumedThing` I wonder how we could best test the server part.

A solution as the one above is *VERY* flucky and requires to know the internals. Do you have a good advice what else could be done?

PR #428, opened 2021-06-10
Streaming errors with new APis

relu91:
This PR is mainly focused on fixing errors with the new stream interface. Now the counter example should work as expected.


danielpeintner (approved):
LGTM

checked locally also

Issue #427, opened 2021-06-10
td variable passed vs template - object modification 


danielpeintner:
https://github.com/eclipse/thingweb.node-wot/blob/7d3217984004f74d3daabb1ce158af5c441a3371/packages/core/src/wot-impl.ts#L88-L90

we should use the variable `template` instead of `td`.
This *bug* was introduced in v0.8.x ... as long as there is no copy of the object it is fine as it is but for sanity we should change it.

Moreover, I wonder whether we should do a deep copy of the object before modifying it OR we do `addDefaultLanguage(template)` later where we clone the object in `ExposedThing` constructor...

Any opinion?

It is correct in v0.7.x

https://github.com/eclipse/thingweb.node-wot/blob/5ffd659a2502657151ad4927dacab3849a09a060/packages/core/src/wot-impl.ts#L89-L91

danielpeintner:
fixed by https://github.com/eclipse/thingweb.node-wot/pull/429

Issue #426, opened 2021-06-10
Handling Thing Models


egekorkan:
a TM/Partial TD that is passed into the produce method can contain `"@type":"tm:ThingModel"`. This should be however removed when the TD is produced as mentioned by @relu91 at 

>The tm:ThingModel value of the top-level @type MUST to be replaced by the value Thing in the Partial TD instance.

Also, theoritically, this should happen when the expose method is called, I think. Also, what is passed into produce does not have to contain `@type` keyword or it can contain but its value can contain something else, like `saref:Sensor`. 
Nothing difficult but a simple replacement without some if checks will result in wrong behavior.

danielpeintner:
Yes, the place might be close to 

https://github.com/eclipse/thingweb.node-wot/blob/5ffd659a2502657151ad4927dacab3849a09a060/packages/core/src/wot-impl.ts#L89-L90

danielpeintner:
https://github.com/eclipse/thingweb.node-wot/pull/429/commits/1f7bdde1000eeb3f778b7a0a96564ec3143c0a43 should fix the issue for the upcoming API.

We still have v.07.x .. I guess we need to change it there also.. it gets now more complicated to keep up several branches...

danielpeintner:
Note: `@type` can be an array, so this should be handled also

relu91:
Thanks, Ege for posting the issue! Although the suggestion would improve the situation I am afraid is not the final solution to handle TMs. If we allow TMs in the `produce` method this means that it should be also able to handle tm:ref and tm:extend right? We discussed having full support for TMs in `produce` function in the Scripting API call but the final agreement was that TMs should be first pre-processed by **external-service** and then passed to the produce. 

See the long discussion here: https://github.com/w3c/wot-scripting-api/issues/304

egekorkan:
Well I agree with this that this is what should be done but nothing stops a person from entering a TM into the produce method. One solution would be to detect this and fail early saying that this is not allowed. node-wot should not generate invalid TDs no matter what in my opinion

relu91:
> Well I agree with this that this is what should be done but nothing stops a person from entering a TM into the produce method. One solution would be to detect this and fail early saying that this is not allowed. node-wot should not generate invalid TDs no matter what in my opinion

[...]

PR #429, opened 2021-06-11
Single place of input modifications

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/427
~~fixes https://github.com/eclipse/thingweb.node-wot/issues/426 for v.08.x~~

### DIFF
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -68,6 +68,10 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
         let clonedModel = JSON.parse(JSON.stringify(thingModel))
         Object.assign(this, clonedModel);
 
+        // unset "@type":"tm:ThingModel" (see https://github.com/eclipse/thingweb.node-wot/issues/426)
+        if (this["@type"] && this["@type"] === "tm:ThingModel") {
+            delete this["@type"];
+        }
         // set default language
         this.addDefaultLanguage(this);
         // extend interactions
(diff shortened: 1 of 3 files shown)


packages/core/src/exposed-thing.ts
```diff
@@ -67,9 +67,48 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
+                    if (arr[i] === "tm:ThingModel") {
+                        arr.splice(i, 1);
+                        i--;
+                    }
+                }
+            }
+        }
+        // set default language
+        this.addDefaultLanguage(this);
+        // extend interactions
```
relu91, line 87: I would approve the change but see my comment: https://github.com/eclipse/thingweb.node-wot/issues/426#issuecomment-859660742

relu91 (commented):
PR looks good but sadly we have to wait until the discussion on https://github.com/w3c/wot-scripting-api/issues/304 is settled

danielpeintner:
I think this PR should resolve #427.

Having said that, we might need to continue discussing #426.
I kept the _current_ code for #426 in, by commenting it out. Not sure whether we should totally remove it as part of this PR. Any opinion?

relu91 (approved):
Yeah, I agree, I would even uncomment the tests cause they provide useful information anyway. But it is fine, we will fix it when #426 is settled

danielpeintner:
@relu91 thanks for the feedback. Let's merge this PR and resolve #426 related issues in a later PR (the *initial* code and the commented tests are there...)

Issue #410, opened 2021-04-19
OPC UA Issues


sebastiankb:
I just used the OPC UA binding and faced some issues:

- if I use a property address like `opc.tcp://localhost:4840/ns=3;s="Case_Lamp_Variable"` the UA binding transforms the address to `opc.tcp://localhost:4840/ns=3;s=%22Case_Lamp_Variable%22`. The node-opcua package is not able to process such kind of string pattern and needs more the '"' character again. A quick fix can be done in extract_params function in opcua-client.ts  
- when I initiated a write property I received an error 'The server does not support writing the combination of value, status and timestamps provided.' . In `writeResource` I had to delate the `sourceTimestamp` and `statusCode` of the `value` variable.
- It seems that in `writeResource` the  UA data type and its corresponding ID is not provided properly (e.g., see [here](https://node-opcua.github.io/api_doc/2.0.0/enums/datatype.html)). 
- Within the `nodeToWrite` in `writeResource` the full `payload` is assigned. However, it should be more `payload.payload`

For my application I found a quick fix, but it should be solved generically.

danielpeintner:
> - if I use a property address like `opc.tcp://localhost:4840/ns=3;s="Case_Lamp_Variable"` the UA binding transforms the address to `opc.tcp://localhost:4840/ns=3;s=%22Case_Lamp_Variable%22`. The node-opcua package is not able to process such kind of string pattern and needs more the '"' character again. A quick fix can be done in extract_params function in opcua-client.ts  

Mhh, the url is encoded, see https://www.urldecoder.org

I still think this is the right process. Hence I wonder whether it is
-  an issue of the node-opcua package and we should report it or
- we should simply url-decode before passing it to the library.

Any thoughts?

With regards to the other aspects. Who is able to work on the OPCUA aspects?

lukesmolo:
Hi guys, thank you for reporting the problems you encountered. I made some fixes and I am going to make a PR.
Regarding the four points:

- I added a _decode_ operation inside the extract_params function, so we are sure that the string is passed as decoded to the library. I added some tests for this
- for the moment I removed the **timestamp** and **statusCode** values from the payload. I guess they should become optional as soon as the form will accept it
- I made a fix and added some further tests on purpose
- I made a fix

Please let me know if something else should be fixed/improved.

sebastiankb:
[...]

PR #433, opened 2021-06-30
fix: fix for opcua issues eclipse/thingweb.node-wot#410. Node-opcua l…

lukesmolo:
Fix for OPCUA issues #410

### DIFF
--- a/examples/servients/opcua-cli/package.json
+++ b/examples/servients/opcua-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@node-wot/opcua-cli",
-  "version": "0.7.6",
+  "version": "0.7.7",
   "description": "servient command line interface",
   "author": "Eclipse Thingweb <thingweb-dev@eclipse.org> (https://thingweb.io/)",
   "license": "EPL-2.0 OR W3C-20150513",
@@ -24,9 +24,9 @@
     "wot-typescript-definitions": "0.7.3"
   },
   "dependencies": {
-    "@node-wot/binding-file": "0.7.5",
-    "@node-wot/binding-http": "0.7.5",
-    "@node-wot/binding-opcua": "0.7.5",
+    "@node-wot/binding-file": "^0.7.6",
+    "@node-wot/binding-http": "^0.7.6",
+    "@node-wot/binding-opcua": "^0.7.6",
     "@types/node-fetch": "^2.5.10",
     "fs": "0.0.1-security",
     "path": "^0.12.7"
--- a/examples/servients/opcua-cli/package.json
+++ b/examples/servients/opcua-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@node-wot/opcua-cli",
-  "version": "0.7.2",
+  "version": "0.7.6",
   "description": "servient command line interface",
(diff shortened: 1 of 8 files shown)


examples/servients/opcua-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@node-wot/opcua-cli",
-  "version": "0.7.2",
+  "version": "0.7.6",
```
danielpeintner, line 3: We are about to move to "0.7.7" ... can you update this as well. Thanks!

examples/servients/opcua-cli/package.json
```diff
@@ -18,16 +18,16 @@
...
     "wot-typescript-definitions": "0.7.3"
   },
   "dependencies": {
-    "@node-wot/core": "0.7.2",
-    "@node-wot/binding-http": "0.7.2",
-    "@node-wot/binding-file": "0.7.2",
-    "@node-wot/binding-opcua": "0.7.2",
+    "@node-wot/binding-file": "0.7.5",
+    "@node-wot/binding-http": "0.7.5",
+    "@node-wot/binding-opcua": "0.7.5",
```
danielpeintner, line 29: here it should be "0.7.6"

danielpeintner:
Thanks @lukesmolo 
I noted that an OPCUA test was failing..

```
  1) OPCUA client test
       should read a property:

      AssertionError: expected 2 to equal 1
      + expected - actual

      -2
      +1
```

examples/servients/opcua-cli/package.json
lukesmolo: ok

examples/servients/opcua-cli/package.json
lukesmolo: ok, sure

lukesmolo:
> Thanks @lukesmolo
> I noted that an OPCUA test was failing..
> 
> ```
>   1) OPCUA client test
>        should read a property:
> 
>       AssertionError: expected 2 to equal 1
>       + expected - actual
> 
>       -2
>       +1
> ```
Damn, I got it, I think there is a race condition since for instance, I cannot see the same issue in my local environment or also [here](https://github.com/lukesmolo/thingweb.node-wot/runs/2951271996). The _write Property_ arrives before the _read_  one. Anyway, let me fix it.

examples/servients/opcua-cli/package.json
relu91: > here it should be "0.7.6"

Did you mean 0.7.7 ?

danielpeintner: In  this section we seem to be sticking to the _released_ version that currently is 0.7.6
Maybe it is even easier to have something like `0.7.x` here

relu91 (approved):

danielpeintner (approved):

danielpeintner:
@sebastiankb since you raised https://github.com/eclipse/thingweb.node-wot/issues/410
Can you have a look

sebastiankb:
@lukesmolo many thanks. I will test this PR with a S7 and let you know if its working now.

sebastiankb (approved):

sebastiankb:
Just test the new version with the S7 and indeed, seems to work now :-)

@lukesmolo many thanks again

PR #435, opened 2021-07-02
Use InteractionInput when needed

relu91:
As the title say this pr aims to clean up old code and use the newly defined interface for affordance inputs.


danielpeintner (approved):
LGTM

PR #438, opened 2021-07-07
Implement functionality to destroy (remove) a thing

### LINKED ISSUE #431: How can i stop a servient? - Error on shutdown with MQTT binding
I am trying to stop my servient with the following function:

`servient.shutdown()`

but it throws the following error:

`(node:10160) UnhandledPromiseRejectionWarning: TypeError: _this.broker.stop is not a function
    at C:\Users\alvar\Desktop\Servient\node_modules\@node-wot\binding-mqtt\dist\mqtt-broker-server.js:244:26
    at new Promise (<anonymous>)
    at MqttBrokerServer.stop (C:\Users\alvar\Desktop\Servient\node_modules\@node-wot\binding-mqtt\dist\mqtt-broker-server.js:241:16)
    at C:\Users\alvar\Desktop\Servient\node_modules\@node-wot\core\dist\servient.js:212:64
    at Array.forEach (<anonymous>)
    at Servient.shutdown (C:\Users\alvar\Desktop\Servient\node_modules\@node-wot\core\dist\servient.js:212:22)
    at C:\Users\alvar\Desktop\Servient\air.js:174:22
    at Layer.handle [as handle_request] (C:\Users\alvar\Desktop\Servient\node_modules\express\lib\router\layer.js:95:5)
    at next (C:\Users\alvar\Desktop\Servient\node_modules\express\lib\router\route.js:137:13)
    at Route.dispatch (C:\Users\alvar\Desktop\Servient\node_modules\express\lib\router\route.js:112:3)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:10160) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:10160) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.`

So my question is, what is the right way to stop the servient from running?
Thanks in advance.

danielpeintner:
fixes #431

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -164,10 +164,23 @@ export default class CoapServer implements ProtocolServer {
     });
   }
 
-  public destroy(thingId: string): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
-      // TODO destroy the thing with the given id
-      resolve();
+  public destroy(thingId: string): Promise<boolean> {
+    console.debug("[binding-coap]", `CoapServer on port ${this.getPort()} destroying thingId '${thingId}'`);
+    return new Promise<boolean>((resolve, reject) => {
+      let removedThing: ExposedThing = undefined;
+      for (let name of Array.from(this.things.keys())) {
+        let expThing = this.things.get(name);
+        if (expThing != null && expThing.id != null && expThing.id === thingId) {
+          this.things.delete(name);
+          removedThing = expThing;
+        }
+      }
+      if (removedThing) {
+        console.info("[binding-coap]", `CoapServer succesfully destroyed '${removedThing.title}'`);
+      } else {
+        console.info("[binding-coap]", `CoapServer failed to destroy thing with thingId '${thingId}'`)
+      }
+      resolve(removedThing != undefined);
     });
   }
 
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -170,7 +170,7 @@ export default class CoapServer implements ProtocolServer {
       let removedThing: ExposedThing = undefined;
       for (let name of Array.from(this.things.keys())) {
         let expThing = this.things.get(name);
-        if (expThing != null && expThing.id != null && expThing.id === thingId) {
+        if (expThing?.id === thingId) {
           this.things.delete(name);
           removedThing = expThing;
         }
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -164,6 +164,13 @@ export default class CoapServer implements ProtocolServer {
     });
   }
 
+  public destroy(thingId: string): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      // TODO destroy the thing with the given id
+      resolve();
+    });
+  }
+
   private handleRequest(req: any, res: any) {
     
     console.debug("[binding-coap]",`CoapServer on port ${this.getPort()} received '${req.method}(${req._packet.messageId}) ${req.url}' from ${Helpers.toUriLiteral(req.rsinfo.address)}:${req.rsinfo.port}`);
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -233,10 +233,23 @@ export default class HttpServer implements ProtocolServer {
     }
   }
 
-  public destroy(thingId: string): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
-      // TODO destroy the thing with the given id
-      resolve();
+  public destroy(thingId: string): Promise<boolean> {
+    console.debug("[binding-http]", `HttpServer on port ${this.getPort()} destroying thingId '${thingId}'`);
+    return new Promise<boolean>((resolve, reject) => {
+      let removedThing: ExposedThing = undefined;
+      for (let name of Array.from(this.things.keys())) {
+        let expThing = this.things.get(name);
+        if (expThing != null && expThing.id != null && expThing.id === thingId) {
+          this.things.delete(name);
+          removedThing = expThing;
+        }
+      }
+      if (removedThing) {
+        console.info("[binding-http]", `HttpServer succesfully destroyed '${removedThing.title}'`);
+      } else {
+        console.info("[binding-http]", `HttpServer failed to destroy thing with thingId '${thingId}'`)
+      }
+      resolve(removedThing != undefined);
     });
   }
 
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -233,6 +233,13 @@ export default class HttpServer implements ProtocolServer {
     }
   }
 
+  public destroy(thingId: string): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      // TODO destroy the thing with the given id
+      resolve();
+    });
+  }
+
   public addEndpoint(thing: ExposedThing, tdTemplate: WoT.ThingDescription, base: string) {
       for (let type of ContentSerdes.get().getOfferedMediaTypes()) {
 
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -239,7 +239,7 @@ export default class HttpServer implements ProtocolServer {
       let removedThing: ExposedThing = undefined;
       for (let name of Array.from(this.things.keys())) {
         let expThing = this.things.get(name);
-        if (expThing != null && expThing.id != null && expThing.id === thingId) {
+        if (expThing?.id === thingId) {
           this.things.delete(name);
           removedThing = expThing;
         }
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -272,10 +272,10 @@ export default class MqttBrokerServer implements ProtocolServer {
     });
   }
 
-  public destroy(thingId: string): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
+  public destroy(thingId: string): Promise<boolean> {
+    return new Promise<boolean>((resolve, reject) => {
       // TODO destroy the thing with the given id
-      resolve();
+      resolve(false);
     });
   }
 
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -272,6 +272,13 @@ export default class MqttBrokerServer implements ProtocolServer {
     });
   }
 
+  public destroy(thingId: string): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      // TODO destroy the thing with the given id
+      resolve();
+    });
+  }
+
   public start(servient: Servient): Promise<void> {
     return new Promise<void>((resolve, reject) => {
 
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -273,9 +273,22 @@ export default class MqttBrokerServer implements ProtocolServer {
   }
 
   public destroy(thingId: string): Promise<boolean> {
+    console.debug("[binding-mqtt]", `MqttBrokerServer on port ${this.getPort()} destroying thingId '${thingId}'`);
     return new Promise<boolean>((resolve, reject) => {
-      // TODO destroy the thing with the given id
-      resolve(false);
+      let removedThing: ExposedThing = undefined;
+      for (let name of Array.from(this.things.keys())) {
+        let expThing = this.things.get(name);
+        if (expThing != null && expThing.id != null && expThing.id === thingId) {
+          this.things.delete(name);
+          removedThing = expThing;
+        }
+      }
+      if (removedThing) {
+        console.info("[binding-mqtt]", `MqttBrokerServer succesfully destroyed '${removedThing.title}'`);
+      } else {
+        console.info("[binding-mqtt]", `MqttBrokerServer failed to destroy thing with thingId '${thingId}'`)
+      }
+      resolve(removedThing != undefined);
     });
   }
 
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -98,8 +98,7 @@ export default class MqttBrokerServer implements ProtocolServer {
 
     console.debug("[binding-mqtt]",`MqttBrokerServer at ${this.brokerURI} exposes '${thing.title}' as unique '${name}/*'`);
     return new Promise<void>((resolve, reject) => {
-
-      // clean-up on destroy and stop
+      
       this.things.set(name, thing);
 
       for (let propertyName in thing.properties) {
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -99,7 +99,7 @@ export default class MqttBrokerServer implements ProtocolServer {
     console.debug("[binding-mqtt]",`MqttBrokerServer at ${this.brokerURI} exposes '${thing.title}' as unique '${name}/*'`);
     return new Promise<void>((resolve, reject) => {
 
-      // TODO clean-up on destroy and stop
+      // clean-up on destroy and stop
       this.things.set(name, thing);
 
       for (let propertyName in thing.properties) {
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -107,7 +107,14 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
 
     /** @inheritDoc */
     destroy(): Promise<void> {
+        console.debug("[core/exposed-thing]",`ExposedThing '${this.title}' destroying the thing and its interactions`);
+
         return new Promise<void>((resolve, reject) => {
+            this.getServient().removeThing(this.id);
+            // remove alle listeners and such?
+
+            // inform TD observers that thing is gone?
+            this.getSubjectTD().next(null);
             resolve();
         });
     }
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -110,23 +110,25 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
         console.debug("[core/exposed-thing]", `ExposedThing '${this.title}' destroying the thing and its interactions`);
 
         return new Promise<void>((resolve, reject) => {
-            this.getServient().removeThing(this.id);
-            // indicate possible subscriptions that subject has been completed
-            for (let propertyName in this.properties) {
-                let ps: PropertyState = this.properties[propertyName].getState();
-                if (ps.subject) {
-                    ps.subject.complete();
+            this.getServient().destroyThing(this.id).then(() => {
+                // indicate to possible subscriptions that subject has been completed
+                for (let propertyName in this.properties) {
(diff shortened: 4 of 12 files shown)


danielpeintner:
NoteA: the destroy() call reports a boolean whether a given thingId could have been removed 
NoteB: I think we need to look into details like any pending resources (subscriptions et cetera)

Do you think anything else is missing?

egekorkan:
For MQTT we should do unsubscriptions or simply disconnect from the broker. Regarding disconnecting, I am not entirely sure what happens if we have two Things since the servient connects to the broker but the Things are exposed/destroyed.

danielpeintner:
> For MQTT we should do unsubscriptions or simply disconnect from the broker. Regarding disconnecting, I am not entirely sure what happens if we have two Things since the servient connects to the broker but the Things are exposed/destroyed.

This is exactly the thing where I need help since I am not very familiar with MQTT.

w.r.t. the other bindings i will try to look into and see what else should be done.. like _unsubscribing_ ..

packages/binding-coap/src/coap-server.ts
```diff
@@ -164,6 +164,26 @@ export default class CoapServer implements ProtocolServer {
     });
   }
 
+  public destroy(thingId: string): Promise<boolean> {
+    console.debug("[binding-coap]", `CoapServer on port ${this.getPort()} destroying thingId '${thingId}'`);
+    return new Promise<boolean>((resolve, reject) => {
+      let removedThing: ExposedThing = undefined;
+      for (let name of Array.from(this.things.keys())) {
+        let expThing = this.things.get(name);
+        if (expThing != null && expThing.id != null && expThing.id === thingId) {
```
relu91, line 173: I think you can compact this into:
`if(expThing?.id === thingId)`

packages/binding-http/src/http-server.ts
```diff
@@ -233,6 +233,26 @@ export default class HttpServer implements ProtocolServer {
     }
   }
 
+  public destroy(thingId: string): Promise<boolean> {
+    console.debug("[binding-http]", `HttpServer on port ${this.getPort()} destroying thingId '${thingId}'`);
+    return new Promise<boolean>((resolve, reject) => {
+      let removedThing: ExposedThing = undefined;
+      for (let name of Array.from(this.things.keys())) {
+        let expThing = this.things.get(name);
+        if (expThing != null && expThing.id != null && expThing.id === thingId) {
```
relu91, line 242: See the comment above

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -99,7 +99,7 @@ export default class MqttBrokerServer implements ProtocolServer {
     console.debug("[binding-mqtt]",`MqttBrokerServer at ${this.brokerURI} exposes '${thing.title}' as unique '${name}/*'`);
     return new Promise<void>((resolve, reject) => {
 
-      // TODO clean-up on destroy and stop
+      // clean-up on destroy and stop
```
relu91, line 102: Isn't this solved with this PR?

packages/core/src/protocol-interfaces.ts
```diff
@@ -55,6 +55,7 @@ export interface ProtocolClientFactory {
 export interface ProtocolServer {
   readonly scheme: string;
   expose(thing: ExposedThing, tdTemplate?: WoT.ThingDescription): Promise<void>;
+  destroy(thingId: string): Promise<boolean>; // reports success
```
relu91, line 58: I am wondering if it makes more sense to throw if the thing was not destroyed instead of returning false. Consider also if we just return false we can't handle the causes from the code, but we probably need to check the logs to understand why.

packages/core/src/servient.ts
```diff
@@ -160,6 +160,18 @@ export default class Servient {
         }
     }
 
+    public removeThing(thingId: string): boolean {
+        if (this.things.has(thingId)) {
+            this.servers.forEach( (server) => {
+                server.destroy(thingId);
```
relu91, line 166: `server.destroy` returns a promise, we should wait until competition otherwise we might create problems down the road

packages/core/test/ServerTest.ts
```diff
@@ -115,6 +119,32 @@ class WoTServerTest {
...
+                }
+            }
+        }`;
+        let thing = await WoTServerTest.WoT.produce(JSON.parse(desc));
+        expect(thing).to.exist;
+        // test TD
+        expect(thing.getThingDescription()).to.have.property("title").to.equal("myDestroyThing");
+        expect(thing.getThingDescription()).to.have.property("id").to.equal("1234567");
+        // test presence (and destroy)
+        if (WoTServerTest.servient.getThing("1234567") == null) {
```
relu91, line 139: Why don't use `expect` here? Fore example: `expect(WoTServerTest.servient.getThing("1234567")).to.not.be.null`

packages/core/test/ServerTest.ts
```diff
@@ -115,6 +119,32 @@ class WoTServerTest {
...
+            fail("Thing not properly added");
+        }
+        await thing.destroy(); // destroy -> remove
+        if (WoTServerTest.servient.getThing("1234567") != null) {
```
relu91, line 143: Same as above

relu91 (commented):
I think the PR is in a good shape, but I added just minor questions/suggestions

packages/binding-mqtt/src/mqtt-broker-server.ts
danielpeintner: Yes, that's why I removed the TODO. Not sure about the intent of the original author but for me it is just a reminder that the following lines need to cleaned-up in case of destroy. Do you suggest removing the comment entirely?

packages/core/src/protocol-interfaces.ts
danielpeintner: Mhh, the reason why I came up with a boolean is that some bindings might not even support destroying (for example some bindings like websocket/oracle do not have an implementation yet and can indicate that destroy did not work)

Moreover, the servient that calls all bindings to destroy() a given thing can decide what do do.. throwing to me means a failure happened while `false` to me means it did not work...
Does that make sense?

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: Ah ok, now I see the point! then it can stay 👍🏻

packages/core/test/ServerTest.ts
danielpeintner: makes sense

packages/core/test/ServerTest.ts
danielpeintner: makes sense

packages/binding-coap/src/coap-server.ts
danielpeintner: +1 makes sense (old school developer ;-))

packages/binding-http/src/http-server.ts
danielpeintner: makes sense

packages/core/src/protocol-interfaces.ts
relu91: It is reasonable, we just need to think about it thoughtfully. So from the servient point of (the call of each `destroy` function), I can expect two unwanted outcomes:
- The function returned `false`
- The function throws an error

Can we define two different behaviors for those conditions? From my point of view, I would just assume that the service that I wanted (removing the thing) was not performed. So it feels like an error state, isn't it? 

About the `not implemented` situation: does it makes sense to have bindings that do not ever implement the `destroy` functionality? Otherwise, returning `false` is just a //TODO: placeholder, right?

packages/binding-mqtt/src/mqtt-broker-server.ts
danielpeintner: To avoid confusion let's remove the comment entirely

packages/core/src/servient.ts
danielpeintner: +1

danielpeintner: Question: `removeThing()` in servient.ts is a sync call while `server.destroy()` is async call.
What is the best way to handle this or to put it differently how to wait for all promises to be returned. Any proposal @relu91

packages/core/src/protocol-interfaces.ts
danielpeintner: > Can we define two different behaviors for those conditions? From my point of view, I would just assume that the service that I wanted (removing the thing) was not performed. So it feels like an error state, isn't it?

Let's assume the binding.destroy() failed (due to what every reason). Should the servient.destroy() remove the thing from the list or not? I am not sure here either ....

> About the `not implemented` situation: does it makes sense to have bindings that do not ever implement the `destroy` functionality? Otherwise, returning `false` is just a //TODO: placeholder, right?

is possible that a binding can expose a thing but not destroying it again. Not sure.

packages/core/src/servient.ts
danielpeintner: What I have now obviously doesn't work. The return does not return the outer call.

https://github.com/eclipse/thingweb.node-wot/blob/36029d479ab50279115cfb4a680f01f65c871654/packages/core/src/servient.ts#L167-L174

packages/core/src/protocol-interfaces.ts
relu91: > Let's assume the binding.destroy() failed (due to what every reason). Should the servient.destroy() remove the thing from the list or not? I am not sure here either ....

Hmm not a pleasant state, indeed. I have a feeling that removing it would be wrong... but still if we keep it who knows which is the actual state of the thing... it might be half-working like receiving requests only for one affordance type. In my opinion, we should check what `sockets` do when you close them. 

> is possible that binding can expose a thing but not destroying it again. Not sure.
So it creates it, but not destroys it... IMHO it would be better than if it creates resources for exposing it MUST also clean its own mess.

packages/core/src/servient.ts
danielpeintner: Question: Should the servient.removeThing() return a promise?
Note: servient.addThing() does return a plain boolean

packages/core/src/servient.ts
```diff
@@ -164,14 +164,10 @@ export default class Servient {
...
             this.servers.forEach((server) => { serverPromises.push(server.destroy(thingId)); });
-            Promise.all(serverPromises).then(() => {
-                this.things.delete(thingId);
-                console.warn("[core/servient]", `Servient removed thing with ID '${thingId}'`);
-                return true;
-            }).catch((err) => {
-                console.warn("[core/servient]", `Servient removing thing with ID '${thingId}' failed`);
-                return false;
-            });
+            var results = (async () => await Promise.all(serverPromises));
```
relu91, line 167: I think you need to call the function otherwise you are just creating a lambda function:
`(async () => await Promise.all(serverPromises))()`

danielpeintner: Are you sure?  I tried it locally and after destroy the thing is not available any longer with the current code

packages/core/src/protocol-interfaces.ts
danielpeintner: I am not sure how to proceed here, mhh.
@egekorkan do you have any opinion?

packages/core/src/servient.ts
danielpeintner: I renamed it to `destroyThing()` and returning a promise also.
Moreover, this is a function called in core only.. so we _could_ still change it in future...

packages/core/src/servient.ts
danielpeintner: I renamed it to `destroyThing()` and changed it to return a promise.
I think this resolved all issues... do you agree @relu91 ?

relu91: yes! thank you 👍🏻

packages/core/src/protocol-interfaces.ts
egekorkan: I am not sure in which cases the destroy would not work but having the boolean as return value creates some flexibility in case we want to communicate such cases

danielpeintner: @relu91 @egekorkan  do we agree to keep the interface and the code as is?

egekorkan: I'd say yes for boolean interface above

relu91: > I am not sure in which cases the destroy would not work but having the boolean as return value creates some flexibility in case we want to communicate such cases

Exactly because we are not sure which are the cases when a destroy function would return false (i.e., fail) I would prefer using void. When design functions I follow this rule:

> The rule is pretty simple: Use exceptions on exceptional, erroneous, unpredicted failures. Don't use exceptions when you expect something to happen or when something happens really often.

Are we _expecting_ that the destroy function would return a boolean value? Looking around the close functions return void and throw if something is wrong. I would follow this pattern, also considering the principle of the least surprise. 

[Java](https://docs.oracle.com/javase/7/docs/api/java/io/Closeable.html)
[Nodejs](https://nodejs.org/api/stream.html#stream_writable_destroy_error)

egekorkan: I see your argument and agree with it now. Letting a failure be reported with a boolean=false would also "encourage" developers to simplify/abstract the information related to failure which would not be very helpful to detect the failure.

danielpeintner: It seems I have the opposite opinion 🙈

* I suggest to throw errors if some code part fails or an error occurs
* in this case a binding may choose not support removing/destroying a thing (hence returning false seems ok to me)

Currently we have bindings like fujitsu, oracle, ws that have not implemented an actual `destroy()`. At the moment the call succeeds and returns `false` to indicate no destroy/remove took place. IF we decide to go with throws I think we loose the information that there isn't an issue but it rather did not work...

Similarly, what should happen if a provided thingId was not found. `throw` an error or indicate `false`?

To me it is very similar to removing an entry from a list.
see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#remove-java.lang.Object-
It may return boolean or throw in case of failure.

danielpeintner: Question: let's assume we go with void and `ExposedThing.destroy()` walks over all bindings and calls `destroy(thingId)`. What should happen it it fails for one binding?

relu91: > in this case a binding may choose not support removing/destroying a thing (hence returning false seems ok to me)

Most implementations use the famous [NotImplementedExecption](https://softwareengineering.stackexchange.com/questions/323498/why-is-having-a-notimplementedexception-a-good-thing), which is a thing also in TS codebases. 

As I said above a binding implementation must do something in the destroy method, cause in principle an ExposedThing is occupying system resources. Even in ws situations, you might need to close all the open connections, right?

> Similarly, what should happen if a provided thingId was not found. throw an error or indicate false?
>
>Similarly, what should happen if a provided thingId was not found. throw an error or indicate false?
>
>To me it is very similar to removing an entry from a list.
see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#remove-java.lang.Object-
It may return boolean or throw in case of failure.

Now I see, but I would explicitly use the return value only to indicate that a Web Thing was not found. We should write that in jsdoc and not use it for other errors (like not implemented above). Something like:
```js
/**
 * @param thingId: id of the thing to destroy 
 * @returns true if the thing was found and destroyed; false if the thing was not in the list
 * @throws if the binding couldn't destroy completely the thing with thingId.
**/
function destroy(thingId: string): Promise<boolean>;
```

I would still argue that from a consumer point of view the returned value is not so important. For example, if I want to remove an element from the list if it wasn't there I still obtain the effect that the *element is not in the list*. Some way here, if I try to destroy a thing that was not in the binding, well *nothing to be cleaned*. But I guess you got a point.

danielpeintner: Looks very promising to me!
I think we get close to agreement.

Some final questions:
* for the oracle-server I do no not really know what to do fro destroy(). Hence I would like to throw `NotImplementedException`. Is this really a error we should use, not really defined in https://nodejs.org/api/errors.html
* another question (the same as my [comment above](https://github.com/eclipse/thingweb.node-wot/pull/438#discussion_r671114750)). Let's assume ExposedThing.destroy() walks over all bindings and calls destroy(thingId). What should happen it it fails for one binding (e.g., the oracle binding)? The [current code](https://github.com/eclipse/thingweb.node-wot/blob/a39fbc643248863ee5c2d376e1550af7e83b8d32/packages/core/src/exposed-thing.ts#L109-L133) would reject since a error was thrown... or shall we look into the error type and ignore `NotImplementedException` ?

egekorkan: I think that the main argument should be based on how the Scripting level code should look like. If the code can do with the knowledge of the boolean value, we can have it. Mostly, the code wants to simply destroy the Thing and move on and not handle true, false and errors independently no?

relu91: > I think that the main argument should be based on how the Scripting level code should look like. If the code can do with the knowledge of the boolean value, we can have it. Mostly, the code wants to simply destroy the Thing and move on and not handle true, false and errors independently no?

Yes, correct, but here we are discussing on the internal level Core <-> Bindings. `ExposeThing.destroy()` returns `void`. 

> for the oracle-server I do no not really know what to do fro destroy(). Hence I would like to throw NotImplementedException. Is this really a error we should use, not really defined in https://nodejs.org/api/errors.html

I'd go for something like this:
```ts
function _notImplemented() : never {
     throw new Error("Not Implemented")
}
```

> another question (the same as my comment above). Let's assume ExposedThing.destroy() walks over all bindings and calls destroy(thingId). What should happen it it fails for one binding (e.g., the oracle binding)? The current code would reject since a error was thrown... or shall we look into the error type and ignore

Be transparent with the application code if something fails give more control and leave room for customization. I'd throw if _at least_ one binding fails. Like go through every biding, collect errors, if `errors.length >0` -> throw. What do you think?

packages/core/src/servient.ts
```diff
@@ -160,6 +160,21 @@ export default class Servient {
         }
     }
 
+    public destroyThing(thingId: string): Promise<boolean> {
+        return new Promise<boolean>((resolve, reject) => {
+            if (this.things.has(thingId)) {
+                console.debug("[core/servient]", `Servient destroying thing with id '${thingId}'`);
+                this.things.delete(thingId);
+                let serverPromises: Promise<boolean>[] = [];
+                this.servers.forEach((server) => { serverPromises.push(server.destroy(thingId)); });
```
relu91, line 169: What I found a little "concerning" is that we ourselves are not using the returned boolean value... it seems like we introduced it but we don't need it.

packages/core/src/exposed-thing.ts
```diff
@@ -107,8 +107,28 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
 
     /** @inheritDoc */
     destroy(): Promise<void> {
+        console.debug("[core/exposed-thing]", `ExposedThing '${this.title}' destroying the thing and its interactions`);
+
         return new Promise<void>((resolve, reject) => {
-            resolve();
+            this.getServient().destroyThing(this.id).then(() => {
```
relu91, line 113: Same here.

relu91 (approved):
Sorry. I've just seen the new changes. We can go on, but I am still a little bit concerned about the return value of the `ProtocolServer.destroy` method.

packages/core/src/servient.ts
danielpeintner: I see what you are heading at but I am not sure whether it might turn out useful in the future.

I still think it does not hurt and we can go on but if you feel uncomfortable I am also OK with changing...

relu91: Let's go on, for now, the core <-> protocol binding interface it's still evolving.

PR #440, opened 2021-07-09
Add an example for modbus encoding

relu91:
A minor PR with an example of how bytes are encoded using different byte orders.

### DIFF
--- a/packages/binding-modbus/README.md
+++ b/packages/binding-modbus/README.md
@@ -93,7 +93,7 @@ Along with content type `application/octet-stream`, this protocol binding accept
 
 **Note**: the list may be extended in the future. 
 
-In particular, the decimal number `9545 22880` will be encoded in as follows:
+In particular, the decimal numbers `9545` and `22880` will be encoded as follows:
 - `BIG_ENDIAN`: `25 49 59 60`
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):
LGTM

test failures are unrelated

packages/binding-modbus/README.md
```diff
@@ -93,6 +93,12 @@ Along with content type `application/octet-stream`, this protocol binding accept
 
 **Note**: the list may be extended in the future. 
 
+In particular, the decimal number `9545 22880` will be encoded in as follows:
```
danielpeintner, line 96: - decimal format seems a bit strange...  4 digits followed by 5 digits ?
- I am not native speaker but isn't it "will be encoded in as follows" --> "will be encoded as follows"

danielpeintner (commented):
is the example correct?

decimal number 954522880 in octal 7071156400 ?
As mentioned I might not have enough insights how the mapping is done...

packages/binding-modbus/README.md
relu91: yes sorry, my bad they are actually *two* decimal numbers. Fixing.

danielpeintner (approved):

Issue #439, opened 2021-07-09
Modbus silent crash


LennyHEVS:
I'm writing a simple node application that polls every 5 seconds the value of a property on a modbus electrical counter. I'm accessing this modbus counter via a SSH tunnel with modbus over tcp. 

```Javascript
//Execute this test program with command :
//	node counter-test.js
Servient = require("@node-wot/core").Servient
ModbusClientFactory = require("@node-wot/binding-modbus").ModbusClientFactory

// create Servient
let servient = new Servient();
servient.addClientFactory(new ModbusClientFactory(null));

servient.start().then((WoT) => {
	let td = 'theTDString'; //Not written to reduce code length
	
	WoT.consume(JSON.parse(td)).then((thing) => {
	
		console.log("Consumed " + thing.getThingDescription().title);
		
		setInterval(() => {
		
			thing.readProperty("vl1n").then((volt) => {
				console.log("Voltage L1-N:",volt);
			}).catch((e) => { console.error(e) });
			
		},5000);
	});
});
```

The application works correctly until it doesn't. When this happens, the modbus connection closes but the `Promise` return by the `readProperty` is never resolved nor rejected. The problem seems to be only local to the `ModbusClient`. If I create a second instance of the application while the first one is showing this unwanted behavior, the second application works fine but eventually end up showing the same behavior as the first one.
Here is what's logged :

<details>
  <summary>Logs</summary>

```
[core/helpers] Helpers found scheme 'modbus+tcp'
[core/consumed-thing] ConsumedThing 'Electric counter 1' chose cached client for 'modbus+tcp'
[core/consumed-thing] ConsumedThing 'Electric counter 1' reading modbus+tcp://127.0.0.1:8181
[binding-modbus] Reusing ModbusConnection for  127.0.0.1:8181
[binding-modbus] ModbusTransaction:trigger
[binding-modbus] ModbusConnection:trigger
[binding-modbus] Trying to connect to 127.0.0.1
[binding-modbus] Modbus connected to 127.0.0.1
[binding-modbus] ModbusConnection:trigger
[binding-modbus] Trigger read operation on 3027 len 2
[binding-modbus] Invoking read transaction
[binding-modbus] Got result from read operation on 3027 len 2
[binding-modbus] Operation done
[binding-modbus] ModbusConnection:trigger
[core/content-senders] ContentSerdes deserializing from application/octet-stream
Voltage L1-N: 236.20999145507812
[binding-modbus] Closing unused connection
[...]

PR #442, opened 2021-07-15
Fix 439

relu91:
A minor fix for issue #439. Now the Modbus client properly initializes the connection with the right timeout. Plus this PR also adds a convenient launch task to debug the Modbus test suite.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -85,6 +85,27 @@
                 "--timeout",
                 "999999",
                 "--colors",
+                "${file}"
+            ],
+            "internalConsoleOptions": "openOnSessionStart",
+            "cwd": "${workspaceFolder}/packages/binding-modbus/",
+            "name": "Modbus Test current File",
+            "program": "${workspaceFolder}/packages/binding-http/node_modules/mocha/bin/_mocha",
+            "request": "launch",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "type": "node"
+        },
+        {
+            "args": [
+                "-u",
+                "tdd",
+                "--compilers",
+                "ts:ts-node/register",
+                "--timeout",
+                "999999",
+                "--colors",
                 "${workspaceFolder}/packages/binding-coap/test"
             ],
             "internalConsoleOptions": "openOnSessionStart",
--- a/packages/binding-modbus/src/modbus-client.ts
+++ b/packages/binding-modbus/src/modbus-client.ts
@@ -97,7 +97,7 @@ export default class ModbusClient implements ProtocolClient {
 
     if (!connection) {
       console.debug('[binding-modbus]', 'Creating new ModbusConnection for ', hostAndPort);
-      this._connections.set(hostAndPort, new ModbusConnection(host, port));
+      this._connections.set(hostAndPort, new ModbusConnection(host, port, {connectionTimeout: form['modbus:timeout'] || DEFAULT_TIMEOUT}));
       connection = this._connections.get(hostAndPort);
     }else {
       console.debug('[binding-modbus]', 'Reusing ModbusConnection for ', hostAndPort);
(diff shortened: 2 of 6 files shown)


danielpeintner:
Note: test fail seems unrelated, is about MQTT

danielpeintner (approved):
LGTM, but I am not very familiar with Modbus.

@LennyHEVS can you check also?

LennyHEVS:
The changes are good and seem to be working nicely.

But it exposed another problem. When closing the unused connection fails (with a TCP timeout for example), the flag `connected` of `ModbusConnection` cannot be trusted anymore. For futur requests, I suggest using the `this.client.isOpen` instead of `this.connected` to check the connection.

relu91:
> But it exposed another problem. When closing the unused connection fails (with a TCP timeout for example), the flag connected of ModbusConnection cannot be trusted anymore. For futur requests, I suggest using the this.client.isOpen instead of this.connected to check the connection.

That's a good point, my assumption was that if the `close` method fails then the connection is actually not fully closed, therefore connected -> `true`. Do you think we should revisit this implementation? maybe it might worth assigning the underlying `this.client.isOpen` to `this.connected`. What do you think?

LennyHEVS:
What I saw in the library is that if a callback method is called, the callback function pointer is then set to `null` in the `modbus-serial` library. I noticed that a TCP Timeout could occur between the call of the `close` method and the time when port actually closes.

> That's a good point, my assumption was that if the `close` method fails then the connection is actually not fully closed, therefore connected -> `true`.

In this case, closing does not fail but we will not be able to catch the "on close" signal afterwards. In the timeout callback, the port is still open, which makes assigning `this.client.isOpen` to `this.connected` obsolete after the "on close" signal.

> Do you think we should revisit this implementation? maybe it might worth assigning the underlying `this.client.isOpen` to `this.connected`. What do you think?

Completly revisiting the implementation might be too much of a task. What I think would be great is if we could ensure that the state of the connection at all time. Therefore we will be able to connect/reconnect/close the connection when needed.

relu91:
Sorry @LennyHEVS are you saying that this: 
```ts
 private modbusstop() {
    console.debug('[binding-modbus]', 'Closing unused connection');
    this.client.close((err: string) => {
      if (!err) {
        console.debug('[binding-modbus]', 'session closed');
        this.connecting = false;
        this.connected = this.client.isOpen;
      } else {
        console.error('[binding-modbus]', 'cannot close session ' + err);
      }
    });
    clearInterval(this.timer)
    this.timer = null;
  }
```

Would not work? Sorry I might get a little lost cause I don't remember too much about the internals of Modbus TCP library.

LennyHEVS:
> Would not work? Sorry I might get a little lost cause I don't remember too much about the internals of Modbus TCP library.

Yes, unfortunatly `this.client.isOpen` will return `true` if the callback arrives before the `close` signal.

LennyHEVS:
An easy fix would be to get rid of the flag `connected` and to replace the lines
```Javascript
96    if (!this.connecting && !this.connected) {
with
96    if (!this.connecting && !this.client.isOpen) {
```
,
```Javascript
129    if (!this.connecting && !this.connected) {
with
129    if (!this.connecting && !this.client.isOpen) {
```
and
```Javascript
144    } else if (this.connected && this.currentTransaction == null && this.queue.length > 0) {
with
144    } else if (this.client.isOpen && this.currentTransaction == null && this.queue.length > 0) {
```
But I don't know if you want to keep this flag and what happens then with the `connecting` flag

relu91:
@LennyHEVS Thank you for expanding! I think it makes more sense to not trying to introduce another variable to keep in sync with isOpen. Plus all the tests are green so I think we'll not miss `this.connected`

relu91:
Given the thumbs up by @LennyHEVS, I'd wait for the final approval by @danielpeintner and then merge 🥳

PR #443, opened 2021-07-16
Upgrade tasks file to v2.0.0

relu91:
VS code deprecated v1.0.0 version of the task file, considering that vscode is usually automatically updated I think we are safe to upgrade the task file to v2.0.0. Plus it doesn't really like the file and every time a user oper the repository it automatically replaces it with the new version.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,19 +1,30 @@
 {
     // See https://go.microsoft.com/fwlink/?LinkId=733558
     // for the documentation about the tasks.json format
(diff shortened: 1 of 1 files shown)


danielpeintner:
Note: test fail again unrelated (MQTT test issue)

danielpeintner (approved):

relu91:
> Note: test fail again unrelated (MQTT test issue)

We'll try to look into this MQTT issue 👍🏻

egekorkan (approved):

PR #446, opened 2021-07-19
Fix 437

### LINKED ISSUE #437: Browser HTTPS invokeAction body is missing
So I'm trying to write a simple Javascript web application using the browser bundle and when I try to `POST` something with a body, the thing doesn't receive any body. The request has done by the consumed thing has a `content-lengh` of zero.
```Javascript
// create Servient, add needed credentials and setup HTTPS configuration
let servient = new Wot.Core.Servient();
servient.addCredentials(
	{
		"urn:dev:org:wot-thing": {
			clientId: "theID",
			clientSecret: "theSecret"
		}
	}
);
let httpsConfig = {
    allowSelfSigned: true, // client configuration
};
servient.addClientFactory(new Wot.Http.HttpsClientFactory(httpsConfig));
let wotHelper = new Wot.Core.Helpers(servient);


$(document).ready(function() {
	wotHelper.fetch("https://localhost:443").then(async (td) => {
		servient.start().then((factory) => {
			factory.consume(td).then((thing) => {
				
				let stringBody= '{"string": "someString"}';
				
				thing.invokeAction("postString",JSON.parse(stringBody)).then((res) => {
					console.log("Result:",res);
				});
			});
		});
	});
});
```
The Thing that I'm using uses HTTPS with the flow `client-crendentials` of the `oauth2` security scheme
If I re-write this code in a Node application it works fine. But with the web app the `stringBody` is not there.

This is what I'm receiving
```
Request!
        Method:  POST
        URL:  /postString
        Variables:  URLSearchParams {}
Data:  undefined
Params:  { uriVariables: {} }
```
I tried to follow the request in the browser-bundle library and in the Typescript library but I don't find the source of the problem.
The `request` returned by `generateFetchRequest` in the `http-client` file seems to have a key "bodyUsed" set to false at some point.

Thanks for the help already !

relu91:
This PR fix #437 plus a minor addition for testing OAuth 2.0 with browsers. Basically, I found that in the previous implementation we cloned the request many times. This process worked well on nodes but on browsers, it fails to copy the body in the last `const useNewURL = new Request(tempRequest.url,mergeHeaders)` assigned. Sadly, I didn't have time to dig more about the reasons why this is happing (maybe browsers don't like to copy too many times a ReadableStream?) but  `const useNewURL = new Request(tempRequest.url,mergeHeaders)` was just a precaution to ensure that the URL was not rewritten by `token.sign`. Indeed `const mergeHeaders = new Request(request,tempRequest)` should be already sufficient

### DIFF
--- a/examples/security/oauth/package.json
+++ b/examples/security/oauth/package.json
@@ -12,6 +12,7 @@
   "license": "EPL-2.0 OR W3C-20150513",
   "devDependencies": {
     "@types/express-oauth-server": "^2.0.2",
+    "cors": "^2.8.5",
     "express-oauth-server": "^2.0.0"
   }
 }
--- a/examples/security/oauth/server.js
+++ b/examples/security/oauth/server.js
@@ -17,12 +17,18 @@
  */
 const OAuthServer = require('express-oauth-server')
 const bodyParser = require('body-parser');
+const cors = require('cors');
 const https = require('https');
 const fs = require('fs')
 const express = require('express')
 const Memory = require('./memory-model');
 
+
+
+
 var app = express();
+app.use(cors());
+app.options('*', cors());
 
 const model = new Memory()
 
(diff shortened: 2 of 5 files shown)


danielpeintner:
@LennyHEVS  do you want to take a look also?

danielpeintner (approved):

LennyHEVS:
This has successfully fixed the issue !
Thanks for all the work !

Issue #448, opened 2021-07-28
octect codec do not support half precision float


relu91:
```
[core/content-serdes] ContentSerdes serializing to application/octet-stream;byteSeq=BIG_ENDIAN;length=2
(node:27750) UnhandledPromiseRejectionWarning: Error: Wrong buffer length for type 'number', must be 4 or 8, is 2
```

Solution: decode and encode also `float16`

danielpeintner:
resolved by https://github.com/eclipse/thingweb.node-wot/pull/449

Issue #447, opened 2021-07-28
octetstream-codec use @type instead of type


sebastiankb:
@fatadel and I reviewed this [if statement](https://github.com/eclipse/thingweb.node-wot/blob/1de334c65b3cfc4248b69fd65593eb14d5ca059a/packages/core/src/codecs/octetstream-codec.ts#L169):

```js
        if(schema['@type'] !== undefined) {
            let semTypes: Array<string> = schema["@type"];
            // check for endianness semantic type
            // see http://www.meta-share.org/ontologies/meta-share/meta-share-ontology.owl/documentation/index-en.html
            let endianSem = semTypes.find(v => v.endsWith(":bigEndian") || v.endsWith(':littleEndian'));
            bigendian = endianSem === undefined ? bigendian : endianSem.endsWith(":bigEndian");
            // check for numeric datatype semantic
            // see https://www.w3.org/TR/xmlschema-2/
            let typeSem = semTypes.find(v => /:(unsigned)?(short|int|long|float|double|byte)/.test(v.toLowerCase()));
            if(typeSem) {
                // check for sign semantic type
                signed = typeSem.toLowerCase().indexOf('unsigned') === -1;
                let numberSem = /(short|int|long|float|double|byte)/.exec(typeSem.toLowerCase())[1];
                dataType = numberSem === "int" || numberSem === "short" || numberSem === "byte" ? "integer" : "number";
            }
        }
```

1. `let semTypes: Array<string> = schema["@type"];` --> expect an array, however, it can be both string and array
2. `var endianSem = semTypes.find(function (v) { return v.endsWith(":bigEndian") || v.endsWith(':littleEndian'); });` --> uses also bigEndian and littleEndian. This may be in conflict with the modbus-based [definition](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus#dataschema)

danielpeintner:
resolved by https://github.com/eclipse/thingweb.node-wot/pull/449

PR #449, opened 2021-07-28
Octet stream encoding process and support for float16

relu91:
This PR fixes issues #448 and #447.  Notice that `@type` is not used as the source of information for the encoding format. The library now uses `type` and a set of well know identifiers for complex types like `uint8`.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -24,13 +24,14 @@
     "typescript-standard": "0.3.36"
   },
   "dependencies": {
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.7",
-    "uuid": "3.4.0",
+    "@petamoriken/float16": "^3.1.1",
+    "@types/uritemplate": "0.3.4",
     "rxjs": "5.5.11",
     "uritemplate": "0.3.4",
-    "@types/uritemplate": "0.3.4",
-    "vm2": "^3.9.2"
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -24,13 +24,14 @@
     "typescript-standard": "0.3.36"
   },
   "dependencies": {
-    "wot-typescript-definitions": "0.7.3",
     "@node-wot/td-tools": "0.7.7",
-    "uuid": "3.4.0",
+    "@petamoriken/float16": "^3.1.1",
+    "@types/uritemplate": "0.3.4",
```
relu91, line 29: I don't know why but every time I install dependency things get shuffled. It seems my npm version doesn't like the current ordering... In this PR I added `@petamoriken/float16` for `float16` coding and deconding

danielpeintner: NPM seems to order them alphabetically. Do you know?

Maybe we should go once over all dependencies and sort them alphabetically so that we do not get these unnecessary diffs all the time. What do you think?

danielpeintner (approved):
LGTM

packages/core/package.json
danielpeintner: let's discuss that in https://github.com/eclipse/thingweb.node-wot/issues/450

sebastiankb:
@relu91 many thanks

PR #451, opened 2021-07-29
refactor: increase timeout to stabilize MQTT test

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/434 ?

### DIFF
--- a/packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
+++ b/packages/binding-mqtt/test/mqtt-client-subscribe-test.ts
@@ -30,7 +30,7 @@ import MqttClientFactory from "../dist/mqtt-client-factory";
 @suite("MQTT implementation")
 class MqttClientSubscribeTest {
 
-    @test(timeout(5000)) "should expose via broker"(done: Function) {
+    @test(timeout(10000)) "should expose via broker"(done: Function) {
 
         try {
             let servient = new Servient();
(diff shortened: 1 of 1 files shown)


danielpeintner:
let's try whether increasing timeout helps to stabilize the MQTT test on GitHub.
Locally I do not see fails :-(

relu91:
btw I forget to send this PR: https://github.com/relu91/thingweb.node-wot/pull/11 it should solve the issue, it is kinda related to timeouts. Should we use that instead?

danielpeintner:
> btw I forget to send this PR: [relu91#11](https://github.com/relu91/thingweb.node-wot/pull/11) it should solve the issue, it is kinda related to timeouts. Should we use that instead?

Yes, if it fixes the issue, great!
I noticed that https://github.com/relu91/thingweb.node-wot/pull/11/checks?check_run_id=3189248515 fails for CoAP but this might be another issue.

Otherwise I would start to ignore the MQTT test ...

danielpeintner:
For the time being we can also simply *skip* the MQTT test .. that's what it is done in this PR

danielpeintner:
@relu91 @egekorkan since MQTT test are unstable I suggest to *skip* them till we have a solution.
Otherwise we keep **red** flags for all other PRs.

Do you agree?

relu91:
I looked in https://github.com/relu91/thingweb.node-wot/pull/11 it seems that the `coap` problem was just another fluctuation. At least it stabilizes mqtt and opc-ua tests. I'll bring it here and I would favor that one instead of skipping.

danielpeintner:
withdrawn in favor of https://github.com/eclipse/thingweb.node-wot/pull/459

PR #452, opened 2021-07-29
Validate ExposeThingInit using PartialTD validation algorithm

relu91:
This PR covers the new specification about the [produce function](https://w3c.github.io/wot-scripting-api/#the-produce-method). In particular, the new algorithm specifies how to [validate](https://w3c.github.io/wot-scripting-api/#validating-an-exposedthinginit) the input of the function.

We employed ajv since is already a dependency of this version of node-wot. Sadly, this change required upgrading all TS versions because ajv. And it implies that node-wot dependants now are required to use a TS version greater than 4.1.3. I still don't know the real reasons behind this, cause we already used ajv in the value function without needing to upgrade all the packages. Let me know if you think it is worth digging more on this.   

Finally,  The PR contains also small code improvements and a convenient npm script to just test the whole codebase.


egekorkan (approved):

danielpeintner (approved):

danielpeintner:
thanks @relu91  and @fillobotto

PR #459, opened 2021-08-04
Wait for connection in mqtt

relu91:
Reporting @fillobotto PR into the main repo. His comment:
>MQTT fixed test: c8f2d13
Doubled test timeout to 10000ms and consequently increased timeout between emitted events from the exposed thing.
Tested on 11 CI runs (12 tests each).
>OPCUA fixed test: 20e283b
With the newer OPCUA client binding a breaking change was introduced: endpoint_must_exist is now endpointMustExist

I've tested locally and remotely aside from one failure in coap binding I've never seen a failure in mqtt binding. I would say this is an improvement. One odd thing is that 20e283b changes opcua-client.ts but is not reported among the changed files. Maybe it was already updated meanwhile?


danielpeintner (approved):
LGTM

PR #460, opened 2021-08-06
Added M-Bus binding

LennyHEVS:
Because I need to use a Thing using M-Bus over TCP in a project, I added here a binding for the M-Bus protocol. It only supports property reads with no security schemes. It is using node-mbus to communicate.
I provided an example TD and script but couldn't test them because it is using the CLI which doesn't support M-Bus clients yet.
I also did not include tests for this bindings.
My project is using this piece of code without any noticable problems.

### DIFF
--- a/packages/binding-mbus/package.json
+++ b/packages/binding-mbus/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@node-wot/binding-mbus",
+  "version": "0.7.7",
+  "description": "M-Bus TCP client protocol binding for node-wot",
+  "author": "Eclipse Thingweb <thingweb-dev@eclipse.org> (https://thingweb.io/)",
+  "license": "EPL-2.0 OR W3C-20150513",
+  "repository": "https://github.com/thingweb/node-wot/tree/master/packages/binding-mbus",
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist/"
+  ],
+  "main": "dist/mbus.js",
+  "types": "dist/mbus.d.ts",
+  "devDependencies": {
+    "@types/chai": "^4.2.7",
+    "@types/chai-as-promised": "^7.1.2",
+    "@types/mocha": "^5.2.7",
+    "@types/node": "^10.9.4",
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
+    "chai-spies": "^1.0.0",
+    "mocha": "^7.0.0",
+    "ts-node": "^8.6.2",
+    "typescript": "^3.9.10",
+    "typescript-standard": "^0.3.36"
+  },
+  "dependencies": {
+    "@node-wot/core": "0.7.7",
+    "@node-wot/td-tools": "0.7.7",
+    "node-mbus": "^1.2.2",
+    "wot-typescript-definitions": "0.7.3"
+  },
+  "scripts": {
+    "build": "tsc",
+    "test": "mocha --require ts-node/register test/*"
+  }
+}
--- a/packages/binding-mbus/src/mbus-client-factory.ts
+++ b/packages/binding-mbus/src/mbus-client-factory.ts
@@ -1,0 +1,13 @@
+import { ProtocolClientFactory, ProtocolClient } from '@node-wot/core';
+import MBusClient from './mbus-client';
+
+export default class MBusClientFactory implements ProtocolClientFactory {
+  public readonly scheme: string = 'mbus+tcp';
+
+  public getClient(): ProtocolClient {
+    console.log(`MBusClientFactory creating client for '${this.scheme}'`);
+    return new MBusClient();
+  }
+  public init = () => true;
+  public destroy = () => true;
+}
--- a/packages/binding-mbus/tsconfighead.json
+++ b/packages/binding-mbus/tsconfighead.json
@@ -1,0 +1,19 @@
+{
+	"compilerOptions": {
+		"target": "es5",
+		"lib" : ["es2015","dom"],
+		"module": "commonjs",
+		"outDir": "dist",
+		"alwaysStrict": true,
+		"sourceMap": true,
+		"noImplicitAny": true,
+		"removeComments": true,
+		"noLib": false,
+		"preserveConstEnums": true,
+		"experimentalDecorators": true,
+		"declaration": true
+	},
+	"include": [
+		"src/**/*"
+	]
+}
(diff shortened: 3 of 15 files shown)


packages/binding-mbus/src/mbus-client-factory.ts
```diff
@@ -0,0 +1,13 @@
+import { ProtocolClientFactory, ProtocolClient } from '@node-wot/core';
+import MBusClient from './mbus-client';
+
+export default class MBusClientFactory implements ProtocolClientFactory {
+  public readonly scheme: string = 'mbus+tcp';
+
+  public getClient(): ProtocolClient {
+    console.log(`MBusClientFactory creating client for '${this.scheme}'`);
```
relu91, line 8: We have a [log convention](https://github.com/eclipse/thingweb.node-wot#logging) here. Can you please change this to:
```js
console.info("[binding-mbus]",`MBusClientFactory creating client for '${this.scheme}'`
```

packages/cli/src/cli-default-servient.ts
```diff
@@ -28,6 +28,7 @@ import { HttpsClientFactory } from "@node-wot/binding-http";
 import { CoapClientFactory } from "@node-wot/binding-coap";
 import { CoapsClientFactory } from "@node-wot/binding-coap";
 import { MqttClientFactory }  from "@node-wot/binding-mqtt";
+import { MBusClientFactory }  from "@node-wot/binding-mbus";
```
relu91, line 31: Sorry, but I would not include this binding into the default servient. We don't have defined a clear policy yet but as you can see a lot of other bindings are actually missing. Taking into account the fact that this binding is not fully implemented I would prefer to leave it out.

packages/cli/src/cli-default-servient.ts
```diff
@@ -113,6 +114,7 @@ export default class DefaultServient extends Servient {
         this.addClientFactory(new CoapClientFactory(coapServer));
         this.addClientFactory(new CoapsClientFactory());
         this.addClientFactory(new MqttClientFactory());
+		this.addClientFactory(new MBusClientFactory());
```
relu91, line 117: the above applies also here.

packages/binding-mbus/tsconfighead.json
```diff
@@ -0,0 +1,19 @@
+{
```
relu91, line 1: May I ask you to follow the same configuration of the other packages? Here's an example:
```json
{
	"compilerOptions": {
		"outDir": "dist"
	},
	"include": [
		"src/**/*"
	]
}
```
It is a little bit inconvenient handling different tsc configurations per package

relu91 (changes requested):
First of all, I want to say thank you for having an interest in the project and providing a good contribution. Node-wot is not as easy as it should be to tackle 😄 . I left some minor remarks below. 

Besides that I want to add some general comments:
- Could you add a warning in the readme about the fact that this is an experimental binding? Since it is only tested in your use case I am afraid it might not work for others. Having a disclaimer just warns the users to be careful. 
- About the example you might want to check the Modbus example, as you can see in that case we added a custom servient inside the `servient` folder. I think you should re-use the same pattern in yours. 
- It would be cool if you can provide your example in TypeScript under `packages/examples` 

Once again well done!

relu91:
Oh almost forgot, could you also check our [contribution guidelines](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md) and sign the Eclipse Contributor Agreement

LennyHEVS:
Okay thank you for the feedback !
I will post the changes during the following days

LennyHEVS:
About the contribution Agreement, when creating my account on Eclipse, an error occured. I'm currently contacting the support to resolve the issue so I'll let you know as soon as possible when it is updated and the agreement signed.

relu91 (approved):
Thanks! About the license issue, @danielpeintner should we wait until is resolved?

danielpeintner:
> Thanks! About the license issue, @danielpeintner should we wait until is resolved?

I hope it will not take that long to resolve the ECA issues.
@LennyHEVS  Once the issue is settled please let us know and we restart the checks to get a green light

packages/binding-mbus/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@node-wot/binding-mbus",
+  "version": "0.7.7",
```
danielpeintner, line 3: Can you update to
`"version": "0.7.8"`

danielpeintner (commented):
Do you think you can setup some simple tests?

packages/binding-mbus/package.json
danielpeintner: Note: I think this causes the tests to fail, see https://github.com/eclipse/thingweb.node-wot/pull/460/checks?check_run_id=3281511482
`npm ERR! 404  '@node-wot/binding-mbus@0.7.8' is not in the npm registry.`

danielpeintner:
The setup is still *partly* wrong (tests expected), see https://github.com/eclipse/thingweb.node-wot/pull/460/checks?check_run_id=3282122515

```
@node-wot/binding-mbus@0.7.8 test /home/runner/work/thingweb.node-wot/thingweb.node-wot/packages/binding-mbus
mocha --require ts-node/register test/*
lerna ERR! npm run test stderr:
Error: No test files found: "test/*"
```

There are no tests. As mentioned before simple tests are very beneficial.

LennyHEVS:
Yes I will try to add some tests to the library. I'll take a look at the other binding tests to make something similar

LennyHEVS:
Knowing I cannot create M-Bus servers with this binding and the underlying library, I just tested connection timeouts and the created client functions.

LennyHEVS:
Yes sorry I missed that I can't properly make sure that the tests are correct but I hope it's all fixed now

LennyHEVS:
Because I'm using a library which isn't in Typescript. I don't know how to declare my library to ts-node. I tried adding the file module-declaration.d.ts but it doens't seem to resolve it.

danielpeintner:
> Because I'm using a library which isn't in Typescript. I don't know how to declare my library to ts-node. I tried adding the file module-declaration.d.ts but it doens't seem to resolve it.

The are still issues, see https://github.com/eclipse/thingweb.node-wot/pull/460/checks?check_run_id=3300303790

```
TSError: ⨯ Unable to compile TypeScript:
Error: src/mbus-connection.ts(1,29): error TS7016: Could not find a declaration file for module 'node-mbus'. '/Users/runner/work/thingweb.node-wot/thingweb.node-wot/packages/binding-mbus/node_modules/node-mbus/index.js' implicitly has an 'any' type.
  Try `npm install @types/node-mbus` if it exists or add a new declaration (.d.ts) file containing `declare module 'node-mbus';`
```

You should be able to start `npm run test` in the `binding-mbus` folder and see the same errors locally.

danielpeintner:
LGTM.

The test failures are coming from MQTTS, this is something we should fix :-(
Mbus seems fine.

@relu91 @egekorkan any more comments?

PR #465, opened 2021-08-10
docs: fix copyright years based on "git log"

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/464

### DIFF
--- a/examples/servients/coap-servient/src/coap-servient.ts
+++ b/examples/servients/coap-servient/src/coap-servient.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2018 - 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018 - 2020 Contributors to the Eclipse Foundation
  * 
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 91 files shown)


relu91 (approved):

danielpeintner:
Note: I added copyright header to files that did not have one.

I also added the mini script that does the work in future..

relu91 (approved):
Good! Even if it would have been better if the script was js or ts so that it can be invoked by the current tool chain.

danielpeintner:
> I would say that the script should be in js (or ts) in order to let other contributors use it without needing to install Java dependencies

Anyone willing to write it in JS/TS can do so. I would welcome any contribution and we can replace the current Java file.

The fix-copyright process is something most likely one of us runs once in a while to sync up... not really party of the tool-chain. This process is somewhat nasty and I assume there is a reason why everyone is doing it somewhat differently..

Moreover, not sure how easy it is to parse the `git log` result since it need to grap it from the console (see https://github.com/eclipse/thingweb.node-wot/blob/823647b999e1a6c3ca2c4d5706d69836c1a329b1/utils/fix-copyright/CopyrightFix.java#L65-L88). As mentioned I felt more comfortable to write a single Java class doing that...

PR #467, opened 2021-08-13
fix: take into account security at form level

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/466

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -121,19 +121,30 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
         return form;
     }
 
-    ensureClientSecurity(client: ProtocolClient) {
-        // td-tools parser ensures this.security is an array
-        if (this.security && this.securityDefinitions && Array.isArray(this.security) && this.security.length > 0) {
-            console.debug("[core/consumed-thing]",`ConsumedThing '${this.title}' setting credentials for ${client}`);
-            let scs: Array<TD.SecurityScheme> = [];
-            for (let s of this.security) {
-                let ws = this.securityDefinitions[s + ""]; // String vs. string (fix wot-typescript-definitions?)
-                // also push nosec in case of proxy
-                if (ws) {
-                    scs.push(ws);
-                }
+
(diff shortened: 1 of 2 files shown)


danielpeintner:
@relu91 @egekorkan a proper test is missing.

Having said that, we do not seem to have a good set of security tests where I could add one. Did I oversee something (and you can point me to it).. since I am somewhat in a hurry 🙈

packages/core/src/consumed-thing.ts
```diff
@@ -121,19 +121,30 @@ export default class ConsumedThing extends TD.Thing implements WoT.ConsumedThing
...
-                let ws = this.securityDefinitions[s + ""]; // String vs. string (fix wot-typescript-definitions?)
-                // also push nosec in case of proxy
-                if (ws) {
-                    scs.push(ws);
-                }
+
+    getSecuritySchemes(security: Array<String>): Array<TD.SecurityScheme> {
+        let scs: Array<TD.SecurityScheme> = [];
+        for (let s of security) {
+            let ws = this.securityDefinitions[s + ""]; // String vs. string (fix wot-typescript-definitions?)
```
relu91, line 128: Is this comment still valid? 🤔

relu91 (approved):
Changes look good. I agree that we would need a proper unit test for this feature. I don't recall to have found any. In `binding-http` there are just some tests about credentials and OAuth2.0 flows.

packages/core/src/consumed-thing.ts
danielpeintner: What is strange is the following:
security on thing level is of type: `security: Array<String>`
security on form level is of type: `security?: Array<string>`
`String` vs. `string`

danielpeintner:
> I agree that we would need a proper unit test for this feature.

I added 2 very simple tests.

packages/core/src/consumed-thing.ts
relu91: I checked the new TD types and I don't see this problem. We should remember to remove the comment once we merge with 0.8.x

relu91 (approved):
Good to go!

PR #472, opened 2021-08-25
fix: print correct error message

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/470


maxim-s-barabash:
Excellent. :)

Issue #475, opened 2021-08-26
MBus tests failing


danielpeintner:
see  https://github.com/eclipse/thingweb.node-wot/pull/473/checks?check_run_id=3433642871

2 failing tests

```

  1) MBus connection
       should throw for unknown host:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/thingweb.node-wot/thingweb.node-wot/packages/binding-mbus/test/mbus-connection-test.ts)
  

  2) MBus connection
       Operation
         should fail for unknown host:
     Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/thingweb.node-wot/thingweb.node-wot/packages/binding-mbus/test/mbus-connection-test.ts)
```

My first attempt was to increase timeout to 10 secs since I saw locally 2 seconds.
Note: Locally the tests run stable and fine on my machine.

Anyhow, not sure what is causing these issues, i skipped the failings tests, see https://github.com/eclipse/thingweb.node-wot/pull/474

@LennyHEVS can you take a look?

LennyHEVS:
I modified the connection test file in the PR #504 . I'm not sure if this is going to work or not.
I'd like to see the result of the tests to be sure of the fix.

PR #483, opened 2021-08-31
Npm workspaces

### LINKED ISSUE #346: Use NPM workspaces to manage multiple packages from within a singular top-level root package
https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/

Worth investigating ...

relu91:
This PR refactors the current project structure using npm workspaces. As a result of this refactoring `lerna` is no longer among our dev dependencies and everything has been ported to the new workspaces APIs. 

Open topics:
- Didn't test the publication procedure
- Didn't test CI (we'll see it after the creation of this PR)

Other points:
- After the refactoring npm added additional info to package.json of the bindings. I left that changes cause the were reasonable
- Opc-ua tests were not working, and I don't understand how the ever worked cause `url.origin` returns `'null'` with the current setup, and even testing online it returns the same ([see on runkit](https://runkit.com/embed/7grj9q5lpmxt)). I think the current code is more stable.

Closes #346


danielpeintner:
@relu91 I think the [issues](https://github.com/eclipse/thingweb.node-wot/pull/483/checks?check_run_id=3472929027) appear for nodejs version 10
see https://github.com/eclipse/thingweb.node-wot/blob/master/.github/workflows/ci.yaml#L13

Shall we remove `10.x` and add `16.x` ?

danielpeintner:
Note: I limited GitHub actions to LTS and current, see https://github.com/eclipse/thingweb.node-wot/commit/f398f79a3f567f2eebd7f260b25735a1cd896543

egekorkan:
Sadly not a super useful comment but when we were restructuring TD playground, we had considered both options. I don't exactly remember why but lerna was better _at the time_. There is this SO thread from 9 months ago: https://stackoverflow.com/questions/64909635/is-lerna-needed-anymore-with-npm-7-0-0s-workspaces

I have also read that lerna is having difficulties maintaining itself but was there a reason for the change?

relu91:
> @relu91 I think the issues appear for nodejs version 10
see https://github.com/eclipse/thingweb.node-wot/blob/master/.github/workflows/ci.yaml#L13

It might be related but I am not sure is completely the problem. Before lunch I was digging into it, I've just finished a call and I'll try to understand what's going on. 

> Shall we remove 10.x and add 16.x ?

Looking around 10.x is considered outdated. I would completely ok to remove the support for it in 0.8.x. Is everybody ok with it? Should we open an issue and call for opinions? 

>Sadly not a super useful comment but when we were restructuring TD playground, we had considered both options. I don't exactly remember why but lerna was better at the time. There is this SO thread from 9 months ago: https://stackoverflow.com/questions/64909635/is-lerna-needed-anymore-with-npm-7-0-0s-workspaces
>
>I have also read that lerna is having difficulties maintaining itself but was there a reason for the change?

It seems that we read the same stackoverflow question and took different decisions. What I read there is that `lerna` is still useful but for complementary tooling and commands. Tools and commands that we don't use. In `node-wot` we mostly used `lerna` as a device for handling our monorepo. Consequently, I think moving to `workspaces` is the right choice here. A couple of thoughts:
- It is better supported (I don't think `npm` is going to die soon, plus is well founded)
- I like the symlink approach (also used by yarn) -> less disk space used
- just one `pakcage-lock.json` 
- It should be well integrated with the publishing workflow

What do you think?

egekorkan:
I see the advantages @relu91. We can just go ahead :)

relu91:
Ok, I found out it's a well-known bug: https://github.com/npm/cli/issues/2632 . There's also a possible [workaround described](https://github.com/npm/cli/issues/2632#issuecomment-888960315), I am not a fan of it... but it is better than nothing 😄 .

danielpeintner:
@relu91 merging in master causes merge conflicts.. sorry for that

relu91:
don't worry not a big deal 😄

danielpeintner:
Do you know why *no* GitHub jobs have been started?

Will do a local check also...

relu91:
> Do you know why _no_ GitHub jobs have been started?
> 
> Will do a local check also...

It is probably because it was a merge commit... 🤔 but I checked on my fork. There was still a problem. Let's see if now it is solved

relu91:
Just to keep you posted, one issue down. The current one is about typescript. It seems that in a clean repository it cannot find the types cause it actually builds the packages following the lexicographic order, not the dependency tree.

relu91:
Ok, now it seems that the CI windows version does not like some native packages. I wasn't able to reproduce inside my machine yet. Let's see.

packages/binding-mbus/package.json
```diff
@@ -35,5 +35,12 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/*"
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
+  },
+  "homepage": "https://github.com/thingweb/node-wot/tree/master/packages/binding-mbus#readme",
```
danielpeintner, line 42: Link is broken,  should be
https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-mbus#readme

packages/binding-modbus/package.json
```diff
@@ -38,5 +38,12 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/*"
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
+  },
+  "homepage": "https://github.com/thingweb/node-wot/tree/master/packages/binding-modbus#readme",
```
danielpeintner, line 45: Link broken also.. eclipse etc missing

packages/binding-netconf/package.json
```diff
@@ -37,5 +37,13 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/netconf-client-test.ts"
-  }
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
+  },
+  "homepage": "https://github.com/thingweb/node-wot/tree/master/packages/binding-netconf#readme",
```
danielpeintner, line 44: Broken link

packages/binding-wotfirestore-browser-bundle/package.json
```diff
@@ -2,7 +2,10 @@
   "name": "@node-wot/binding-wotfirestore-browser-bundle",
   "version": "0.8.0",
   "description": "A binding-wotfirestore bundle that can run in a web browser",
-  "repository": "https://github.com/hidetak/thingweb.node-wot",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hidetak/thingweb.node-wot.git"
+  },
```
danielpeintner, line 8: Why does repository look different here, we seem to use something like this elsewhere

"repository": "https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-wotfirestore",

packages/binding-wotfirestore-browser-bundle/package.json
```diff
@@ -21,5 +24,10 @@
   },
   "scripts": {
     "build": "browserify -r vm:vm2 index.js --external coffee-script -o dist/binding-wotfirestore-bundle.js"
-  }
+  },
+  "bugs": {
+    "url": "https://github.com/hidetak/thingweb.node-wot/issues"
+  },
+  "homepage": "https://github.com/hidetak/thingweb.node-wot#readme",
+  "keywords": []
```
danielpeintner, line 32: Similar question, shouldn't it point to

https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-wotfirestore

packages/binding-wotfirestore/package.json
```diff
@@ -2,7 +2,10 @@
   "name": "@node-wot/binding-wotfirestore",
   "version": "0.8.0",
   "description": "Firestore binding for node-wot",
-  "repository": "https://github.com/hidetak/thingweb.node-wot",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hidetak/thingweb.node-wot.git"
+  },
```
danielpeintner, line 8: Similar comment, not pointing to package

packages/binding-wotfirestore/package.json
```diff
@@ -38,5 +41,13 @@
     "typescript": "4.1.3",
     "typescript-standard": "0.3.36",
     "wot-typescript-definitions": "^0.8.0-SNAPSHOT.18"
-  }
+  },
+  "bugs": {
+    "url": "https://github.com/hidetak/thingweb.node-wot/issues"
+  },
+  "homepage": "https://github.com/hidetak/thingweb.node-wot#readme",
```
danielpeintner, line 48: Similar comment, not pointing to package

danielpeintner (changes requested):
See some broken links issues.
Also, I will run a local test and apart from that it looks good to me. thanks!

danielpeintner:
> I'll try to fix the m-bus issues before the end of next week. I'll let you know when I'm done with this and with #475

Thanks @LennyHEVS 
I created https://github.com/eclipse/thingweb.node-wot/issues/488 to keep track

packages/binding-mbus/package.json
```diff
@@ -35,5 +35,12 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/*"
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
```
danielpeintner, line 40: still wrong link

packages/binding-modbus/package.json
```diff
@@ -38,5 +38,12 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/*"
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
```
danielpeintner, line 43: link is wrong

packages/binding-netconf/package.json
```diff
@@ -37,5 +37,13 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/netconf-client-test.ts"
-  }
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
```
danielpeintner, line 42: wrong also

packages/binding-opcua/package.json
```diff
@@ -39,5 +39,13 @@
   "scripts": {
     "build": "tsc",
     "test": "mocha --require ts-node/register test/opcua-client-test.ts"
-  }
+  },
+  "bugs": {
+    "url": "https://github.com/thingweb/node-wot/issues"
```
danielpeintner, line 44: to be fixed

relu91:
Another update, I tried to re-introduce mbus cause I think the real problem is the fact that the ci does not have the right window-build-tools installed. However, it seems that installing those tools on the ci is causing an infinite loop... I'll revert the commit and address @danielpeintner comments above.

danielpeintner:
I think we can go ahead merging this PR "as is" and solve mbus issues in a separate PR.
@relu91 if you agree please merge 👍

Issue #481, opened 2021-08-30
NPM install shows oudated dependencies warnings


danielpeintner:
running `npm install` for `v.0.8.x` shows

```
npm WARN deprecated ini@1.3.5: Please update to ini >=1.3.6 to avoid a prototype pollution issue
npm WARN deprecated mkdirp-promise@5.0.1: This package is broken and no longer maintained. 'mkdirp' itself supports promises now, please switch to that.
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated har-validator@5.1.3: this library is no longer supported
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated mkdirp@0.5.1: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated request@2.88.0: request has been deprecated, see https://github.com/request/request/issues/3142
```

We should try to resolve these warnings...

egekorkan:
I don't know where request is used now but we already have node-fetch and request is deprecated as the warning is shown

danielpeintner:
> I don't know where request is used now but we already have node-fetch and request is deprecated as the warning is shown

We use it in:
* HTTP: https://github.com/eclipse/thingweb.node-wot/blob/b3e8535394f04e59edca1c7b506d5056f8a940be/packages/binding-http/package.json#L33
* MQTT: https://github.com/eclipse/thingweb.node-wot/blob/b3e8535394f04e59edca1c7b506d5056f8a940be/packages/binding-mqtt/package.json#L23
* Examples: https://github.com/eclipse/thingweb.node-wot/blob/b3e8535394f04e59edca1c7b506d5056f8a940be/examples/templates/exposed-thing/package.json#L21

and it seems other dependencies seem to refer to it as well

I like the following image very much ;-) 

![grafik](https://user-images.githubusercontent.com/11281461/131514269-9b9dd90e-05b5-463a-8f4e-b250b616e205.png)

egekorkan:
Regarding where it is _used_:
- The example uses it for a normal HTTP request, to register a TD. This is a legit usage.
[...]

PR #491, opened 2021-09-02
Npm fix audit

relu91:
**WARNING**: This pr is based on #483

We have accumulated a fair amount of npm auditing errors plus deprecation warnings as explained in #481. This PR tries to solve most of them. Also, it tries to clean up not-used libraries like request in http-binding. 

Current open issues:
- Most of the auditing problems were caused by binding-wotfirestore. I solve them by upgrading to firebase 9. As a consequence now the code is not buildable anymore. I'd ask @hidetak to check what's going on. Meanwhile, the binding is excluded from the package list.
- Even if I upgraded to firebase 9, it seems that firebase-tool is still referencing deprecated versions of `request` which in turn uses a bunch of other deprecated libraries (e.g. uuid 3)
- 3 of the current auditing warnings are caused by [epress-oauth-server](https://www.npmjs.com/package/express-oauth-server). However, there is no new version published yet. The development seems stale to me. This package is needed for testing OAuth 2 implementation. 
- Another warning is given by oracle binding. Specifically, the library iot-csl-js seems to have some old dependencies. It looks like is not open source... probably we can't do anything about it
- Moreover in oracle-binding node-pre-gyp required by sqlite is outdated.


danielpeintner:
> * Most of the auditing problems were caused by binding-wotfirestore. I solve them by upgrading to firebase 9. As a consequence now the code is not buildable anymore. I'd ask @hidetak to check what's going on. Meanwhile, the binding is excluded from the package list.

Makes sense.
Another alternative is to remove it for now since we had test issues also after the new API changes. The code is available and can be brought back later....

> * 3 of the current auditing warnings are caused by [epress-oauth-server](https://www.npmjs.com/package/express-oauth-server). However, there is no new version published yet. The development seems stale to me. This package is needed for testing OAuth 2 implementation.

Mhh, seems we need to live with it.

 
> * Another warning is given by oracle binding. Specifically, the library iot-csl-js seems to have some old dependencies. It looks like is not open source... probably we can't do anything about it
> 
> * Moreover in oracle-binding node-pre-gyp required by sqlite is outdated.

Maybe it is time to remove oracle-binding from v0.8.x on since development has been stale. We (or anyone else) can still bring it back if here is interest.

hidetak:
@relu91 @danielpeintner 
I apologize for the many audit errors that remain.
I know that this Binding doesn't work in Firebase 8 and above (it can't do Browserify).
I will try to find out if it is possible to prevent the audit error in Firebase 7.

relu91:
> @relu91 @danielpeintner
> I apologize for the many audit errors that remain.

Don't worry 👍🏻  

> I will try to find out if it is possible to prevent the audit error in Firebase 7.

From what I understood, it seems that Firebase 7 was at its lastest version and I don't know if you can really solve those audit issues without upgrading... 😢 On possible solution, to not cluttering node-wot might be to have it published separately and just mention/link the implementation inside the folder.

hidetak:
@relu91 
Thank you for your comment.

> On possible solution, to not cluttering node-wot might be to have it published separately and just mention/link the implementation inside the folder.

Does that mean I should do my own publishing to npm in another repository instead of this one?

I would like to find a way to solve this problem first.
If I can't find a solution, I would be happy to consult with you again.

danielpeintner:
I re-run the tests 3 times on GitHub and we are always getting (different) errors.
* nan.h(56,10): fatal error C1083: Cannot open include file: 'node.h': No such file
* OPCUA failures

Not sure whether it is due the changes or other aspects that are GH-related...

danielpeintner (approved):
LGTM

PR #492, opened 2021-09-06
chore(binding-coap): bump node-coap to 0.25.0

JKRhb:
This PR updates the `coap` package used in the `binding-coap` package. As the new version now also supports the `application/td+json` Content-Format, this PR makes it possible to also fetch TDs which are provided as such without throwing an error.

This behavior fixed in the new version in general, so an invalid Content-Format cannot cause an application to crash anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4338,9 +4338,9 @@
       }
     },
     "node_modules/coap": {
-      "version": "0.24.0",
-      "resolved": "https://registry.npmjs.org/coap/-/coap-0.24.0.tgz",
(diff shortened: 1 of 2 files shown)


JKRhb:
I just realized that the `master` branch is probably not the best target for this PR considering the restructuring for 0.8.x. Do you think this could be integrated in the new release?

JKRhb:
> I just realized that the `master` branch is probably not the best target for this PR considering the restructuring for 0.8.x. Do you think this could be integrated in the new release?

I now updated the target branch accordingly.

I guess the package-lock file should be updated as well?

JKRhb:
I also updated the package-lock file as well – the PR should now be ready to be merged :)

danielpeintner (approved):
LGTM

relu91 (approved):

Issue #243, opened 2020-06-09
Remove codestyle scripts in package.json


danielpeintner:
Relates to https://github.com/eclipse/thingweb.node-wot/issues/239#issuecomment-640479670

We should remove `"codestyle": "standard --pretty"` or alike in *all* package.json
e.g., https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-coap/package.json#L37

We should properly handle it with [.editorconfig](https://github.com/eclipse/thingweb.node-wot/blob/master/.editorconfig)

danielpeintner:
@relu91 could you check PR https://github.com/eclipse/thingweb.node-wot/pull/259 ?

In a next step we can try to use VSC auto format feature and see how many diffs we see...

relu91:
>  @relu91 could you check PR #259 ?

I've just done my review.

> In a next step we can try to use VSC auto format feature and see how many diffs we see...

Yes exactly. However, I found that `.editorconfig` is quite limited regarding the expressivity of how the code should be formatted.  I have to admit that I do not use `.editorconfig` usually, so maybe I have to understand it a little bit better. `eslint` seems more advanced.

relu91:
closing we now have a new fancy code-style system

PR #493, opened 2021-09-06
Codestyle and format

### LINKED ISSUE #239: Inconsistent code style

relu91:
This PR addresses concerns raised in #239. It uses `eslint` and `prettier` as the main tools to achieve a consistent code style across the codebase. I decided to use the standard recommendation of `eslint` but people still can discuss this decision on this PR; once is merged is done. 

This PR is in a draft mode cause:
- I want to give people a chance to discuss it
- I currently limited `eslint` and `prettier` to the core package

Todos:
- [x] Since it is based on #491 first we need to fix the ci tests there
- [ ] Create a ci script verifying that submitted code is well formated. Should we add a new workflow or extend the current one?
- [ ] extends to other packages
- [ ] fix errors 
- [x] fix warnings (?)

Fixes #239 and #243


JKRhb:
I would say a separate workflow makes sense. This way it should be easier to add, for example, annotations into PR diffs when the codestyle isn't followed. It should also be easier to spot if the CI fails due to a style issue or because of a failing test.

danielpeintner:
I think the PR goes into the right direction 👍 . We need some consistent code style.

Having said that, we do not need to solve "all" in this PR.
i.e., the GitHub script to check can come later etc.

_My_ proposal is to apply the style to all packages and merge...

JKRhb:
I took the freedom of creating both a workflow for running eslint and prettier and would open a PR for that soon which could be merged after this PR.

danielpeintner:
To me it makes perfect sense but I would like to put it in @relu91  hands since he was the original author.

Note: we might take over the todos mentioned in this PR to another to better keep track.

danielpeintner:
@relu91 ready to be merged ... other aspects can come later?

JKRhb:
Maybe the final fixes can also be added in #503 as the linter should already point out what is missing.

relu91:
> Maybe the final fixes can also be added in #503 as the linter CI should already point out what is missing.

yes, why not! 👍🏻 @fatadel can you base your changes on #503? if you already started don't worry we'll merge your changes separately.

relu91:
Ok decided with @danielpeintner to merge this and continue on #503 🎉

fatadel:
> > Maybe the final fixes can also be added in #503 as the linter CI should already point out what is missing.
> 
> yes, why not! 👍🏻 @fatadel can you base your changes on #503? if you already started don't worry we'll merge your changes separately.

Yes, no problem for me 👍🏻

danielpeintner:
@fatadel  branch `v0.8.x` has been merged into `master`.
Hence all PRs from now on should be against  `master`....

PR #494, opened 2021-09-08
fix: rename wotfirestore to firestore

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/489


hidetak:
@danielpeintner 
Thank you for correcting the Binding name. I appreciate it very much.

I have checked the fix and there is no problem.
I'm sorry for my slow response.

hidetak (approved):
Thank you very much !

PR #503, opened 2021-09-10
chore: Add linting and code style checks to the CI

JKRhb:
This PR is intended to be a follow-up to #493 and adds both a linting action (using eslint) and style check action (using prettier) to the CI. Both actions are currently failing because there the code base is not yet formatted correctly but once #493 is completed all checks will hopefully turn green :)

I am not sure yet, however, if the prettier action is configured correctly (or if it should be swapped out for a similar prettier action) as it does not add annotations in the "files changed" tab yet. I will have a look into that.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,3 +47,19 @@ jobs:
       uses: icrawl/action-eslint@v1
       with:
         custom-glob: packages
+
+  prettier:
+    name: Check coding style
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: install node v12
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: yarn install
+      run: yarn install
+    - name: Prettify code
+      uses: creyD/prettier_action@v4.0
+      with:
+        dry: True
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,3 +31,19 @@ jobs:
 
       - name: Test
         run: npm run test
+
+  eslint:
+    name: eslint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: install node v12
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12
+    - name: yarn install
+      run: yarn install
+    - name: eslint
+      uses: icrawl/action-eslint@v1
+      with:
+        custom-glob: packages
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,13 @@ Simplify the API by reducing the number of functions.
 
 Closes #110
 BREAKING CHANGE: Change all calls to the API to the new `do()` function.
-``` 
+```
+
+## Coding Style
+
+Eclipse Thingweb uses `eslint` and `prettier` to enforce a consistent coding style.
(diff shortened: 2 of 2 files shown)


JKRhb:
I just realized that prettier seems to only be able to indicate if a file is not prettified and not which lines are containing style issues. So maybe the state of this PR is already fine.

danielpeintner:
Note: we might still need to do proper linting for packages other than `core`. I would expect to see many minor diffs..

JKRhb:
> Note: we might still need to do proper linting for packages other than `core`. I would expect to see many minor diffs..

Oh, yes, that is true! I think the actions should already check all packages, however.

danielpeintner:
Questions:
* Can we mark eslint issues like warnings only? Not sure.. e.g., https://github.com/bfanger/eslint-plugin-only-warn
* We might want to tweak some errors that are more a warning than an actual error... I think .. i see errors like the following in https://github.com/eclipse/thingweb.node-wot/pull/503/checks?check_run_id=3566959654
  *  Error:    47:9   error    Unexpected var, use let or const instead                               no-var
  *  Error:    97:23  error    ["events"] is better written in dot notation                           dot-notation
  *  Error:   103:7   error    Identifier 'td1_V2' is not in camel case                               camelcase

JKRhb:
> Questions:
> 
> * Can we mark eslint issues like warnings only? Not sure.. e.g., https://github.com/bfanger/eslint-plugin-only-warn

Sure! If we reconfigure eslint to use warnings instead of errors then the annotations should appear as warnings as well. This should also have the positive side-effect of making it unnecessary to change the rules from errors to warnings, by hand :)

By the way: I just realized that @relu91 already included prettier in the eslint config which makes the separate prettier check obsolete, I assume. I would therefore remove it, as it is less precise in pointing out errors anyway.

danielpeintner:
Note: I think we need to describe how anyone can do a local `eslint` check before creating a PR.
I guess we would need to add some documentation in README ...

JKRhb:
> Note: I think we need to describe how anyone can do a local `eslint` check before creating a PR.
> I guess we would need to add some documentation in README ...

Maybe this could be added to `CONTRIBUTING.md`?

JKRhb:
> > Note: I think we need to describe how anyone can do a local `eslint` check before creating a PR.
> > I guess we would need to add some documentation in README ...
> 
> Maybe this could be added to `CONTRIBUTING.md`?

I added a proposal in 9a970f580fbd421769d03b6da962d337731120e4. Feel free to request any changes :)

relu91 (approved):
PR looks good. The only problem is that eslint actually does not warn you if your file is not well-formatted with prettier (tried locally). Maybe it is a miss-configuration of es-lint. I'll double-check and then we can merge it.

relu91:
Reading [these notes](https://prettier.io/docs/en/integrating-with-linters.html#notes) from the prettier site it seems that is better not to tangle the two tools together. Can you re-add the prettier step in the workflow? We should be able to recognize those annoying end of line and tabs problems.

JKRhb:
> Reading [these notes](https://prettier.io/docs/en/integrating-with-linters.html#notes) from the prettier site it seems that is better not to tangle the two tools together. Can you re-add the prettier step in the workflow? We should be able to recognize those annoying end of line and tabs problems.

Sure, will do :)

JKRhb:
I think the action should work now :) I think there are no annotations but the action gives a "dry" output of what would have been changed and fails if there are any style issues.

relu91 (approved):
Good for me, merge? @danielpeintner

danielpeintner (approved):

PR #510, opened 2021-09-16
style: Reformat codebase using prettier

JKRhb:
This PR applies prettier to the whole codebase. The README files are currently being reformatted as well – let me know they should be excluded :)


JKRhb:
Hmm, if I am not mistaken, the prettier CI is going to fail because the rest of the files (e. g. `examples/`, `package.json`, `package-lock.json`, ...) are not formatted yet. Should prettier be applied to all files? Or only the ones in the `packages` directory?

danielpeintner:
> Hmm, if I am not mistaken, the prettier CI is going to fail because the rest of the files (e. g. `examples/`, `package.json`, `package-lock.json`, ...) are not formatted yet. Should prettier be applied to all files? Or only the ones in the `packages` directory?

w.r.t. `examples/`: some parts are manually written code, other parts like `examples/scripts/` are auto-generated. Anyhow, I don't think it would harm to apply prettier everywhere.

w.r.t. `package.json`: i think you got all of them... expect in `examples/` right? 

w.r.t. `package-lock.json`: I think there should be only on left, in the root folder and since this is automatically created it should be fine.

JKRhb:
@danielpeintner Thank you for your feedback! I now reformatted everything except for `package-lock.json`. Maybe we can add an ignore rule for this file?

danielpeintner:
> @danielpeintner Thank you for your feedback! I now reformatted everything except for `package-lock.json`. Maybe we can add an ignore rule for this file?

Thanks. Let's see whether this is actually necessary.. I hope it is generated already properly. If not we should put it on the ignore list.

JKRhb:
> Thanks. Let's see whether this is actually necessary.. I hope it is generated already properly. If not we should put it on the ignore list.

I think the generated file uses two spaces while prettier uses 4. Therefore, at the moment all lines are changed. However, I noticed that you can add a `.prettierignore` file which should solve this problem :)

JKRhb:
I think now everything should work :)

JKRhb:
> I think now everything should work :)

... except for one example file 😅 Should be fixed now :)

JKRhb:
Hmm. For some reason, the Action does not use  the`.prettierignore` file, even when explicitly pointed to. This seems to be a known issue: https://github.com/creyD/prettier_action/issues/64 I will try out a different action from the Github Marketplace

relu91:
How did you format all the codebase? I think this should be available from the main package.json. So that `npm run lint` would also cover the additional files covered by this PR. Can you modify the root folder `package.json` to make this happen?

JKRhb:
> How did you format all the codebase? I think this should be available from the main package.json. So that `npm run format` would also cover the additional files covered by this PR. Can you modify the root folder `package.json` to make this happen?

I actually simply ran `prettier --write .` – I'll add it to `package.json` :)

relu91:
What about integrating it with what we have now? Something like this:
```
"format": "prettier --write . && npm run format --silent --workspaces --if-present"
```

JKRhb:
> I actually simply ran `prettier --write .` – I'll add it to `package.json` :)

Hmm, `npm run format` seems to already work as well,? Oh, but it only covers the files in packages, right?

JKRhb:
> What about integrating it with what we have now? Something like this:
> 
> ```
> "format": "prettier --write . && npm run format --silent --workspaces --if-present"
> ```

Great, that looks nice! :)

JKRhb:
I now added an Action back in that simply reports if a file is unprettified. As formatting is done automatically anyway, this should be enough I guess :) I would clean up the commit history a little and then I think this PR could be merged :)

JKRhb:
Hmm, sorry for creating so many actions by the way. I guess, by now we could cancel all actions but the last one.

danielpeintner (approved):
LGTM.

@relu91 any pushback?

After merging this PR and merging the v0.8.x branch into master we should tackle the eslint warnings/errors.
Having said that we can postpone most of the eslint work and also do some actual *new* work for the PlugFest ;-)

relu91 (approved):
green light! 🟢

PR #512, opened 2021-09-16
chore: replace external eslint action, continue on failure

JKRhb:
This PR replaces the external Github Action for `eslint` by first installing it using `npm ci` and then running `npm run lint`. This removes the action as one additional dependency and also replicates the local environment for linting while still keeping the annotations made in the "Files changed" tab.

Furthermore, this PR lets the linting workflow pass even if errors occur, making it easier to see if there is an actual failing test. Once all linting issues have been resolved, this addition can be removed again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,6 +52,7 @@ jobs:
 
             - name: Lint
               run: npm run lint
+              continue-on-error: true
 
     prettier:
         name: Check coding style
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,17 +35,23 @@ jobs:
         name: eslint
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/checkout@v1
-            - name: install node v12
+            - name: Checkout
+              uses: actions/checkout@v2
+
+            - name: Use Node.js 14
               uses: actions/setup-node@v1
               with:
-                  node-version: 12
-            - name: yarn install
-              run: yarn install
-            - name: eslint
-              uses: icrawl/action-eslint@v1
-              with:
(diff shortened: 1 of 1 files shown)


danielpeintner:
General question: I wonder whether we want to have a green light in case of linting errors?

At the moment it is somewhat annoying to see the red x.
Having said that, once everything is in place I guess I would like to see the red x even in case for linting. I mean I can still check which task failed. Without any notification I assume it easily happens to merge PRs that should be first properly linted.

Any thoughts?

JKRhb:
> General question: I wonder whether we want to have a green light in case of linting errors?
> 
> At the moment it is somewhat annoying to see the red x.
> Having said that, once everything is in place I guess I would like to see the red x even in case for linting. I mean I can still check which task failed. Without any notification I assume it easily happens to merge PRs that should be first properly linted.
> 
> Any thoughts?

This is exactly what I was thinking :) Therefore this PR gives a green light for the linting action even if it fails but still displays the errors on the "Files changed" tab. When everything is fixed we can let the CI fail again if any issues are detected but for now this should make the process of opening PRs less annoying.

relu91:
Ok for me, but please open an issue as a reminder to re-enable linting when we have finished the fix the linting process. Having a red flag for linting error is really the point of having linting in the codebase :)

relu91 (approved):
Just to be clear we can go on :) @danielpeintner I let you have the final decision.

JKRhb:
> Ok for me, but please open an issue as a reminder to re-enable linting when we have finished the fix the linting process. Having a red flag for linting error is really the point of having linting in the codebase :)

Done :) I opened #516 as a reminder :)

danielpeintner (approved):
LGTM, thanks!

PR #513, opened 2021-09-17
fix: http server reporting circular structure

danielpeintner:
Note: this PR is not meant to be merged as is.

I just run the counter sample and with the current code I get

```
[binding-http] HttpServer on port 8080 got internal error on read '/counter/properties/count': Converting circular structure to JSON
    --> starting at object with constructor 'ReadableStream'
    |     property '_readableStreamController' -> object with constructor 'ReadableStreamDefaultController'
    --- property '_controlledReadableStream' closes the circle
```

I tried to fix it but I am not sure whether this is the right way...

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -749,20 +749,21 @@ export default class HttpServer implements ProtocolServer {
                                         .observeProperty(
                                             segments[3],
                                             async (value: InteractionOutput) => {
-                                                let content;
                                                 try {
                                                     const contentType = ProtocolHelpers.getPropertyContentType(
                                                         thing.getThingDescription(),
                                                         segments[3],
                                                         "http"
                                                     );
-                                                    content = ContentSerdes.get().valueToContent(
+                                                    const content = ContentSerdes.get().valueToContent(
                                                         value.data && !value.dataUsed
                                                             ? value.data
                                                             : await value.value(),
                                                         property.data,
                                                         contentType
                                                     );
+                                                    // send event data
+                                                    content.body.pipe(res);
                                                 } catch (err) {
                                                     console.warn(
                                                         "[binding-http]",
@@ -774,8 +775,6 @@ export default class HttpServer implements ProtocolServer {
                                                     res.end("Invalid Event Data");
                                                     return;
                                                 }
-                                                // send event data
-                                                content.body.pipe(res);
                                             },
                                             options
                                         )
@@ -979,18 +978,19 @@ export default class HttpServer implements ProtocolServer {
                                     .subscribeEvent(
                                         segments[3],
                                         async (value) => {
-                                            let content;
                                             try {
                                                 const contentType = ProtocolHelpers.getEventContentType(
                                                     thing.getThingDescription(),
                                                     segments[3],
                                                     "http"
                                                 );
-                                                content = ContentSerdes.get().valueToContent(
+                                                const content = ContentSerdes.get().valueToContent(
                                                     value.data && !value.dataUsed ? value.data : await value.value(),
                                                     event.data,
                                                     contentType
                                                 );
+                                                // send event data
+                                                content.body.pipe(res);
                                             } catch (err) {
                                                 console.warn(
                                                     "[binding-http]",
@@ -1002,8 +1002,6 @@ export default class HttpServer implements ProtocolServer {
                                                 res.end("Invalid Event Data");
                                                 return;
                                             }
-                                            // send event data
-                                            content.body.pipe(res);
                                         },
                                         options
                                     )
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -919,7 +919,7 @@ export default class HttpServer implements ProtocolServer {
                                     thing
                                         .invokeAction(segments[3], input, options)
                                         .then(async (output: InteractionOutput) => {
-                                            if (output) {
+                                            if (output && action.output) {
                                                 const contentType = ProtocolHelpers.getActionContentType(
                                                     thing.getThingDescription(),
                                                     segments[3],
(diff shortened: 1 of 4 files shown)


danielpeintner:
relates to 
https://github.com/eclipse/thingweb.node-wot/blob/59fe3264bb4cfa19a3c8cddb0f1c68b4685a4a47/packages/binding-http/test/http-server-test.ts#L73-L74

packages/binding-http/src/http-server.ts
```diff
@@ -769,21 +770,21 @@ export default class HttpServer implements ProtocolServer {
...
                                                 <any>property,
                                                 contentType
-                                            );
-                                            res.setHeader("Content-Type", content.type);
+                                            );*/
+                                            res.setHeader("Content-Type", contentType); // content.type);
                                             res.writeHead(200);
-                                            content.body.pipe(res);
+                                            res.end(await value.value() + "");
+                                            // content.body.pipe(res);
```
relu91, line 787: This commented code was needed to support Streams. This means that we'll use support for ExposedThing to answer with a stream of data, it can only answer with DataSchema values.

packages/core/src/interaction-output.ts
```diff
@@ -68,11 +68,11 @@ export class InteractionOutput implements WoT.InteractionOutput {
 
         const validate = ajv.compile<T>(this.schema);
 
-        // is content type vaild?
-        if (!this.form || !ContentSerdes.get().isSupported(this.content.type)) {
+        // is content type valid?
+        /* if (!this.form || !ContentSerdes.get().isSupported(this.content.type)) {
```
relu91, line 72: This code is needed, see the scripting API algorithm: https://w3c.github.io/wot-scripting-api/#the-value-function step 4

relu91 (changes requested):
I think we should find a more general approach to solve this issue. One starter is to define an internal content type (e.g., "application/internal" or similar) for values that are passed from the old consumed thing methods in `ExposedThing`. Later on, we should evaluate a refactoring where the `ExposedThing` class does not have `ConsumedThing` methods anymore.

packages/binding-http/src/http-server.ts
danielpeintner: added again

packages/core/src/interaction-output.ts
danielpeintner: applied again

packages/binding-http/test/http-server-test.ts
```diff
@@ -111,16 +110,16 @@ class HttpServerTest {
         console.log("Testing", uri);
 
         body = await (await fetch(uri + "properties/test")).text();
-        expect(body).to.equal('"off"');
+        expect(body).to.equal("off");
```
danielpeintner, line 113: I am a bit unsure why there was `'"off"'` in before?

packages/binding-http/src/http-server.ts
```diff
@@ -683,11 +691,24 @@ export default class HttpServer implements ProtocolServer {
...
+                                            let content = ContentSerdes.get().valueToContent(
+                                                value.data && !value.dataUsed ? value.data : await value.value(),
+                                                undefined,
+                                                "application/json"
+                                            );
+                                            // TODO ProtocolHelpers.readStreamFully() does not work for counter example for SVG countAsImage
+                                            const data = await ProtocolHelpers.readStreamFully(content.body);
+                                            recordReponse[key] = data.toString(); // contentType handling?
+                                        }
+                                        res.setHeader("Content-Type", "application/json"); // contentType handling?
```
danielpeintner, line 709: read all properties is *very* messy... piping does not work for me.

Moreover, `ProtocolHelpers.readStreamFully()` does not worj for all properties, e.g. SVG image in counter

danielpeintner:
Note: I would like to merge this PR soon so that we continue the work (e.g., https://github.com/eclipse/thingweb.node-wot/issues/522) and do some more fixes later.

Hence I propose to post-pone:
 - [ ] improve readAllProperties() with piping in HTTP
 - [ ] fixing readAllProperties with contentType (e.g., XML SVG in counter)

packages/binding-http/src/http-server.ts
relu91: I agree it is messy, but I think this is something from the spec itself. In all the aggregated operations we are failing to describe how the payload will look like... I remembered that we decided to go with the fact that each property value should be transmitted inside a JSON object similar to what you're doing. However, don't you think is better to just call the `value` function on each `key` in `propMap` and If it fails read the stream? Why are you using `valueToContent` ?

packages/binding-http/src/http-server.ts
```diff
@@ -752,7 +775,7 @@ export default class HttpServer implements ProtocolServer {
                                                     return;
                                                 }
                                                 // send event data
-                                                res.end(content.body);
+                                                content.body.pipe(res);
```
relu91, line 778: while we are here I think this can be moved inside the try block.

packages/core/src/interaction-output.ts
```diff
@@ -42,7 +42,9 @@ export class InteractionOutput implements WoT.InteractionOutput {
         this.form = form;
         this.schema = schema;
 
-        this.data = ProtocolHelpers.toWoTStream(content.body);
+        if (content && content.body) {
```
relu91, line 45: `body` should be mandatory in content (and also `content` should be different from undefined), do we have a calling code that is passing undefined content?

relu91 (changes requested):
PR looks good! there are some minor comments to be fixed. I have also another concern but we can fix it in a follow-up PR: I think that the protocol bindings server should not care about content types as much as possible (maybe write/readall is an exception). But in principle, they should just receive the InteractionOuput (already encoded) from the upper level and pipe it to the output stream

packages/binding-http/src/http-server.ts
danielpeintner: 👍

packages/core/src/interaction-output.ts
danielpeintner: There were issues with it, e.g., failing tests with actions that do not return any data.

I thought this is safer since (at least at the moment the http tests fail without).
I added your point to the follow-up list

relu91 (approved):

PR #514, opened 2021-09-18
fix(binding-mqtt): fix 'TypeError: error is not a function"'

miguelrk:
I found a minor bug in the `binding-mqtt/src/mqtt-client.js` lines 70 and 76 with a duplicate `error` definition (in subscribeResource() and then again in "error" handler).

This PR simply changes lines 70 and 76 to fix this.

This was throwing the following error:

```bash
[ERROR] 17:13:25 TypeError: error is not a function
```


danielpeintner (approved):
Good catch. Thanks!

PR #517, opened 2021-09-20
fix(coap): Remove double registration of TD content-format

JKRhb:
This PR removes the unnecessary registration of the content-format `application/td+json` from the coap package. It is now supported [by default](https://github.com/mcollina/node-coap#coapregisterformatname-value) in `node-coap`, using the correct [IANA ID 432](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats).

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -43,10 +43,6 @@ export default class CoapClient implements ProtocolClient {
 
         // WoT-specific content formats
         coap.registerFormat(ContentSerdes.JSON_LD, 2100);
-        // TODO also register content fromat with IANA
-        // from experimental range for now
-        coap.registerFormat(ContentSerdes.TD, 65100);
-        // TODO need hook from ContentSerdes for runtime data formats
     }
 
     public toString(): string {
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -48,10 +48,6 @@ export default class CoapServer implements ProtocolServer {
 
         // WoT-specific content formats
         coap.registerFormat(ContentSerdes.JSON_LD, 2100);
-        // TODO also register content fromat with IANA
-        // from experimental range for now
-        coap.registerFormat(ContentSerdes.TD, 65100);
-        // TODO need hook from ContentSerdes for runtime data formats
     }
 
     public start(servient: Servient): Promise<void> {


packages/binding-coap/src/coap-client.ts
```diff
@@ -44,9 +44,6 @@ export default class CoapClient implements ProtocolClient {
         // WoT-specific content formats
         coap.registerFormat(ContentSerdes.JSON_LD, 2100);
         // TODO also register content fromat with IANA
-        // from experimental range for now
-        coap.registerFormat(ContentSerdes.TD, 65100);
-        // TODO need hook from ContentSerdes for runtime data formats
```
JKRhb, line 49: Hmm, should this TODO be kept?

packages/binding-coap/src/coap-client.ts
```diff
@@ -44,9 +44,6 @@ export default class CoapClient implements ProtocolClient {
         // WoT-specific content formats
         coap.registerFormat(ContentSerdes.JSON_LD, 2100);
```
JKRhb, line 45: Another point: I guess there are no plans to actually register the JSON-LD content-format for IANA, right?

packages/binding-coap/src/coap-client.ts
danielpeintner: Good question, not really sure. Work was mainly driven by @mkovatsc at that time.

packages/binding-coap/src/coap-client.ts
```diff
@@ -44,9 +44,6 @@ export default class CoapClient implements ProtocolClient {
...
         // TODO also register content fromat with IANA
```
danielpeintner, line 46: typo: fromat

packages/binding-coap/src/coap-client.ts
danielpeintner: Not as far I know 🙈

packages/binding-coap/src/coap-client.ts
JKRhb: I now simply removed these lines as the registration of `application/td+json` is already done :) Or do you think this is still a TODO for the WoT Working Group?

danielpeintner: This is more a question of the working group rather than node-wot. Hence I think it is ok!

danielpeintner (approved):

danielpeintner:
We did not get any further concerns w.r.t. this PR so I guess it is fine to go ahead.

Thanks @JKRhb

PR #518, opened 2021-09-20
fix(docker): fix docker file

erossignon:
- make sure typescript source are included in packages


erossignon:
## how it works

The script create tar.gz package for all node-wot related modules, and store them in the tmp folder.
The docker file  then simply installed the tar.gz packages and that's it.

Rather that copying the source tree in the docker container and rebuilding the project entirely inside the container, it simply create packages as tar.gz file externally from the host machine, transfer them to the docker container and install them as packages.

As a consequence, the build time of the docker container is quick, the final image size is also far smaller.

danielpeintner:
Question: Who can review this PR ?

relu91:
@hyperloris did the first contribution about the docker file, I think he can give us feedback. 

> Rather that copying the source tree in the docker container and rebuilding the project entirely inside the container, it simply create packages as tar.gz file externally from the host machine, transfer them to the docker container and install them as packages.

As far as I remember the dockerfile is designed on purpose to build the source file inside the container. This should help to have reproducible builds. For example, if you have native dependencies what happens with your approach? will they work if they are compiled on my windows machine and then moved inside the docker container? I don't think so... 🤔

erossignon:
> compiled on my windows machine and then moved inside the docker container? I don't think so... 🤔

In fact, I generate the same package that would be published on NPM, so none of the dependencies are included. Dependencies will be fetch and installed from within the container.

relu91:
Ah I see! but how this would be faster than just doing it inside the container? Can you point me to an example?

erossignon:
> Ah I see! but how this would be faster than just doing it inside the container? Can you point me to an example?

I started this work while before 0.8 was merged on master, on top of 0.7 at that time the docker creation time was long and the size of the imaged around 1GB or more . This was due to the fact that packages were installed locally inside each sub modules.
The initial rational of my attempt to improve the docker image generation comes from here . I only created the PR after 0.8 was merged to main branch.

Since then a few things have happened:
  - we use npm@7 and workspace => as a consequence, shared modules get installed into the top node_modules folder
  - we fix the version discrepancies 
  - as a consequence also: the docker file is not functioning anymore .

I have try to fix the original docker file yesterday by using npm@7 install. the packages get build but the COPY command in the second part of the dockerfile lead to copying the submodule to the final image without the need external modules, because those one are now installed at the top-level node_module folder. This defeats the assumptions made by the original docker file. 

I think that the new method I am proposing is still valid and is fixing all of this.

erossignon:
As an experiment, I have added a second DockerFile that copies the sources from the external directory and rebuilds everything from within the container. It uses also a 2 step image build, so we can get rid of the tooling in the second image.

./docker/Dockerfile => use externally generated tar.gz packages, build in one step
./docker/DockerVariation2 => copy source in docker image , build from within container , 2 step image generation

The 2 technics produce almost similar image size, time to build is not a problem for either build method

wiresio:
Just tested, but the scripts terminates:

```
Step 3/18 : COPY tmp/*.tgz .
When using COPY with more than one source file, the destination must be a directory and end with a /
```

erossignon:
the build process makes the assumption that the project has been build on the host first. I will add a `npm run build` step to the build.sh script, just to make sure that this is done.

erossignon:
I have added two scrips in the package.json to build the 2 flavors of docker images:


````
npm run build:docker # to build the wot-servient  image
````

````
npm run build:docker2 # to build the wot-servient2  image
````

File: docker/DockerfileVariation2 [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+    linux-headers \
+    udev \
+    python3
+RUN npm install -g npm@7
+
+WORKDIR /app
+COPY ./package.json .
+COPY ./packages packages/
+
+RUN npm install --production
```
relu91, line 15: I think you're missing to build the typescript files using `tsc`.

relu91 (changes requested):
> we use npm@7 and workspace => as a consequence, shared modules get installed into the > top node_modules folder
> we fix the version discrepancies
> as a consequence also: the docker file is not functioning anymore .

I see, but still, I think the PR is not ready to be merge. First of all, I think we should choose one of the two proposed approaches cause they seem to be equal. Second, I really like conventions and here we are forcing people to understand why we use two docker files and why one has an additional build step using sh. 

My suggestion to improve this proposal would be:
- getting rid of the build.sh ( if you think about it sh is not really supported on windows and this would force people to install MinGW or similar) 
- move DockerfileVariation2 to ./dockerfile
- run `npm run build` inside the building container

erossignon:
Thank you @relu91 , so let's choose the variation2 as the official way to build the image.
I'll update the PR.

relu91:
I moved back the docker file to the root folder and used npm prune instead of npm install --production again. 

Using workspaces there might be slowdowns about caching (i.e. now it re-install every time you need to build the image), we can improve it in a future PR.

PR #519, opened 2021-09-20
fix(binding-opcua;binding-coap) fix some test issues and resource leaks

erossignon:
in particular, some resources leaks in coap and opcua test that prevents tests to terminate..

As a consequence it should be possible after this fix to run mocha without specifying --exit options to force termination.

### DIFF
--- a/packages/binding-coap/test/coap-client-test.ts
+++ b/packages/binding-coap/test/coap-client-test.ts
@@ -86,13 +86,12 @@ class CoapClientTest {
         await coapServer.stop();
     }
 
-
     @test async "should re-use port"() {
         const coapServer = new CoapServer(56834, "localhost");
         await coapServer.start(null);
         const coapClient = new CoapClient(coapServer);
-        const res =  await coapClient.readResource({
-            href: "coap://localhost:56834/"
+        const res = await coapClient.readResource({
+            href: "coap://localhost:56834/",
         });
         false && console.log(res);
         await coapServer.stop();
@@ -104,9 +103,11 @@ class CoapClientTest {
         const coapClient = new CoapClient(coapServer);
         const form: CoapForm = {
             href: "coap://127.0.0.1:56834",
-            "coap:methodCode": "GET",
+            "cov:methodName": "GET",
         };
-        const subscription = await  coapClient.subscribeResource(form, (value) => { /**  */});
+        const subscription = await coapClient.subscribeResource(form, (value) => {
+            /**  */
+        });
         subscription.unsubscribe();
         await coapServer.stop();
     }
--- a/packages/binding-http/test/http-client-test.ts
+++ b/packages/binding-http/test/http-client-test.ts
@@ -74,7 +74,7 @@ class TestHttpServer implements ProtocolServer {
     }
 
     public async start(): Promise<void> {
-        return await new Promise<void>((resolve, reject) => {
+        return new Promise<void>((resolve, reject) => {
             this.server.once("listening", () => {
                 resolve();
             });
@@ -83,7 +83,7 @@ class TestHttpServer implements ProtocolServer {
     }
 
     public async stop(): Promise<void> {
-        return await new Promise<void>((resolve, reject) => {
+        return new Promise<void>((resolve, reject) => {
             this.server.close(() => {
                 console.error("STOPPED");
                 resolve();
@@ -173,7 +173,7 @@ class TestHttpServer implements ProtocolServer {
     }
 }
 
-@suite("HTTP client implementation 1")
+@suite("HTTP client basic operations")
 class HttpClientTest1 {
     static httpServer: TestHttpServer;
     static async before(): Promise<void> {
@@ -316,7 +316,7 @@ const express = require("express");
 const serveStatic = require("serve-static");
 const SseStream = require("ssestream");
 
-@suite("HTTP client implementation 2")
+@suite("HTTP client subscriptions")
 class HttpClientTest2 {
     @test "should register to sse server and get server sent event"(done: any) {
         // create sse server


packages/binding-http/test/http-client-test.ts
```diff
@@ -13,6 +13,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
 
+/* eslint-disable @typescript-eslint/no-var-requires */
```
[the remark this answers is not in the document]
erossignon: done

danielpeintner:
The tests fail
* not really sure why, see https://github.com/eclipse/thingweb.node-wot/pull/519/checks?check_run_id=3671865991
* coding style, please call `npm run format`

relu91:
I am wondering why we didn't see any blocking issue before. 🤔 Even the latest PRs completed the test successfully. 

Another note: it seems that this PR also solves eslint errors, but it is quite confusing since opc-ua is not actually included (yet) among eslint checked packages. In other words, it does not have the lint script in the package `json`, it should be added before fixing the eslint errors.

erossignon:
> lint issue:

My vscode is configured with the Prettier and eslint and highlights automatically the formatting or eslint error in the editor,
Error are easily spotted. it doesn't harm to fix some error now even if this project  is not formally checked yet. we can add this in a different PR.

> why problem was not spotted before ?

because mocha is configured to uses the --exit flag, and the --exit flag hides the problem. I will remove the --exit flag in a second phase.

relu91:
Where is the exit flag? 
https://github.com/eclipse/thingweb.node-wot/blob/1cd63abc520b9dcba96e3b88e348cbd83752f673/packages/binding-opcua/package.json#L51

https://github.com/eclipse/thingweb.node-wot/blob/1cd63abc520b9dcba96e3b88e348cbd83752f673/packages/binding-coap/package.json#L47

> My vscode is configured with the Prettier and eslint and highlights automatically the formatting or eslint error in the editor,
Error are easily spotted. it doesn't harm to fix some error now even if this project is not formally checked yet. we can add this in a different PR.

Yeah also mine, but I would prefer to have consistency so if you don't mind I'd add the lint command in this PR.

JKRhb:
It seems to me as if the CI got stuck in the tests for 06eb231. Maybe this run can be cancelled?

danielpeintner:
> It seems to me as if the CI got stuck in the tests for [06eb231](https://github.com/eclipse/thingweb.node-wot/commit/06eb2310aec5a46e7c928ac237377de42c5f3d1c). Maybe this run can be cancelled?

cancelled the workflow

JKRhb:
Hmm, I am afraid the CI got stuck again.

JKRhb:
Running the tests for this PR locally, it seems that the OPCUA client tests pass but the test suite doesn't terminate.

erossignon:
Sorry, I  put this PR into draft mode in order to prevent unnecessary reviewing while this is being still underwork, 
I have excluded temporarily the attempt to fix the binding-opcua => this will come in a separate PR.

I will remove the "Draft" status when the PR will be ready for review.

erossignon:
I have separated the going fix for the binding-opcua into an other PR.

danielpeintner:
Most of the changes are in HTTP. @relu91 may I ask you to take a look?

packages/binding-http/test/http-client-basic-test.ts
```diff
@@ -70,33 +74,51 @@ class HttpClientBasicTest {
...
+        const error = await new Promise<Error>((resolve, reject) => {
+            this.client
+                .readResource({ href: "https://localhost:3001" })
+                .then(() => reject("Expecting readResource to fail"))
+                .catch((err: unknown) => {
+                    resolve(err as Error);
+                });
+        });
+
+        error.message.should.eql("Client error: Unauthorized");
```
relu91, line 122: can you rewrite this check using `try-catch + fail`?

packages/binding-http/test/http-client-test.ts
```diff
@@ -66,19 +68,22 @@ class TestHttpServer implements ProtocolServer {
...
         }
+        this.server = http.createServer((req, res) => {
+            this.checkRequest(req, res);
+        });
     }
 
-    public start(): Promise<void> {
-        return new Promise<void>((resolve, reject) => {
+    public async start(): Promise<void> {
+        return await new Promise<void>((resolve, reject) => {
```
relu91, line 77: I would keep the previous code, I don't see the benefit of this inline promise.

packages/binding-http/test/http-client-test.ts
```diff
@@ -66,19 +68,22 @@ class TestHttpServer implements ProtocolServer {
...
                 resolve();
             });
             this.server.listen(this.port, this.address);
         });
     }
 
-    public stop(): Promise<void> {
-        return new Promise<void>((resolve, reject) => {
+    public async stop(): Promise<void> {
+        return await new Promise<void>((resolve, reject) => {
```
relu91, line 86: same as above

packages/binding-http/test/http-client-test.ts
```diff
@@ -168,127 +173,153 @@ class TestHttpServer implements ProtocolServer {
...
+        HttpClientTest1.httpServer.setTestVector(inputVector4);
+        const invokeResourceResult2 = await this.client.invokeResource(inputVector4.form);
     }
+}
+
+const express = require("express");
+const serveStatic = require("serve-static");
+const SseStream = require("ssestream");
 
+@suite("HTTP client implementation 2")
```
relu91, line 319: Can't we call these: "HTTP client subscriptions"?

packages/binding-http/test/http-client-test.ts
```diff
@@ -168,127 +173,153 @@ class TestHttpServer implements ProtocolServer {
     }
 }
 
-@suite("HTTP client implementation")
-class HttpClientTest {
-    @test async "should apply defaults"() {
-        let inputVector;
+@suite("HTTP client implementation 1")
```
relu91, line 176: What about "HTTP client basic operations"?

relu91 (commented):
I have couple of comments/suggestions after resolving those we can merge it 😃

relu91:
Just to be sure, so the blocking issues were not in the HTTP binding right? I'm just concerned about the fact that we've never spotted it before.

packages/binding-http/test/http-client-basic-test.ts
```diff
@@ -70,33 +74,51 @@ class HttpClientBasicTest {
         });
     }
 
+    static async after(): Promise<void> {
+        return new Promise<void>((resolve) => {
+            HttpClientBasicTest.server.close(() => resolve());
+        });
+    }
+
```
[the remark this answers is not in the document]
erossignon: Agreed, but promisify was only used once here. I used the more common pattern used elsewhere for consistency.

packages/binding-http/test/http-client-test.ts
erossignon: arguably, any method returning Promise shall be marked as async. I used this convention all across node-opcua for consistency, and I idiomatically applied the same convention here.

I will follow your convention in node-wot if you find appropriate to sometime have the async keyword and sometimes not , but I will also need clear rule to understand when to apply it and when not.

relu91: Sorry I was referring to `await new Promise` at the end isn't it the same as simply returning ` new Promise`? 

About `async`, @danielpeintner already knows that I am fan of `async` keyword, so I would never stop you to add it 👍🏻 .

packages/binding-coap/test/coap-client-test.ts
```diff
@@ -97,7 +97,9 @@ class CoapClientTest {
                     return coapServer.stop();
                 })
                 .then(() => {
-                    done();
+                    coapServer.stop().then(() => {
```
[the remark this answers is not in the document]
erossignon: agreed

relu91:
One note: once we have fixed the CI issues here, we have to check if some of the [coap open issues](https://github.com/eclipse/thingweb.node-wot/issues?q=is%3Aopen+is%3Aissue+label%3Abinding-coap) are solved by this.

danielpeintner:
> may I suggest that we use "rebase merging" as a default merge to master strategy to help creating a nice linear history in git ?

I don't have a strong opinion.
I just noticed that @JKRhb squashes commits usually ... hence there seem to be different preferences.

PR #520, opened 2021-09-20
refactor: align with eslint rules #2

fatadel:
The second portion of eslint alignments (previous one - https://github.com/eclipse/thingweb.node-wot/pull/511).

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -498,7 +498,8 @@ export default class CoapServer implements ProtocolServer {
                                 res._packet.token = res._request.token;
                                 // end of work-around
 
-                                thing.subscribeEvent(
+                                thing
+                                    .subscribeEvent(
                                         segments[3],
                                         // let subscription = event.subscribe(
                                         (data) => {
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -32,9 +32,11 @@ export default class CoapServer implements ProtocolServer {
 
     private readonly port: number = 5683;
     private readonly address: string = undefined;
+
     private readonly server: any = coap.createServer((req: any, res: any) => {
         this.handleRequest(req, res);
     });
+
     private readonly things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
     private servient: Servient = null;
 
@@ -200,7 +202,7 @@ export default class CoapServer implements ProtocolServer {
             } else {
                 console.info("[binding-coap]", `CoapServer failed to destroy thing with thingId '${thingId}'`);
             }
-            resolve(removedThing != undefined);
+            resolve(removedThing !== undefined);
         });
     }
 
@@ -324,7 +326,7 @@ export default class CoapServer implements ProtocolServer {
                                     });
                                 // observeproperty
                             } else {
-                                var oInterval = setInterval(function () {
+                                const oInterval = setInterval(function () {
                                     thing
                                         .readProperty(segments[3])
                                         // property.read() periodically
@@ -344,6 +346,9 @@ export default class CoapServer implements ProtocolServer {
                                             res.write(content.body);
 
                                             res.on("finish", function (err: Error) {
+                                                if (err) {
+                                                    console.error(err);
+                                                }
                                                 clearInterval(oInterval);
                                                 res.end();
                                             });
@@ -485,7 +490,7 @@ export default class CoapServer implements ProtocolServer {
                                 packet.payload = "";
                                 packet.reset = false;
                                 packet.ack = true;
-                                packet.token = new Buffer(0);
+                                packet.token = Buffer.alloc(0);
 
                                 res._send(res, packet);
 
@@ -493,8 +498,7 @@ export default class CoapServer implements ProtocolServer {
                                 res._packet.token = res._request.token;
                                 // end of work-around
 
-                                const subscription = thing
-                                    .subscribeEvent(
+                                thing.subscribeEvent(
                                         segments[3],
                                         // let subscription = event.subscribe(
                                         (data) => {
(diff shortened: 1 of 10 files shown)


danielpeintner:
Thanks @fatadel !
Can you also run `npm run format` to fix the _coding style_ issues...

packages/binding-coap/src/coap-server.ts
```diff
@@ -344,6 +346,9 @@ export default class CoapServer implements ProtocolServer {
                                             res.write(content.body);
 
                                             res.on("finish", function (err: Error) {
+                                                if (err) {
+                                                    console.error(err);
```
relu91, line 350: For logging we need to use our convention, I am wondering if we can enforce it somehow with eslint 🤔

packages/binding-coap/test/coap-client-test.ts
```diff
@@ -88,7 +86,7 @@ class CoapClientTest {
         await coapServer.stop();
     }
 
-    @test "should re-use port"(done: Function) {
+    @test "should re-use port"(done: () => {}) {
```
relu91, line 89: There's a type in mocha for the done function, I think is `Mocha.Done`

packages/binding-coap/test/coap-client-test.ts
```diff
@@ -30,15 +28,15 @@ should();
...
-        );
+        // const testThing = Helpers.extend(
+        //     {
+        //         name: "Test",
+        //         properties: {
+        //             test: {},
+        //         },
+        //     },
+        //     new ExposedThing(null)
+        // );
```
relu91, line 39: Reminder to ourselves, we should avoid commented code wherever is possible. (I don't blame you here cause I think is correct to comment it in the scope of this PR. Perhaps, can you open an issue pointing to this test? in that way we can review and implement a new test or decide to remove it once for all

relu91 (changes requested):
Small changes, but it looks good!

fatadel:
> npm run format

Didn't know about this, thank you!

packages/binding-coap/src/coap-server.ts
fatadel: What do you mean by "our convention"?
I hope you've understood that I've added that code because of this rule - https://eslint.org/docs/rules/handle-callback-err.

packages/binding-coap/test/coap-client-test.ts
fatadel: Done ✅

fatadel:
> Thanks @fatadel !
> Can you also run `npm run format` to fix the _coding style_ issues...

Done ✅

packages/binding-coap/test/coap-client-test.ts
fatadel: Done! - https://github.com/eclipse/thingweb.node-wot/issues/523
Yeah, here I've commented it out because the variable is created but never used (because the lines where it's used were already commented).

packages/binding-coap/src/coap-server.ts
danielpeintner: If you take a look at our packages and the console logs you will notice that we add a prefix to the error so that we can easily  detect from which binding the error came from...

e.g., 

https://github.com/danielpeintner/thingweb.node-wot/blob/59fe3264bb4cfa19a3c8cddb0f1c68b4685a4a47/packages/binding-coap/src/coap-server.ts#L72-L74

relu91: yes exactly, can you add that prefix, please? :)

fatadel: Aah okay, sure! I thought you mean there is some kind of convention for the way we handle callback errors.

Will be fixed now 😊

fatadel: Done ✅ Is this sufficient?

danielpeintner: maybe it can be more verbose mentioning the TS file causing the issue

console.error("[binding-coap]", `CoapServer for port ${this.port} failed on observe with: ${err.message}`);

fatadel: I've fixed it. But a short question - won't there be a problem with referencing `this`?

danielpeintner: I don't think so. It is used elsewhere in the file also. Or do I miss anything?

danielpeintner (approved):

danielpeintner:
@relu91 do you see any more issues to be resolved?

packages/binding-coap/src/coap-server.ts
relu91: > I've fixed it. But a short question - won't there be a problem with referencing this?

Actually, he might be right, cause the whole code is wrapped inside `function in this case if I recall correctly this points to the `function` object. @fatadel can you do a couple of tests with a small script reproducing the situation? Then we can merge this! 😃

relu91: while I was here I did the test myself 😄 :
```js
class Test {
    constructor() {
        this.title = "Test";
    }
    testMeFunction(){
        setInterval( function () {
            console.log("function", this.title);
        }, 1000);
    }

    testMeArrow() {
        setInterval( () => {
            console.log("arrow",this.title);
        }, 1000);
    }
}

new Test().testMeFunction();
new Test().testMeArrow();

```
output: 
```
function undefined
arrow Test
function undefined
arrow Test
function undefined

arrow Test
```

Consequently, I'd ask @fatadel to change `function` into an arrow function :)

danielpeintner: Note: I guess we need to check other places than also...

fatadel: @relu91, done!
Should I also check for `this` inside functions in other places or let's leave it for a separate PR?

relu91 (approved):
Looks good let's move on.

PR #526, opened 2021-09-22
Subscribe now returns a Promise

relu91:
As specified in the [new scripting API spec](https://w3c.github.io/wot-scripting-api/#the-subscribeevent-method)(step 8), now the subscribe primitive returns a Promise that is resolved once the subscription completes successfully.


danielpeintner (approved):
LGTM

PR #528, opened 2021-09-22
fix: allow invokeResource without any parameter (CoAP client)

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/522

Note1: tested with counter sample using CoAP for all forms.

Note2: I would like to tackle https://github.com/eclipse/thingweb.node-wot/issues/523 in a separate PR.

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -95,31 +95,25 @@ export default class CoapClient implements ProtocolClient {
     }
 
     public invokeResource(form: CoapForm, content?: Content): Promise<Content> {
-        return new Promise<Content>((resolve, reject) => {
-            ProtocolHelpers.readStreamFully(content.body)
-                .then((buffer) => {
-                    const req = this.generateRequest(form, "POST");
+        return new Promise<Content>(async (resolve, reject) => {
+            const req = this.generateRequest(form, "POST");
 
-                    console.debug("[binding-coap]", `CoapClient sending ${req.statusCode} to ${form.href}`);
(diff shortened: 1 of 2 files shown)


packages/binding-coap/src/coap-client.ts
```diff
@@ -95,31 +95,25 @@ export default class CoapClient implements ProtocolClient {
...
-                })
-                .catch(reject);
+            req.on("response", (res: any) => {
+                console.debug("[binding-coap]", `CoapClient received ${res.code} from ${form.href}`);
+                console.debug("[binding-coap]", `CoapClient received Content-Format: ${res.headers["Content-Format"]}`);
+                console.debug("[binding-coap]", `CoapClient received headers: ${JSON.stringify(res.headers)}`);
+                const contentType = res.headers["Content-Format"];
+                resolve({ type: contentType, body: Readable.from(res.payload) });
+            });
+            req.on("error", (err: Error) => reject(err));
```
JKRhb, line 110: There is also a `timeout` event which can occur in `node-coap` (see [here](https://github.com/mcollina/node-coap#messageontimeout-functionerr--) which should probably be handled like so:

```suggestion
            req.on("error", (err: Error) => reject(err));
            req.on("timeout", (err: Error) => reject(err));
```

However, I guess this should be added in general and we could also open another PR for that.

JKRhb: I just realized that requests are confirmable by default and when they time out, there is always an error emitted. Therefore, this comment is obsolete :)

danielpeintner:
@relu91 any objections to merge this PR?

danielpeintner:
I tend to go ahead since it is a rather minor change. 

@JKRhb seems to have reviewed it also.

PR #531, opened 2021-09-23
chore: add missing eslint and prettier configs

### LINKED ISSUE #530: .eslintrc.json in some packages only
I noticed that we do have the file `.eslintrc.json` in some packages only.

e.g., it is missing for
* binding-mbus
* binding-fujitsu
* binding-mqtt
* ...

Wouldn't it be possible to have it in the root directory only?
@JKRhb

JKRhb:
This PR should resolve #530 by adding missing eslint and prettier configs. At the moment, I added the additional eslint configs to the package.json files as I thought it might make the packages cleaner in general. However, I now think it might be better to have separate `.eslintrc.json` files to make it easier to see if an eslint configuration exists. Let me know what you think about this :)

There is currently still an issue in the `examples` package where the `npm run lint` script fails with the following error message:

```
Oops! Something went wrong! :(

ESLint: 7.32.0

No files matching the pattern "." were found.
Please check for typing mistakes in the pattern.
```

Do you have an idea how to fix this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -233,10 +233,6 @@
             "resolved": "packages/binding-opcua",
             "link": true
         },
-        "node_modules/@node-wot/binding-opcua2": {
-            "resolved": "packages/binding-opcua2",
-            "link": true
-        },
         "node_modules/@node-wot/binding-oracle": {
             "resolved": "packages/binding-oracle",
             "link": true
@@ -9889,6 +9885,7 @@
                 "prettier": "^2.3.2",
                 "ts-node": "10.1.0",
                 "typescript": "4.1.3",
+                "typescript-standard": "^0.3.36",
                 "wot-typescript-definitions": "^0.8.0-SNAPSHOT.18"
             }
         },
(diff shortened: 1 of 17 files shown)


danielpeintner:
Mhh, the only difference I see is that it *might* have a somewhat different structure.

e.g., no TS files in packages\examples\src
or
packages\examples\src\security  ?

JKRhb:
> Mhh, the only difference I see is that it _might_ have a somewhat different structure.
> 
> e.g., no TS files in packages\examples\src
> or
> packages\examples\src\security ?

Hmm, I tried a couple of things out now but nothing seems to fix the problem :/ I guess we can leave it as is for now as a reminder and revisit the problem later – at the moment the check succeeds anyway.

danielpeintner (approved):

JKRhb:
@danielpeintner I rebased the PR and resolved the merge conflicts :) As soon as the CI passes I think we could merge this one :)

Issue #523, opened 2021-09-21
Commented out test in coap-client-test.ts


fatadel:
https://github.com/eclipse/thingweb.node-wot/blob/fc359193faea478e6682946fbde79d35c73205c5/packages/binding-coap/test/coap-client-test.ts#L51

We need to decide either to remove the test completely or reimplement it without massive commented out blocks of code.
This issue was initiated by @relu91 while reviewing my PR - https://github.com/eclipse/thingweb.node-wot/pull/520.

relu91:
After #538 and #602 we should probably re-evaluate this issue. @danielpeintner can you give it a look?

PR #533, opened 2021-09-23
fix: align CoAP vocabulary with current binding-templates

JKRhb:
While having a look at the current CoAP implementation in the context of #523, I noticed that the vocabulary that is being used is not up to date with the current binding-templates. This PR should fix this and should provide the basis for fixing #523.

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -26,7 +26,14 @@ import { Subscription } from "rxjs/Subscription";
 import * as TD from "@node-wot/td-tools";
 
 import { ProtocolClient, Content, ContentSerdes, ProtocolHelpers } from "@node-wot/core";
-import { CoapForm, CoapRequestConfig, CoapOption } from "./coap";
+import {
+    CoapForm,
+    CoapRequestConfig,
+    CoapOption,
+    CoapMethodName,
+    isSupportedCoapMethod,
+    isValidCoapMethod,
+} from "./coap";
 import CoapServer from "./coap-server";
 import { Readable } from "stream";
 const coap = require("coap");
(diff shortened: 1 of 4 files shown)


packages/binding-coap/src/coap-client.ts
```diff
@@ -214,31 +214,7 @@ export default class CoapClient implements ProtocolClient {
...
-                    break;
-                default:
-                    console.warn(
-                        "[binding-coap]",
-                        "CoapClient got invalid 'methodCode', using default",
-                        options.method
-                    );
-            }
-        }
+        options.method = form["cov:methodName"] ? form["cov:methodName"] : dflt;
```
JKRhb, line 217: With this change both the CoAP and the CoAPS clients would be more similar to the HTTP client. Let me know if the debug messages should be kept.

danielpeintner: I wonder what should happen IF it is not a *known* method. Fail I guess?

It might be useful to extract the current list in coap.ts so that here we can check whether its is part of this list before setting the value.

packages/binding-coap/src/coap.ts
```diff
@@ -27,14 +27,14 @@ export * from "./coap-client";
...
+    public "cov:optionName": number;
+    public "cov:optionValue": any;
 }
 
-export class CoapOption {
-    public "coap:optionCode": number;
-    public "coap:optionValue": any;
+export class CoapForm extends Form {
+    public "cov:methodName"?: "GET" | "POST" | "PUT" | "DELETE" | "FETCH" | "PATCH" | "iPATCH";
+    public "cov:options"?: Array<CoapOption> | CoapOption;
```
danielpeintner, line 37: First I was confused about the change but in the end you just reorderd CoapOption and CoapForm which causes these changes... might be ok to leave the order so it is more visible "what" has been changed....

danielpeintner:
Looks good. See some minor comments above.

Question do you plan to tackle #523 in this PR or another PR? This would be very good also!

packages/binding-coap/src/coap.ts
JKRhb: I was a bit confused myself when I was looking at the this change in the PR :D The rationale behind it was that eslint complained that `CoapForm` was used before it was defined. But I agree, to make things easier to track it would be better if the linting change would be done separately. I'll therefore revert the reordering :)

danielpeintner: ahh eslint.. I see
I am fine leaving it as is ... no need to shuffle around again! we try to comply to eslint rules anyway...

JKRhb: Okay, great :)

JKRhb:
> Question do you plan to tackle #523 in this PR or another PR? This would be very good also!

I originally intended to do so but I realized that doing do requires a bit more thought as the code that has been commented out is quite old :/ So I think it might be best to revisit this problem in another PR :)

packages/binding-coap/src/coap-client.ts
JKRhb: The old approach was to use the default value and print a debug message if the method is undefined or unknown. I am not sure if this approach or raising an error would better.

In any case, I think the solution we agree on should probably also be implemented in other bindings to have a consistent behavior, I guess? At least in the `binding-http`, package such a check does not seem to happen if I am not mistaken.

JKRhb: I now reintroduced checks with the default method as a fallback. Let me know if there should be an error thrown instead :)

JKRhb:
I now also added proper typing for the CoAP method names and corresponding assertions. However, I noticed that the PR now introduces some redundancy with the two `determineRequestMethod` methods. As soon as a new version of `node-coap` is released the method for the "regular" CoAP client will have to be changed anyway, so I guess it could stay as it is for now.

One side note: I used a couple of fixup commits in this PR which I would squash before this PR is merged :)

danielpeintner (approved):
LGTM

JKRhb:
> LGTM

Awesome :) I just squashed the commits so this PR would be ready to be merged.

PR #535, opened 2021-09-23
chore: add code coverage with nyc and codecov.io

### LINKED ISSUE #524: Add code coverage indicator
For the future it might be helpful to have an indicator for how much of each package is actually covered by tests. Using tools like `nyc` in the context of workspaces might take some time to figure out, however, so I would consider this a low priority issue for now.

JKRhb:
This PR resolves #524 by adding an npm script for code coverage determination using nyc.

It also lets the CI upload coverage results to codecov.io which gives helpful reports in PRs and also allows to include a coverage badge in the readme file. This would have to be activated at codecov.io to actually work, however.


JKRhb:
FYI: Some information on codecov.io is already available giving some insight in the current coverage status: https://codecov.io/gh/eclipse/thingweb.node-wot/tree/99a5c31046d3a8960179f56388a4c9ac5c257773

package-lock.json
```diff
@@ -4,7 +4,6 @@
     "requires": true,
     "packages": {
         "": {
-            "name": "thingweb.node-wot",
```
relu91, line 7: Minor problem, when I run `npm install` this line is added automatically by my npm, can we re-introduce it? Otherwise it will pop up again later in some other PR 😄

relu91 (approved):
Thank you! it works locally. I think it might be beneficial to have also `coverage:only` which runs `test:only` when for example you have already everything built and you want to check the coverage.

JKRhb:
> Thank you! it works locally. I think it might be beneficial to have also `coverage:only` which runs `test:only` when for example you have already everything built and you want to check the coverage.

Very good suggestion, I added `coverage:only` to the `scripts` section :)

package-lock.json
JKRhb: Sure! :) I am a bit clueless why my npm messes up the `package-lock.json`, though :/

JKRhb: I added it manually now – maybe you could update the `package-lock.json` file on the master branch once the PR is merged just to be safe? :)

relu91:
@JKRhb do you know why CI is failing? your changes shouldn't have mess with the current code-base. I'm wondering how did you install nyc?

JKRhb:
> @JKRhb do you know why CI is failing? your changes shouldn't have mess with the current code-base. I'm wondering how did you install nyc?

Hmm, I think I simply ran `npm install nyc --save-dev` :/ You make edits on my branch, right? Could you maybe regenerate `package-lock.json` there? Maybe this fixes the issue.

relu91:
It seems that now everything is in place, I'm running again the CI just to be sure about the install issues. Have you manually fixed the `package-lock`? 

Btw `npm install nyc --save-dev` seems good, even if I always install tools inside the packages themselves. I think since `nyc` works globally on the repo it totally makes sense to install as you did.

JKRhb:
> It seems that now everything is in place, I'm running again the CI just to be sure about the install issues. Have you manually fixed the `package-lock`?

Yeah, I added the line `"name": "thingweb.node-wot"` back in :) If I am not mistaken the issue with the build on Windows seems to be persistent, though. I'd assume it's probably best if the package-lock file was regenerated on `master` once this PR is merged to have a consistent (and working) lock file.

danielpeintner:
Note: when I build the project locally I also get always a diff with package-lock.
I know it is best practice to use it. Having said that, somehow it always causes (and caused) lots of issues and I think we are not alone

https://github.com/npm/npm/issues/20603

relu91:
> Note: when I build the project locally I also get always a diff with package-lock.

Building should not change the package-lock, I am a little bit puzzled. By the way, you shouldn't use `npm install` but `npm ci` when you want to create the node_modules folder. On the other hand, when you are adding or removing a dependency `npm install X` should work and it's ok if it is changing the package lock. 

What is strange about the CI error is that sometimes it does not find the binaries for serialport... very annoying.

> npm/npm#20603

Interesting reading! I think he has a point, but we had problems with floating dependencies like opc-ua that at every minor release broke every test. Let's see if the npm guys will solve this in the future, at least now we have only one package-lock file to maintain.

relu91:
What I did:
- pull your changes
- `git reset master -- package-lock.json`
- `npm install nyc@^15.1.0 --save-dev`
- `npm install @istanbuljs/nyc-config-typescript@^1.0.1 --save-dev`

It seems to me that the package-lock now looks more consistent ( it doesn't have changes in the seriaport dependencies for example). Are you sure that you haven't run `npm install` by mistake?

danielpeintner:
> It seems to me that the package-lock now looks more consistent ( it doesn't have changes in the seriaport dependencies for example). Are you sure that you haven't run `npm install` by mistake?

Mhh, I still use `npm install` (habit 🙈). Having said that, we still have lots of places mentioning `npm install`
e.g., https://github.com/eclipse/thingweb.node-wot/#windows

are they all ok by now?

relu91:
> e.g., https://github.com/eclipse/thingweb.node-wot/#windows

That should be fine, but I agree we need to document it in the readme, as you said people have the habit to use npm install (I also do this mistake from time to time). 

Having said that, I'm merging this, cause we got a green light from the CI.

JKRhb:
> Are you sure that you haven't run `npm install` by mistake?

I am afraid this might have been the case 😅 Now I know better for the future :) Thank you fixing this!

PR #537, opened 2021-09-27
fix(binding-opcua)  fix binding-opcua

erossignon:
- improve test
- refactor to 0.8.0 api
- improve typescript compliance (avoid using any)
- fix resource leaks
- properly terminate servr
- prevent too many subscriptions to be created.

### DIFF
--- a/packages/binding-opcua/.eslintrc.json
+++ b/packages/binding-opcua/.eslintrc.json
@@ -1,0 +1,7 @@
+{
+    "extends": "../../.eslintrc.json",
+    "parserOptions": {
+        "sourceType": "module",
+        "project": ["./tsconfig.json", "./tsconfig.eslint.json"]
+    }
+}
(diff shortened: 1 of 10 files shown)


relu91 (approved):
Thank you for your contribution to opc-ua implementation! About the stop method, I agree it should be async, it is something that we need to fix.

danielpeintner:
Unfortunately we do have test issues with OPCUA now, see

https://github.com/danielpeintner/thingweb.node-wot/runs/3732161234?check_suite_focus=true

@erossignon do you have an idea what is causing this issues?

erossignon:
> @erossignon do you have an idea what is causing these issues?
On some slow VM, the data change monitoring events are not guaranteed to be send by the server one after the other.
This causes some flaky behavior, I have created a PR to take this into account.

Issue #534, opened 2021-09-23
Track of ESLint Ignored Rules


fatadel:
This issue is created to keep track of ignored eslint rules with respective justifications.

1) **Flie/Line:** `https://github.com/eclipse/thingweb.node-wot/blob/master/packages/td-tools/src/thing-description.ts`
**Ignored Rule:** `no-use-before-define`
**Reason:** Many circular dependencies where interface/type/class A uses interface/type/class B in its definition and interface/type/class B uses interface/type/class A in its definition.

JKRhb:
If I see it correctly, we are targeting ES6 with the TS transpiler at the moment, right? Then I think we could actually disable the rule `no-use-before-define` completely which apparently is only relevant if a JavaScript version prior to ES6 is used at the transpile target (see https://eslint.org/docs/rules/no-use-before-define).

danielpeintner:
> This point has actually been resolved in the meantime, by the way :)

I have striked through it

danielpeintner:
> If I see it correctly, we are targeting ES6 with the TS transpiler at the moment, right? Then I think we could actually disable the rule `no-use-before-define` completely which apparently is only relevant if a JavaScript version prior to ES6 is used at the transpile target (see https://eslint.org/docs/rules/no-use-before-define).

ES6 dates back to ECMAScript 2015.
Hence I think this is okay.

@relu91 any opinion? 

BTW, I noticed that in [.prettierrc.json](https://github.com/eclipse/thingweb.node-wot/blob/59fe3264bb4cfa19a3c8cddb0f1c68b4685a4a47/.prettierrc.json) we still use es5. Shall we change that too?
see https://github.com/eclipse/thingweb.node-wot/search?q=ES5

relu91:
> ES6 dates back to ECMAScript 2015.
> Hence I think this is okay.

Good point, but I think is still a good practice to define stuff before using them (i.e. classes types etc. ). So I would be ok for leaving the rule as it is, but I would not stop you if you think differently :) 

> BTW, I noticed that in [.prettierrc.json](https://github.com/eclipse/thingweb.node-wot/blob/59fe3264bb4cfa19a3c8cddb0f1c68b4685a4a47/.prettierrc.json) we still use es5. Shall we change that too?
see https://github.com/eclipse/thingweb.node-wot/search?q=ES5

Yeah, I think we should.

JKRhb:
> > ES6 dates back to ECMAScript 2015.
[...]

PR #538, opened 2021-09-27
feat(coap): update type definitions

JKRhb:
This PR will update the `binding-coap` package with typescript declarations from `node-coap` (see https://github.com/mcollina/node-coap/pull/278) once a new version is released. It includes my local adjustments used for checking that the type declarations are actually usable. There are a small number of `any`s which still have to be fixed but other than that a large amount of code will be covered now with proper typing.

I also noticed ~~two bugs~~ one minor issue in the `coap` package due to the new type declarations for which I will open a separate PR for an immediate fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "workspaces": [
                 "./packages/td-tools",
                 "./packages/core",
-                "./packages/binding-!(firestore-browser-bundle|mbus|firestore)",
+                "./packages/binding-!(firestore-browser-bundle|firestore)",
                 "./packages/cli",
                 "./packages/examples"
             ],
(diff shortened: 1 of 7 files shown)


JKRhb:
Made some progress, but there is currently still an issue with the "share socket" tests.

JKRhb:
This PR is now ready for review :) I included a commit from @relu91's PR #602 which is why it should probably be merged after it (and after a rebase).

With the merge of this PR, it will be possible to remove the CoAP entry in #534.

relu91:
> This PR is now ready for review :) I included a commit from @relu91's PR #602 which is why it should probably be merged after it (and after a rebase).

Thank you really much! do recommend merging #602 AFTER this PR or the other way around??

relu91 (approved):
Typing wise looks good! 👍🏻 ready to be merged :)

JKRhb:
> do recommend merging #602 AFTER this PR or the other way around??

Initially I thought we might merge after #602 but we could probably also do it the other way around. I think a couple of problems with the `requestUri` that are dealt with in #602 should already be solved by this PR

danielpeintner (approved):
I let you decide which PR should be merged first

relu91:
No conflicts after merging #602 :)

PR #539, opened 2021-09-27
fix(coap): use string type for option name

JKRhb:
This is a minor fix for the `coap` package replacing the wrong type  `number` of `cov:optionName` with `string`.


relu91 (approved):
Looks good, I'll merge right away since is a minor fix.

PR #541, opened 2021-09-28
refactor: align with eslint rules #3

fatadel:
Next iteration of eslint alignments (previous one - https://github.com/eclipse/thingweb.node-wot/pull/520).


danielpeintner (changes requested):
The following eslint message that belongs to a touched file could go with this change

```
Check failure on line 108 in packages/binding-coap/src/coaps-client.ts
'requestUri' is never reassigned. Use 'const' instead
```

danielpeintner:
Note: `npm run format` is missing

fatadel:
> The following eslint message that belongs to a touched file could go with this change
> 
> ```
> Check failure on line 108 in packages/binding-coap/src/coaps-client.ts
> 'requestUri' is never reassigned. Use 'const' instead
> ```

Done ✅

fatadel:
> Note: `npm run format` is missing

Done ✅

I've also removed coap import there because, apparently, it turned to be unused after merging master branch.

packages/binding-coap/src/coaps-client.ts
```diff
@@ -109,7 +109,7 @@ export default class CoapsClient implements ProtocolClient {
             coaps.setSecurityParams(requestUri.hostname, this.authorization);
 
             coaps
-                .observe(form.href, "GET", next)
+                .observe(form.href, "get", next)
```
danielpeintner, line 112: At the moment it is somewhat confusing.

Some methods in the same file are CAPITALIIZED and lowerCased afterwards, some are lowercase only.
Not an issue of your update but I general concern I have.

fatadel: @danielpeintner Agree, I had the same feeling.

danielpeintner (approved):
LGTM

packages/binding-coap/src/coaps-client.ts
```diff
@@ -105,11 +105,11 @@ export default class CoapsClient implements ProtocolClient {
         complete?: () => void
     ): Promise<Subscription> {
         return new Promise<Subscription>((resolve, reject) => {
-            let requestUri = url.parse(form.href.replace(/$coaps/, "https"));
+            const requestUri = url.parse(form.href.replace(/$coaps/, "https"));
```
relu91, line 108: Can't we solve this? it should be siply `url.parse` => `new Url`. Am I missing something?

relu91 (changes requested):
small detail then we merge!

packages/binding-coap/src/coaps-client.ts
fatadel: The URL object doesn't have `query` property. Sometimes we are using it and sometimes not. To be on the safe side, I've decided to leave it as it is (at least for now).

JKRhb: I ran into a similar problem in `node-coap` where I added a couple of dedicated function for retrieving the fields from the `URL` object: https://github.com/mcollina/node-coap/commit/71a90787b4960ebd36f62401338e81b057a0c58f Instead of `query` you can use the `search` property by removing the leading `?`.

This is obviously not very elegant, though. Maybe using something like [uri-js](https://www.npmjs.com/package/urijs) could be a better solution instead.

fatadel: @JKRhb Thanks for your comment!
@relu91 @danielpeintner Which way to go?

danielpeintner: Mhh, no strong opinion either way.

@fatadel I let you to decide to work on this in this PR or create an issue to work on it in a follow-up PR.
@relu91 any preference?

relu91: The point of this PR is improving es-lint errors. So I would go with the constructor and fix it here. If we are using the query parameter @JKRhb propose a good algorithm. However, we might keep track of IPv6 issues.

fatadel: Done ✅

relu91 (approved):

danielpeintner:
Thanks @fatadel

PR #548, opened 2021-09-30
refactor: use /properties instead of /all/properties

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/505


danielpeintner:
Note: the diff is very minor if you hide whitespaces since it just shuffles around code.

see  https://github.com/eclipse/thingweb.node-wot/pull/548/files?diff=split&w=1

relu91 (approved):
Looks good, I wonder why there were so many white spacing changes. is it prettier?

danielpeintner:
> Looks good, I wonder why there were so many white spacing changes. is it prettier?

No, not really... there is one outer `if` which changes each following line...
GH default diff does not do a good job here.. my local tools does not show them ;-)

PR #550, opened 2021-09-30
Cleanup td definitions

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/529

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -26,6 +26,7 @@ import * as url from "url";
 import { AddressInfo } from "net";
 
 import * as TD from "@node-wot/td-tools";
+import * as TDT from "wot-thing-description-types";
 import Servient, {
     ProtocolServer,
     ContentSerdes,
@@ -127,7 +128,7 @@ export default class HttpServer implements ProtocolServer {
                     break;
                 case "oauth2":
                     this.httpSecurityScheme = "OAuth";
-                    const oAuthConfig = config.security as OAuth2ServerConfig;
+                    const oAuthConfig = config.security as any as OAuth2ServerConfig;
                     this.validOAuthClients = new RegExp(oAuthConfig.allowedClients ?? ".*");
                     this.oAuthValidator = createValidator(oAuthConfig.method);
                     break;
@@ -311,9 +312,9 @@ export default class HttpServer implements ProtocolServer {
                     ];
                 }
                 if (!thing.forms) {
-                    thing.forms = [];
+                    thing.forms = undefined; // [];
                 }
-                thing.forms.push(form);
+                thing.forms.push(form as TDT.FormElementRoot);
             }
 
             for (const propertyName in thing.properties) {
@@ -430,7 +431,11 @@ export default class HttpServer implements ProtocolServer {
                 const oAuthScheme = thing.securityDefinitions[thing.security[0] as string] as OAuth2SecurityScheme;
 
                 // TODO: Support security schemes defined at affordance level
-                const scopes = oAuthScheme.scopes ?? [];
+                const scopes = oAuthScheme.scopes
+                    ? typeof oAuthScheme.scopes === "string"
+                        ? [oAuthScheme.scopes]
+                        : oAuthScheme.scopes
+                    : []; // oAuthScheme.scopes ?? [];
                 let valid = false;
 
                 try {
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -312,7 +312,7 @@ export default class HttpServer implements ProtocolServer {
                     ];
                 }
                 if (!thing.forms) {
-                    thing.forms = undefined; // [];
+                    thing.forms = new Array() as [TDT.FormElementRoot];
                 }
                 thing.forms.push(form as TDT.FormElementRoot);
             }
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -26,7 +26,7 @@ import * as url from "url";
 import { AddressInfo } from "net";
 
 import * as TD from "@node-wot/td-tools";
-import * as TDT from "wot-thing-description-types";
+import { FormElementRoot } from "wot-thing-description-types";
 import Servient, {
     ProtocolServer,
     ContentSerdes,
@@ -312,9 +312,9 @@ export default class HttpServer implements ProtocolServer {
                     ];
                 }
                 if (!thing.forms) {
-                    thing.forms = new Array() as [TDT.FormElementRoot];
+                    thing.forms = new Array() as [FormElementRoot];
                 }
-                thing.forms.push(form as TDT.FormElementRoot);
+                thing.forms.push(form as FormElementRoot);
             }
 
             for (const propertyName in thing.properties) {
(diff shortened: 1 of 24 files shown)


danielpeintner:
Note: there are related issues, see https://github.com/w3c/wot-thing-description/issues/1235

relu91:
Hi Daniel! is great that you are starting working on this, I am wondering do we have a particular reason to keep using `import * as TDT ...` instead of a more precise `import {SecurityScheme} ...`. The first one feels unnecessary verbose in this situation: https://github.com/eclipse/thingweb.node-wot/pull/550/files#diff-3f331255e93fb9e3e714530c8787b51ab7615c6cb3ad33a16d1e753d5e5833a7

Do we still have conflicting terms? i.e. TD types defined in node-wot?

packages/binding-http/src/http-server.ts
```diff
@@ -311,9 +312,9 @@ export default class HttpServer implements ProtocolServer {
                     ];
                 }
                 if (!thing.forms) {
-                    thing.forms = [];
+                    thing.forms = new Array() as [FormElementRoot];
```
erossignon, line 315: alternatively : 
```` 
= [] as FormElementRoot[];
```` 
or
````
= < FormElementRoot[]>[];
````

File: packages/binding-http/src/http-server.ts [hunk 2]
```diff
@@ -311,9 +312,9 @@ export default class HttpServer implements ProtocolServer {
...
                 }
-                thing.forms.push(form);
+                thing.forms.push(form as FormElementRoot);
```
erossignon, line 317: why is TD.Form not compatible with FormElementRoot already ?

packages/binding-http/test/http-client-oauth-tests.ts
```diff
@@ -75,26 +76,26 @@ class HttpClientOAuthTest {
...
-        this.client.setSecurity([scheme], { clientId: "thom", clientSecret: "nightworld" });
+        this.client.setSecurity([scheme as SecurityScheme], { clientId: "thom", clientSecret: "nightworld" });
         return this.client.readResource({
             href: "https://localhost:3000/resource",
         });
     }
 
     @test async "should authorize client with resource owener flow"() {
-        const scheme: OAuth2SecurityScheme = {
+        const scheme: TD.OAuth2SecurityScheme = {
```
erossignon, line 92: Could we make TD.OAuth2SecurityScheme extending SecurityScheme to avoid casting ?

danielpeintner: w.r.t. security schemes I still want to wait till a solution for https://github.com/w3c/wot-thing-description/issues/1235 has been landed. At the moment we need to redefine all security schemes. I rather re-use the types defines in TD schema.

packages/binding-http/src/http-server.ts
danielpeintner: This is a good question and and somewhat also complicated given that we re-use the types from [TD JSON Schema](https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json).

There we have  *different* forms for root, properties, actions and events.

https://github.com/w3c/wot-thing-description/blob/0600e4384c129bf1dc1fa4fa1a34074d6b684afa/validation/td-json-schema-validation.json#L254

https://github.com/w3c/wot-thing-description/blob/0600e4384c129bf1dc1fa4fa1a34074d6b684afa/validation/td-json-schema-validation.json#L437

..

The question is whether we should have *different* `form`s also.

packages/binding-http/src/http-server.ts
danielpeintner: neither  `[] as FormElementRoot[]` nor `< FormElementRoot[]>[];` is working

`[] as [FormElementRoot]` does not work either.
Hence  I suggest to keep it as is till we have decided what to do with proper `form` typing.

packages/binding-http/src/http-server.ts
danielpeintner: see https://github.com/eclipse/thingweb.node-wot/issues/561

packages/binding-http/test/http-client-oauth-tests.ts
relu91: btw w3c/wot-thing-description#1235 is not resolved 🎉 . Just an heads up 😃

packages/binding-http/src/http-server.ts
```diff
@@ -430,7 +431,11 @@ export default class HttpServer implements ProtocolServer {
                 const oAuthScheme = thing.securityDefinitions[thing.security[0] as string] as OAuth2SecurityScheme;
 
                 // TODO: Support security schemes defined at affordance level
-                const scopes = oAuthScheme.scopes ?? [];
+                const scopes = oAuthScheme.scopes
```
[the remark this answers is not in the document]
relu91: Actually is not nested `?` operator but it is a TS operator for [Nullish Coalescing](https://mariusschulz.com/blog/nullish-coalescing-the-operator-in-typescript). Is totally fine in TS code. I guess @danielpeintner removed it because oAuthScheme.scope will be never `null | undefined` at the point. right?

danielpeintner:
I think it is better to redo this PR once all parts are fixed in the JSON schema.

* [x] Each schema has now its own definition (https://github.com/w3c/wot-thing-description/pull/1250, Great!)
* [ ] Different FormTypes, see https://github.com/eclipse/thingweb.node-wot/issues/561
   -  node-wot uses *one* Form class 
   - [JSON schema](https://github.com/w3c/wot-thing-description/blob/main/validation/tm-json-schema-validation.json) has `form_element_property`, `form_element_event`, `form_element_action `and `form_element_root`
   - I wonder whether it is possible to have a *generic* parent like `form_element` (others would be specifications of that) that could be defined in the JSON schema and which we could be using. Shall we take a look at the JSON schema and bring that to the TD task force?

@relu91 @egekorkan others? Do you have an opinion on that?

egekorkan:
regarding the second part I will have a look again. I had tried it in the beginning but did not manage. I think simply doing allOf with the specific schema and the generic schema should do it

relu91:
I think we can achieve something good by looking at what we've done with `securityScheme`. It should work the same for `Form`, right?

https://github.com/w3c/wot-thing-description/blob/7343d5ab9d2db1b8016c63a58753337a6481114b/validation/tm-json-schema-validation.json#L1519

egekorkan:
I am not sure if what I did at https://github.com/w3c/wot-thing-description/pull/1319 really fixes the situation. Is the goal that each affordance has always one form element? If that is the case, I don't know how to do that I think.

danielpeintner:
The question is "what" the tool does that converts JSON Schema -> TypeScript defintitions

The _goal_ is that each form (root, property, action and event) can be assigned to an instanceof base ....

Maybe https://github.com/w3c/wot-thing-description/pull/1319 is not doing that... and it should be slightly re-worked...
see comments https://github.com/w3c/wot-thing-description/pull/1319#pullrequestreview-827641477

@relu91 what do you think?

relu91:
> @relu91 what do you think? 

I've just left a review there 👍🏻

danielpeintner:
redo after https://github.com/w3c/wot-scripting-api/pull/372

danielpeintner:
replaced by https://github.com/eclipse/thingweb.node-wot/pull/750

PR #551, opened 2021-10-04
chore: remove oracle and fujitsu binding

danielpeintner:
see missing maintainers in https://github.com/eclipse/thingweb.node-wot/issues/508


packages/examples/src/demo-servients/bridge-cli.ts
```diff
@@ -128,7 +127,7 @@ If the file 'bridge-servient.conf.json' exists, that configuration is applied.
 Options:
   -v, --version           display node-wot version
   -f, --configfile=file   load configuration from specified file
-  -p, --password=secret   pass the password for the Oracle trust store
+  -p, --password=secret   pass the password for the trust store
```
relu91, line 130: It feels that this option was pretty related to the Oracle implementation, should we really keep it?

packages/examples/src/proxy-scripts/festo-live.ts
```diff
@@ -110,7 +110,7 @@ Promise.all(fetchArray).then(async (tdArray: ThingDescription[]) => {
     }, 5000);
 });
 
-// exposed Thing toward Oracle IoT Cloud Service
+// exposed Thing toward IoT Cloud Service
 WoT.produce({
```
relu91, line 114: Will this script still work?

relu91 (approved):
I have just two super small concerns. Other than that it looks ok!

egekorkan:
Another thing: I think they should be also removed but it would be better if they are mentioned in the readme as "here are some protocols that are orphaned due to lack of maintainers" and then link to a 0.7.x release.

packages/examples/src/demo-servients/bridge-cli.ts
danielpeintner: @wiresio ?

packages/examples/src/proxy-scripts/festo-live.ts
danielpeintner: @wiresio ?

wiresio: The mentioned files are related with `binding-oracle` and should be moved there. Maintenance then depends on the maintainer of the binding.

sebastiankb:
I think, it make sense to write an email and share the link to this discussion.

danielpeintner:
I messed up this PR, replaced by https://github.com/eclipse/thingweb.node-wot/pull/593

PR #554, opened 2021-10-06
chore: use typescript composite build for speed

erossignon:
I would like to propose this enhancement that uses typescript composite project build (https://www.typescriptlang.org/tsconfig#composite https://www.typescriptlang.org/docs/handbook/project-references.html) to speed up. I am using this technique in node-opcua.
The main benefit is a speedup in compilation time, once the project has been built once. This allows typescript to compile the entire module in one command rather than visiting each module and running the tsc command.

In brief:
 - the entire project can be built with a single `tsc -b` command in the root folder.
 - each module can be recompile (including their dependency) with  `tsc -b`  run from the module folder.
 -we have to maintain dependencies of modules in the tsconfig.json file

### DIFF
--- a/packages/binding-opcua/package.json
+++ b/packages/binding-opcua/package.json
@@ -29,7 +29,7 @@
         "eslint-plugin-promise": "^5.1.0",
         "eslint-plugin-unused-imports": "^1.1.4",
         "mocha": "^9.1.1",
-        "node-opcua": "^2.55.0",
+        "node-opcua": "~2.55.0",
         "prettier": "^2.3.2",
         "ts-node": "10.1.0",
         "typescript": "4.4.3",


packages/cli/tsconfig.json
```diff
@@ -3,5 +3,14 @@
...
+    "include": ["src/**/*"],
+    "references": [
+        { "path": "../td-tools" },
+        { "path": "../core" },
+        { "path": "../binding-http" },
+        { "path": "../binding-file" },
+        { "path": "../binding-websockets" },
+        { "path": "../binding-coap" },
+        { "path": "../binding-mqtt" }
+    ]
```
danielpeintner, line 15: May I ask why this is necessary. Not just for this but also for other tsconfig files

erossignon: Yes, this is necessary, This is how the typescript transpiler knows how to cascade to the hierarchy of dependent projects.
The top level tsconfig.json should contain reference ptah to all major projects.
Inside a binding-xxxx project you would only have the path to the dependant project (namely, I guess, td-tools and core only).

erossignon: buid is failing due to master failing.  having (ff) #519 merged to master would help here.

packages/binding-http/src/http-server.ts
```diff
@@ -124,10 +121,12 @@ export default class HttpServer implements ProtocolServer {
...
-                    this.httpSecurityScheme = "OAuth";
-                    const oAuthConfig = config.security as OAuth2ServerConfig;
-                    this.validOAuthClients = new RegExp(oAuthConfig.allowedClients ?? ".*");
-                    this.oAuthValidator = createValidator(oAuthConfig.method);
+                    {
+                        this.httpSecurityScheme = "OAuth";
+                        const oAuthConfig = config.security as OAuth2ServerConfig;
+                        this.validOAuthClients = new RegExp(oAuthConfig.allowedClients ?? ".*");
+                        this.oAuthValidator = createValidator(oAuthConfig.method);
+                    }
```
danielpeintner, line 129: Unnecessary change. At least I can not spot the reason

packages/binding-http/src/http-server.ts
```diff
@@ -424,7 +422,7 @@ export default class HttpServer implements ProtocolServer {
             }
             case "Digest":
                 return false;
-            case "OAuth":
+            case "OAuth": {
```
danielpeintner, line 425: Similar, this seems a matter of taste, right? No need to encapsulate..

packages/binding-http/src/http-server.ts
erossignon: This was to fix a ESLint error:
image

packages/binding-http/src/http-server.ts
erossignon: eslint recommends this in the default recommanded rules

.vscode/settings.json
```diff
@@ -39,5 +39,20 @@
...
+    "cSpell.words": [
+        "affordances",
+        "alparser",
+        "bauth",
+        "blockwise",
+        "Coap",
+        "Coaps",
+        "mcollina",
+        "nosec",
+        "observeproperty",
```
erossignon, line 52: I need to add "unobserverproperty" as well to avoid accidental replacement of unobserverproperty by observerproperty by the spelling checker !

packages/binding-coap/src/coap-server.ts
```diff
@@ -140,7 +136,7 @@ export default class CoapServer implements ProtocolServer {
                                 form.op = [];
                             }
                             form.op.push("observeproperty");
-                            form.op.push("unobserveproperty");
+                            form.op.push("observeproperty");
```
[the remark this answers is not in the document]
erossignon: I found the root cause : I have added "observeproperty" to the spell checker dictionary and it offered me to replace unobserveproperty with observeproperty. Solution: I have also added "unobserveproperty" to the spell check dic.

relu91 (commented):
I still have to properly review this PR. What I find difficult is to navigate between the changes because they are mixed with other eslint fixes or compilation fixes. Plus the example folder is kinda special in node-wot, it was meant to be uncorrelated with the main project. The plan was to auto-generate most of the js scripts that you're seeing there. 

In short, I'd kindly ask you to be more focused on the next PRs thank you!

tsconfig.json
```diff
@@ -1,17 +1,50 @@
...
-        "declaration": true
-    }
+        "declaration": true,
+        "esModuleInterop": true,
+        "strictFunctionTypes": false,
+        "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
+        "resolveJsonModule": false,
+        "noStrictGenericChecks": false,
+        "noImplicitThis": true,
+        "skipDefaultLibCheck": false,
```
relu91, line 26: I don't remember exactly where this came from but I think is something particular that should stay on the specific package.

packages/binding-coap/package.json
```diff
@@ -30,11 +29,13 @@
...
         "mocha": "^9.1.1",
         "prettier": "^2.3.2",
-        "ts-node": "10.1.0",
-        "typescript": "4.1.3",
+        "ts-node": "^10.2.1",
+        "typescript": "^4.4.3",
         "typescript-standard": "^0.3.36"
     },
     "dependencies": {
+        "@types/node": "16.4.13",
```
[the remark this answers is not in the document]
erossignon: yes, it is also duplicated.

packages/binding-coap/test.tsconfig.json
```diff
@@ -1,9 +0,0 @@
-{
-    "extends": "../../tsconfig.json",
-    "compilerOptions": {
-        "outDir": "testDist",
-        "resolveJsonModule": true,
-        "esModuleInterop": true
-    },
-    "include": ["test/**/*"]
-}
```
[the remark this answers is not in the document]
erossignon: let me fix this.

packages/binding-http/package.json
```diff
@@ -16,7 +16,7 @@
     "browser": "dist/http-browser.js",
     "devDependencies": {
         "@testdeck/mocha": "^0.1.2",
-        "@types/basic-auth": "1.1.3",
+        "@types/basic-auth": "^1.1.3",
```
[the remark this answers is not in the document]
erossignon: let me revert this.

tsconfig.json
```diff
@@ -1,17 +1,50 @@
...
     "compilerOptions": {
         "target": "es6",
-        "lib": ["es2015", "dom"],
+        "lib": ["dom"],
+        "skipLibCheck": false,
         "module": "commonjs",
         "outDir": "dist",
-        "alwaysStrict": true,
+        "alwaysStrict": false,
+        "strict": false,
```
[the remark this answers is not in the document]
erossignon: yes, correct.

tsconfig.json
erossignon: This flags uses the default value from typescript. we can remove them here, or keep them to make thing more explicit. 
Let me know what works best for you.

packages/core/src/protocol-helpers.ts
```diff
@@ -17,9 +17,18 @@ import * as TD from "@node-wot/td-tools";
...
 
+export interface IManagedStream2 {
+    nodeStream: Readable;
+    wotStream: ReadableStream;
+}
+
+export interface IManagedStream {
+    nodeStream: NodeJS.ReadableStream;
+    wotStream: ReadableStream;
+}
```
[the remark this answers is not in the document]
erossignon: I agreed, Here I try to get the existing code to compile under the new common tsconfig. rule. Let do this on a separate PR.

tsconfig.json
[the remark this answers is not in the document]
erossignon: fixed in 0a9ca0a

packages/binding-opcua/package.json
```diff
@@ -30,24 +29,24 @@
         "eslint-plugin-promise": "^5.1.0",
         "eslint-plugin-unused-imports": "^1.1.4",
         "mocha": "^9.1.1",
-        "node-opcua": "~2.45.0",
+        "node-opcua": "^2.55.0",
```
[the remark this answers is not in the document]
erossignon: 90a4cf746fe0630caec490736a6e9ec4bc61e351

examples/servients/coap-servient/package.json
```diff
@@ -20,7 +20,7 @@
     "devDependencies": {
         "wot-typescript-definitions": "^0.8.0-SNAPSHOT.18",
         "@types/node": "16.4.13",
-        "typescript": "4.1.3"
+        "typescript": "4.4.3"
```
erossignon, line 23: 4.4.3 is in fact now need

tsconfig.json
relu91: I prefer having only configurations that differ from the default, what do you think @danielpeintner ?

examples/servients/coap-servient/package.json
relu91: Good, next time please state this in the PR description. It's our own limitation cause we don't provide a PR template, but I think that changes to dependences should be always justified. See #559.

packages/core/src/protocol-helpers.ts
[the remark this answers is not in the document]
erossignon: This technical refactoring is an attempt to fix this
```diff
-    stream: NodeJS.ReadableStream | { nodeStream: Readable; wotStream: ReadableStream }
```
with
````
+    stream: NodeJS.ReadableStream | IManagedStream
```
to make the code more readable;
As doing this, it became apparent that we have a  hidden bug because `class extends Base` do not follow strictly this signature.
So I have created the IManagedStream2 to better highlight the fact that we need to fix the root cause. but not in the context of this PR.


 ```` diff
+export interface IManagedStream2 {
+    nodeStream: Readable;
+    wotStream: ReadableStream;
+}
+
+export interface IManagedStream {
+    nodeStream: NodeJS.ReadableStream;
+    wotStream: ReadableStream;
+}
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/ban-types
 function ManagedStream<TBase extends new (...args: any[]) => {}>(Base: TBase) {
-    return class extends Base {
+    return class extends Base implements IManagedStream {
         _nodeStream: NodeJS.ReadableStream;
         _wotStream: ReadableStream;
         set nodeStream(nodeStream: NodeJS.ReadableStream) {
@@ -43,11 +52,11 @@ function ManagedStream<TBase extends new (...args: any[]) => {}>(Base: TBase) {
 const ManagedReadable = ManagedStream(Readable);
 const ManagedReadableStream = ManagedStream(PolyfillStream);

-function isManagedReadable(obj: unknown): obj is { nodeStream: Readable; wotStream: ReadableStream } {
+function isManagedReadable(obj: unknown): obj is IManagedStream2 {
     return obj instanceof ManagedReadable;
 }

-function isManagedReadableStream(obj: unknown): obj is { nodeStream: Readable; wotStream: ReadableStream } {
+function isManagedReadableStream(obj: unknown): obj is IManagedStream2 {
     return obj instanceof ManagedReadableStream;
 }
 export default class ProtocolHelpers {
@@ -198,9 +207,7 @@ export default class ProtocolHelpers {
         return undefined; // not found
     }

-    public static toWoTStream(
-        stream: NodeJS.ReadableStream | { nodeStream: Readable; wotStream: ReadableStream }
-    ): ReadableStream | PolyfillStream {
+    public static toWoTStream(stream: NodeJS.ReadableStream | IManagedStream2): ReadableStream | PolyfillStream {
         // TODO USE CLASSES
         if (isManagedReadable(stream)) {
             return stream.wotStream;
@@ -222,10 +229,8 @@ export default class ProtocolHelpers {
         return result;
     }

-    public static toNodeStream(
-        stream: ReadableStream | PolyfillStream | { nodeStream: Readable; wotStream: ReadableStream }
-    ): Readable {
-        // TODO: use proper clases
+    public static toNodeStream(stream: ReadableStream | PolyfillStream | IManagedStream2): Readable {
+        // TODO: use proper classes
         if (isManagedReadableStream(stream)) {
             return stream.nodeStream;
         }
@@ -247,7 +252,8 @@ export default class ProtocolHelpers {
                 stream.cancel(error);
             },
         });
-        result.wotStream = stream;
+        result.wotStream = stream as any;
+
         return result;
     }
```

File: .vscode/settings.json [hunk 14]
```diff
@@ -15,5 +15,21 @@
...
+        "unobserveproperty",
+        "readproperty",
+        "servients",
+        "subprotocol",
+        "writeproperty"
+    ]
```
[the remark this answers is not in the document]
erossignon: offending commit has now been removed .

tsconfig.json
danielpeintner: I agree with @relu91. I suggest we use flags only when our need does not not match the default.

danielpeintner (approved):

PR #557, opened 2021-10-08
docs: fix copyrights based on git log

danielpeintner:


relu91:
> I have spotted copyright not handled by the script in `/.vscode/settings.json`

Good point, but I think it is from an old configuration for a vs code extension. @danielpeintner should we keep it? do you recall the extension name? if you do we can add a configuration to let vs code warn users that we suggest such extension when developing node-wot.

relu91 (approved):

danielpeintner:
> > I have spotted copyright not handled by the script in `/.vscode/settings.json`
> 
> Good point, but I think it is from an old configuration for a vs code extension. @danielpeintner should we keep it? do you recall the extension name? if you do we can add a configuration to let vs code warn users that we suggest such extension when developing node-wot.

Thanks for spotting!

This is somewhat difficult to deal with because for new files it should contain the current year only.
`Copyright (c) 2021 Contributors to the Eclipse Foundation`

I agree with @relu91 I am not sure if it is still in use. I don't think so.
The extension is most likely https://marketplace.visualstudio.com/items?itemName=doi.fileheadercomment

Anyhow, I think we should remove it. So did I in my last commit.
If anyone feels strongly we can still bring it back.

PR #562, opened 2021-10-12
feat: use promise for stop/start in ProtocolClient

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/490


danielpeintner:
Note: see question https://github.com/eclipse/thingweb.node-wot/issues/490#issuecomment-940793244

Issue #569, opened 2021-10-13
Promise async code style


danielpeintner:
_input from @erossignon_ 

We have different code styles used throughout the code for promises.

```JS
method(...): Promise<...> {
  return new Promise<...>((resolve, reject) => {
    // actual code
  }
}
```

or

```JS
async method(...): Promise<...> {
    // actual code
}
```

Shall we leave it as is or go with one or the other?
The 2nd variant seems cleaner and less verbose. I don't believe there is any downside...

PR #573, opened 2021-10-13
chore(core) use async pattern #569

erossignon:


packages/core/src/consumed-thing.ts
```diff
@@ -333,284 +315,227 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
...
+                (err) => {
+                    reject(err);
+                },
+                // complete
+                () => {
                     resolve();
                 }
-            }
+            );
+            // to do : what shall we do with subscriptionPromise ?
```
relu91, line 453: Yeah, this part this a minor change. I can revisit when merging so don't worry.

PR #579, opened 2021-10-13
fix(binding-http): flaky http tests

erossignon:


erossignon:
- avoiding port number above 60000 seems to fix the port access issue.
- i found that some after() test method may return before the async operation they contain is terminated

packages/binding-http/test/http-client-test.ts
```diff
@@ -443,11 +440,11 @@ class HttpClientTest2 {
         let eventSpy = chai.spy(async function (data: any) {
             eventSpy.should.have.been.called.once;
             subscribeSpy.should.have.been.called.once;
-            done();
             server.close();
+            done();
         });
 
-        server.listen(60605, "0.0.0.0");
+        server.listen(34000, "0.0.0.0");
```
relu91, line 447: > Different operating systems (OS) use different port ranges for ephemeral ports. Many Linux versions use port range 32768-61000, while Windows versions (until XP) use 1025-5000, by default.

Aren't we again in the possible range for ephemeral ports? shouldn't it be something below 32000?

packages/binding-http/test/http-client-oauth-tests.ts
```diff
@@ -64,16 +64,16 @@ class HttpClientOAuthTest {
         });
     }
 
-    static after() {
-        return promisify(this.server.close.bind(this.server))();
+    static async after() {
+        return await promisify(this.server.close.bind(this.server))();
```
relu91, line 68: This line still uses `promisify` where the others use the `new Promise` pattern. I'll open an issue to keep track of this we should use one pattern consistently (personally I found `promisify` less verbose and more explicit)

danielpeintner (approved):
LGTM

Let's hope the issues are gone.
I did not check all failing tests but to me it seems to happen on Mac

Issue #586, opened 2021-10-19
Resolve remaining eslint errors/warnings


danielpeintner:
`npm run lint` still shows errors/warnings which we should fix:
Note: `npm run lint` can be run in a dedicated package only to the see the according infos.

* [x] `td-tools`, resolved by https://github.com/eclipse/thingweb.node-wot/pull/581/
* [x] `core`, initially resolved by https://github.com/eclipse/thingweb.node-wot/pull/582/ but I noticed recent updates broke it again, see https://github.com/eclipse/thingweb.node-wot/pull/585#issuecomment-946428271
* [x] `binding-coap`, @JKRhb, https://github.com/eclipse/thingweb.node-wot/pull/588
* [x] `binding-file`, @danielpeintner, see https://github.com/eclipse/thingweb.node-wot/pull/587
* [x] `binding-firestore-browser-bundle` and `binding-firestore`, see https://github.com/eclipse/thingweb.node-wot/pull/808
* [x] `binding-http`, @relu91 ?
* [x] `binding-mbus`, @LennyHEVS @Chxresubles @danielpeintner https://github.com/eclipse/thingweb.node-wot/pull/594
* [x] `binding-modbus`, @relu91 ?
* [x] `binding-mqtt`, @danielpeintner, https://github.com/eclipse/thingweb.node-wot/pull/592
* [x] `binding-netconf`, disabled warnings by https://github.com/eclipse/thingweb.node-wot/pull/818
* [x] `binding-opcua`, @erossignon, https://github.com/eclipse/thingweb.node-wot/pull/591
* [x] `binding-websockets`, @miguelrk, https://github.com/eclipse/thingweb.node-wot/pull/590
* [x] `browser-bundle`, @danielpeintner, https://github.com/eclipse/thingweb.node-wot/pull/597
* [x] `cli`, https://github.com/eclipse/thingweb.node-wot/pull/589
* [x] `examples`, ?, Note: `npm run lint` causes Oops! Something went wrong! :( , see https://github.com/eclipse/thingweb.node-wot/issues/595

relu91:
I've checked the boxes to reflect the latest changes. There are still some warnings in `netconf` and `firestore` is still excluded from eslint checking.

danielpeintner:
> * [x] `binding-firestore-browser-bundle` and `binding-firestore`

I reflected the latest updates. Having said that, `netconf` should have been the only package which needs some care.
Anyone willing to take the task? 

[...]

PR #588, opened 2021-10-19
refactor(coap): fix eslint issues

JKRhb:
This PR fixes a number of eslint issues in the `binding-coap` package (also see #586). Due to the fact that are no typescript bindings for `node-coap` yet some `any` typings couldn't be removed yet. However, with the next version of `node-coap` and the merge of #538 this will hopefully also be resolved eventually.

~~I opened this PR as a draft for now due to some tests that failed locally.~~ The issues should be fixed now :)

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -40,7 +40,9 @@ import coap = require("coap");
 
 export default class CoapClient implements ProtocolClient {
     // FIXME coap Agent closes socket when no messages in flight -> new socket with every request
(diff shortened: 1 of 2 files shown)


JKRhb:
> In my understanding, once we have #538 ready to go, we will have also typings for those any variables. 

Exactly, I hope that the new version of node-coap is going to be ready soon (unfortunately, the release process is a bit delayed at the moment). Therefore, I would tend to keep the warnings as temporary reminders as well but I can also add the ignore statements if you prefer them, @danielpeintner.

PR #590, opened 2021-10-19
fix(binding-websockets): eslint errors/warnings

miguelrk:
This commit resolves remaining errors and warnings for the `binding-websockets` package (see #586). Unimplemented code is simply marked with 'TODO: implement', to avoid throwing lint warnings, yet implementation is still missing. Use of 'any' is replaced with 'unknown' while waiting for official types or specific implementation.

NOTE: This PR resolves all lint errors and warnings, several methods are yet to be implemented for this binding and marked with `TODO: implement`


danielpeintner (changes requested):
I am not sure what happens with

Type 'WebSocketClient' is not assignable to type 'ProtocolClient'.
Signature issues?

PR #592, opened 2021-10-20
fix(binding-mqtt): fix eslint errors & warnings

danielpeintner:


packages/binding-mqtt/src/mqtt-client.ts
```diff
@@ -37,24 +37,24 @@ export default class MqttClient implements ProtocolClient {
...
-            let retain = form["mqtt:retain"]; // TODO: is this needed here?
-            let qos = form["mqtt:qos"]; // TODO: is this needed here?
-            let requestUri = url.parse(form["href"]);
-            let topic = requestUri.pathname.slice(1);
-            let brokerUri: String = "mqtt://" + requestUri.host;
-
-            if (this.client == undefined) {
+            const contentType = form.contentType;
+            // const retain = form["mqtt:retain"]; // TODO: is this needed here?
+            // const qos = form["mqtt:qos"]; // TODO: is this needed here?
```
relu91, line 52: I would get rid of those, if we'll need to re-introduce them we can. We can always use git to keep track of old code.

packages/binding-mqtt/src/mqtt.ts
```diff
@@ -43,8 +43,8 @@ export enum MqttQoS {
 
 export class MqttForm extends Form {
     public "mqtt:qos": MqttQoS = MqttQoS.QoS0;
-    //public 'mqtt:topic': string = ''
-    public "mqtt:retain": Boolean;
+    // public 'mqtt:topic': string = ''
```
relu91, line 46: Remove?

packages/binding-mqtt/src/mqtt-client.ts
danielpeintner: removed commented out code

packages/binding-mqtt/src/mqtt.ts
danielpeintner: removed commented out code

relu91 (approved):

PR #594, opened 2021-10-21
fix(binding-mbus): eslint errors/warnings

danielpeintner:


packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
JKRhb, line 2: Are these necessary? Below you used `eslint-disable-next-line` comments.

PR #601, opened 2021-10-29
fix(http): missing types for accept language parser

relu91:
It seems that the poor accept-language-parse was deleted from the import list. Now it is back with its typing package 🥳 .


relu91:
Merging as it is just a fix.

PR #602, opened 2021-10-29
Fix coap binding with regards to counter example

relu91:
This PR tries to fix the issues when you want to try our counter example like the following:
`node packages/cli/dist/cli.js examples/scripts/counter.js` and `node packages/cli/dist/cli.js --clientonly examples/scripts/counter-client.js`. I've just fixed a small issue but there is more ahead, that's why the PR is in a draft status.

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -192,7 +192,7 @@ export default class CoapClient implements ProtocolClient {
     }
 
     public async stop(): Promise<void> {
-        this.agent.close();
+        // do nothing TODO: understand how to properly close coap agent
     }
 
     public setSecurity = (metadata: Array<TD.SecurityScheme>): boolean => true;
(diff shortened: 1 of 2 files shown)


danielpeintner (changes requested):
LGTM and fixes the coap issue.

Note1: Can you run `npm run format`

Note2: I don't think this belongs to this PR but now HTTP is broken. If I run the counter.js and try to load http://localhost:8080/counter the server crashes ...

```
counter ready
[core/servient] Servient caught Asynchronous script error 'C:\Projects\WoT\thingweb\thingweb.node-wot\examples\scripts\counter.js' and halted with ReferenceError: alparser is not defined
    at HttpServer.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot\packages\binding-http\dist\http-server.js:511:54)
```

see https://github.com/eclipse/thingweb.node-wot/issues/614

danielpeintner (approved):

Issue #599, opened 2021-10-29
Opcua binding compile test files into dist folder


relu91:
> > I assume here it is on purpose, right @erossignon?
> 
> Yes, having `"rootdir": ".", dist:"dist"` will cause ts file in src to go to dist/src and test to dist/test/.
> 
> The benefit is that the same tsconfig.json file is used for eslinting (in VSCode) the src and tst files ... It also allows us to check the syntax of the test file during the `tsc -b` phase.
> 
> the downside is that generated files are not stored in the dist folder anymore but one level below.

In our setup, it does not really make sense to compile test files into the dist folder, for two reasons:
- It pollutes the distributed npm packages with files that will never be used
- Tests are executed using ts-node

I think that also opc-ua binding should follow the same setup of other packages.

_Originally posted by @relu91 in https://github.com/eclipse/thingweb.node-wot/issues/598#issuecomment-954547375_

relu91:
Fixed in #600

PR #603, opened 2021-10-29
Fix/ephemeral ports

erossignon:


packages/binding-coap/test/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
     "extends": "../tsconfig.json",
     "compilerOptions": {
-        "rootDir": ".."
+        "rootDir": "..",
+        "experimentalDecorators": true
     },
-    "include": ["*.ts", "**/*.ts", "../src/**/*.ts"]
+    "include": ["test/*.ts", "test/**/*.ts"]
```
relu91, line 7: Where are these changes coming from? are related to issue #599?

relu91 (approved):
I'm ok with merging, just a small question that needs to be answered before continuing.

packages/binding-coap/test/tsconfig.json
erossignon: This change doesn't belong here! let me remove it

erossignon:
@relu91 , please merge this one in priority , it will fix master as well

relu91:
Checked with daniel, merging.

PR #605, opened 2021-10-30
chore: use pretty-quick to only format changed files

erossignon:


erossignon:
example: 
image

relu91:
> Furthermore, by using husky we could enforce format & lint ( and may be also test) to be run as a commit hook.
>
> This will allow us not to be caught again with the failing format/lint step of the CI/CD.

I use husky in other open-source projects. So far so good. I would not complain about having on node-wot 😄 @danielpeintner what's your opinion?

relu91 (approved):
There are just minor conflicts in the package-lock.

danielpeintner (approved):
LGTM

I tested it locally!

The only remaining question I have is whether we should document it in https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md#coding-style

Having said that, I don't think we really need to do that...

relu91:
> The only remaining question I have is whether we should document it in https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md#coding-style

Since it is automatic I don't think is strictly needed! merging :)

PR #608, opened 2021-11-03
fix(binding-file): remove default utf-8 encoding for binding-file and…

lukesmolo:
… improvements for readStreamFully in protocol-helpers

I think that the binding-file should let the specific contentserdes handle the encoding of the stream, while instead it is now forced to _utf8_. Furthermore, for properly handling strings in the _fetch_ operation, an  additional case is needed in the readStreamFully in protocol-helpers


relu91 (approved):
LGTM

Issue #611, opened 2021-11-08
binding-opcua does not handle arrays properly


lukesmolo:
As shown in node-opcua/node-opcua#1074 the _node-opcua_ is not properly handling arrays while converting a _datavalue_ to a javascript object. This translates into the fact that the binding-opcua cannot handle property properties that have an array as value. According to the author of the _node-opcua_ library, this should be fixed in another version of the opcua binding. Nevertheless, since at the moment the current binding is not properly working, a possible solution could be to use a workaround -as suggested by @erossignon- while waiting for the new _node-opcua-json_  module for converting _datavalue_ to JSON.

danielpeintner:
Is this issue still relevant? Or is it solved in new opcua binding?
@relu91 @erossignon @lukesmolo

erossignon:
@danielpeintner, this issue is not relevant anymore, we now formally use OPCUA JSON encoding with normalized support of Array.

PR #613, opened 2021-11-08
fix(binding-opcua): fix for #611

lukesmolo:
Fix for binding-opcua handling arrays (#611)

### DIFF
--- a/packages/binding-opcua/src/opcua-client.ts
+++ b/packages/binding-opcua/src/opcua-client.ts
@@ -163,7 +163,7 @@ export default class OpcuaClient implements ProtocolClient {
         if (result.statusCode === StatusCodes.BadNodeIdUnknown) {
             throw new Error("Invalid nodeId");
         }
-        const body = JSON.stringify(result, (k, v) => {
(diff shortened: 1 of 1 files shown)


relu91 (approved):

packages/binding-opcua/src/opcua-client.ts
```diff
@@ -163,7 +163,14 @@ export default class OpcuaClient implements ProtocolClient {
         if (result.statusCode === StatusCodes.BadNodeIdUnknown) {
             throw new Error("Invalid nodeId");
         }
-        const body = JSON.stringify(result.toJSON());
+        const body = JSON.stringify(result, (k, v) => {
```
erossignon, line 166: In fact, We should use const body = JSON.stringify(result.toJSON(), (k,v)=> ArrayBuffer.isView(v as ArrayLike<unknown>)? Array.from(v) : v);

erossignon (changes requested):

relu91:
Merging since @erossignon comment has been addressed.

PR #617, opened 2021-11-09
fix(coap): resolve issue with URL pathname

JKRhb:
The merge of #538 has caused a small problem in `coap-server.ts` which should be addressed by this PR :)


danielpeintner (approved):

danielpeintner:
order did matter ;-)

relu91:
Plot twist 🤣

PR #618, opened 2021-11-09
Clean up examples

danielpeintner:


danielpeintner:
I looked at the examples and I think we should clean-up the package a bit. There are many samples in that I think can be removed. I rather prefer having a couple of examples but those should serve a good use-case and should be up-to -date.

Question to everybody. "What" are the examples we should keep?
* [browser](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/browser)
  - keep 
* [proxy-scripts](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/proxy-scripts)
  - needed? @wiresio?
* [scripts](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/scripts)
  - Counter
  - Countdown
  - SmartCoffeeMachine (maybe no security only)
  - Are all other samples needed? mqtt, modbus, opcua, netconf, mbus (shouldn't we put most of them in the according binding?)
* [security](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/security)
  - needed, but need to be updated
* [servients](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/servients)
  - not sure again, having them in the according binding?
* [templates](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/templates)
  - @FadySalama looks at it
* [testthing](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/testthing)
  - I think we can easily make it work again


Opinions are welcome!

danielpeintner:
@relu91 @egekorkan @JKRhb @wiresio @sebastiankb others?

Do you have comments to https://github.com/eclipse/thingweb.node-wot/pull/618#issuecomment-964329859 above?

I would like to clean-up (remove?) unneeded parts and concentrate on some *good* examples and keep them in sync as best as we can...

egekorkan:
- Are all other samples needed? mqtt, modbus, opcua, netconf, mbus (shouldn't we put most of them in the according binding?)

Agreeing on this

- servients

Can and should be removed

relu91 (approved):
LGTM, I totally agree with having a smaller pool of well-maintained examples. 

However, it might be tricky to have protocol-specific examples inside the binding package. Examples usually have dependencies with the cli package and a binding-package does not need it. My opinion is that we should still use the examples package even for protocol binding specific examples. 

My suggestion is to go ahead with this PR and start thinking about a structure for protocol-specific examples. Almost all of them just re-implement the default servient using the binding and provide a simple TD. Maybe with a more flexible CLI we can factor out most of the code that is currently in the binding examples.

PR #619, opened 2021-11-10
Update template

FadySalama:
This pull request updates the template to version v.0.8.x


relu91 (changes requested):
Thank you for updating the template!! Just a minor point below. 

p.s. Just a warning: you forgot to sing the commits

examples/templates/exposed-thing/src/base.ts
```diff
@@ -104,21 +108,35 @@ export class WotDevice {
...
     }
 
-    private myActionHandler(inputData) {
+    private myPropertyWriteHandler(inputData, options?) {
         return new Promise((resolve, reject) => {
-            // do something with inputData
-            resolve();
+            // write something to property
+            this.myProperty = inputData;
+            // resolve that write was succesful
```
egekorkan, line 122: it would be good to show emitting observed properties

examples/templates/exposed-thing/src/base.ts
```diff
@@ -76,14 +80,14 @@ export class WotDevice {
         ).then((exposedThing) => {
             this.thing = exposedThing;
             this.td = exposedThing.getThingDescription();
-            this.add_properties();
-            this.add_actions();
-            this.add_events();
-            this.thing.expose();
+            this.addProperties();   // Initialize properties and add their handlers
```
egekorkan, line 83: I would also propose renaming these methods. They were named like this from the pre 0.7.x times where each affordance was added via methods and not via the produce(). A possible name could be `addPropertyHandlers` but there is also initialization happening.

examples/templates/exposed-thing/src/base.ts
```diff
@@ -104,21 +108,35 @@ export class WotDevice {
...
-    private listen_to_myEvent() {
+
+    private myActionHandler(inputData?, options?) {
+        return new Promise((resolve, reject) => {
+            // do something with inputData if available
+            if(inputData) {
+                this.thing.emitEvent("myEvent") // Emiting an event (may be removed; only for demonstration purposes)
+            }
+            //resolve that action was successful
+            resolve(true);
```
egekorkan, line 135: doesn't resolve here correspond to the `outputData` ?

examples/templates/exposed-thing/src/base.ts
```diff
@@ -127,13 +145,14 @@ export class WotDevice {
...
         //fill in add properties
-        this.thing.writeProperty("myProperty", ""); //replace quotes with the initial value
-        this.thing.setPropertyReadHandler("myProperty", this.myPropertyHandler);
+        this.myProperty = ""  ; // replace quotes with the initial value
+        this.thing.setPropertyReadHandler("myProperty", this.myPropertyReadHandler);   // not applicable for write-only
+        this.thing.setPropertyWriteHandler("myProperty", this.myPropertyWriteHandler); // not applicable for read-only
     }
 
-    private add_actions() {
+    private addActions() {
```
egekorkan, line 155: contrary to my comment about properties, this should be called `addActionHandlers`

examples/templates/exposed-thing/src/base.ts
```diff
@@ -104,21 +108,35 @@ export class WotDevice {
...
 
-    private myActionHandler(inputData) {
+    private myPropertyWriteHandler(inputData, options?) {
         return new Promise((resolve, reject) => {
-            // do something with inputData
-            resolve();
+            // write something to property
+            this.myProperty = inputData;
+            // resolve that write was succesful
+            resolve(true);
```
egekorkan, line 123: not sure what the true value does here, as in I am not sure of the behavior

egekorkan (changes requested):

examples/templates/exposed-thing/src/base.ts
relu91: I think we are missing: https://w3c.github.io/wot-scripting-api/#the-emitpropertychange-method . Is going to be added after I've done reviewing: https://github.com/relu91/thingweb.node-wot/pull/17

danielpeintner:
Note: @FadySalama you also need to sign Eclipse ECA

relu91 (changes requested):
Another, neat pick can we use `unknown` instead of `any` as the return type of action and property handler? Using `any` is usually considered a bad practise and I would discourage using it in a template. (To be honest users will need to replace the type anyway, but I think it is good to provide an already well styled code)

danielpeintner:
@FadySalama do you plan to update this PR?

relu91 (approved):
Thank you!!!

danielpeintner:
@egekorkan please let us know whether your comments were also revolved.

egekorkan:
If I am not wrong, none of my comments are fixed @FadySalama ?

FadySalama:
> If I am not wrong, none of my comments are fixed @FadySalama ?

I am pretty sure I did @egekorkan. Can you read the file again and point out which points weren't change?

egekorkan:
My bad, good to be merged :)

PR #620, opened 2021-11-10
style(http): fix eslint issues

JKRhb:
This PR fixes a number of eslint issues in the `binding-http` package.

### DIFF
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -21,13 +21,17 @@ export abstract class Credential {
     abstract sign(request: Request): Promise<Request>;
 }
 
+export interface BasicCredentialConfiguration {
+    username: string;
+    password: string;
+}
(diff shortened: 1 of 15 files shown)


relu91:
I tried to fix all the errors in the src package, we should be almost there. It would be awesome if JKRhb would fix also the remaining issues in the test folder! 🙏🏻

JKRhb:
> It would be awesome if JKRhb would fix also the remaining issues in the test folder! 🙏🏻

Will do :)

relu91:
@JKRhb if you don't mind I'd merge the PR as it is and ask you to create another per for http-tests so that we can move on with #623, is it ok for you?

JKRhb:
Sure, please go ahead :) I will try to provide the follow-up PR shortly.

danielpeintner (approved):

PR #622, opened 2021-11-10
style(binding-netconf): fix eslint issues

JKRhb:
This PR fixes a number of eslint issues in the `binding-netconf` package.

### DIFF
--- a/packages/binding-netconf/src/xpath2json.ts
+++ b/packages/binding-netconf/src/xpath2json.ts
@@ -12,7 +12,6 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
-const util = require("util");
 
 export function isObject(a: any) {
     return !!a && a.constructor === Object;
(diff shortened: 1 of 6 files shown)


packages/binding-netconf/src/xpath2json.ts
```diff
@@ -12,7 +12,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
-var util = require("util");
+const util = require("util");
```
relu91, line 15: Can't we use `import` here?

relu91 (changes requested):
Thanks!!  just one minor concern below:

packages/binding-netconf/src/xpath2json.ts
JKRhb: I noticed that the import can actually be removed as `util` isn't used anywhere :) I will push an updated commit soon.

relu91: I covered the change :)

relu91 (approved):
Sorry, the PR got polluted with the latest changes on the master cause I branched from it to cover my previous review. I the pr now looks ok.

JKRhb:
Oh, no problem! Should I perform I rebase before this is going to be merged?

relu91:
> Oh, no problem! Should I perform a rebase before this is going to be merged?

Why not 😃 👍🏻

JKRhb:
Done :)

danielpeintner (approved):

Issue #544, opened 2021-09-29
Improve ExposedThing - Binding comunication


relu91:
PR #543 introduced a new method for handling content between the core module and the binding level. We should restructure `ExposedThing` following the same approach for consistency and improving the overall handling of streams also at the exposed side. 

A couple of open questions:
- How to handle the `response` parameter forms?
- A protocol binding must also provide the form where the request was issued, how can it correctly choose it? currently, my heuristic is to find the first form with the right protocol scheme and path. But we have other parameters to check like the contentType. 
- An interaction handler expects also UriVariables filled in, how we can populate them? 
- Should we keep the old ConsumedThing methods?

danielpeintner:
>     * Should we keep the old ConsumedThing methods?

Even if we agree on keeping the methods I would argue to rename them to avoid confusion. It should be clear that `ExposedThing` no longer extends `ConsumedThing`.

Issue #323, opened 2020-09-10
[http/core] ERR_STREAM_WRITE_AFTER_END error after the first event


relu91:
The servient crashes with the error in the title when it sends the second event on an HTTP long polling subscription.

### how to reproduce
consumer.js
```js
WoT.consume({
    title: "test",
    events: {
        test: {
            forms: [
                {
                    href: "http://localhost:8080/test/events/test",
                    contentType: "application/json",
                    subprotocol: "longpoll",
                    op: [
                        "subscribeevent",
                        "unsubscribeevent"
                    ]
                }]
        }
    }
    }).then( thing => {
        thing.subscribeEvent("test",()=> console.log("event"))
    } );
```  
producer.js
```js

WoT.produce({
  "title":"test",
  "events":{
    test:{}
  }
}).then(exposing => {
  exposing.expose();
  setInterval(() => {
    exposing.emitEvent("test")
  }, 1000);
});
```
run:
```js
wot-servient producer.js
```
```js
wot-servient -c consumer.js
```

I can volunteer to work on PR so please assign me.

danielpeintner:
Thanks Cristiano!

relu91:
I dug into the problem. Putting simply, the issue is that we do not unsubscribe event callbacks. The `http-server` subscribes to the `thing` event using `thing.subscribeEvent` every time it receives an HTTP LP request. After one event is sent it unsubscribes using `thing.unsubscribe`, but as you can see we do not actually unsubscribe the callback:
```typescript
public unsubscribeEvent(name: string): Promise<void> {
        return new Promise<void>((resolve, reject) => {
            if (this.events[name]) {
                let sub: Subject<any> = this.events[name].getState().subject;
                // sub.unsubscribe(); // XXX causes loop issue (see browser counter example)
                console.debug("[core/exposed-thing]",`ExposedThing '${this.title}' unsubscribes from event '${name}'`);
            } else {
                reject(new Error(`ExposedThing '${this.title}', no event found for '${name}'`));
            }
        });
    }
```
Consequently, the next time that the thing emits an event, we call also the previous callback which contains a dead HTTP connection ---> ERR_STREAM_WRITE_AFTER_END. 

However, it is more complex than it seems. I can solve the loop problem (it is caused by misuse of `rxjs` API) but, now I wonder what happens when two or more consumers crate an HTTP long polling request? see this simplified code:
```js
// HTTP-server

// new HTTP LP request arrived
[...]

PR #623, opened 2021-11-10
Improve ExposedThing interface

relu91:
This PR tackles the points described in #544, it does not completely fix it (more below). As this PR is quite consistent I'll summarize what has been refactored:
- ExposedThing now has a set of methods just for (protocol biding -- core) interaction. Those methods are meant to be called as a means to convey particular operations that are happening from the network interface. For example, if a protocol binding receives a request for reading a property it should call the `handleReadProperty` method in the `ExposedThing` class. This should clearly separate the *application* operations (e.g. readProperty) from the *network interface* callbacks.  
- Every new method follows this naming pattern `handle{operationType}` what do you think about it? would you prefer another naming convention? 
- The contract with the protocol-binding interface is that the binding will offer: affordance name, data (if any), and a reference to the Form that generates it. The form is later used to choose the right codec for serialization or passed to InteractionOutput. 
- Read/WriteAllproperties operations are still a little bit clunky  
- One method that is notably missing in the ExposeThing is the `notifyPropertyChange`, we should fix it in a future PR
- After this refactor we tested this new strategy on the HTTP and CoAP protocol binding. This means that we changed the server implementations to fulfill the new interface. As future work, we'll need to update also Websocket and Mqtt bindings. 
- As a consequence of this PR issue https://github.com/eclipse/thingweb.node-wot/issues/323 should be now solved
- Issue https://github.com/eclipse/thingweb.node-wot/issues/374 will be covered in the future


packages/core/test/ProtocolHelpersTest.ts
```diff
@@ -0,0 +1,169 @@
...
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+
+/**
+ * Basic test suite for helper functions
+ * uncomment the @skip to see failing tests
+ *
+ * h0ru5: there is currently some problem with VSC failing to recognize experimentalDecorators option, it is present in both tsconfigs
```
danielpeintner, line 21: is this comment from @h0ru5 still relevant?

packages/binding-http/test/http-server-test.ts
```diff
@@ -278,18 +367,18 @@ class HttpServerTest {
         testThing.properties.maintenanceNeeded.forms = [];
         testThing.actions.makeDrink.forms = [];
 
-        const td = testThing.getThingDescription();
+        testThing.getThingDescription();
```
danielpeintner, line 370: doesn't really matter but I think the call can be removed

danielpeintner (commented):
Looks good. I still need to check more closely...

danielpeintner:
>     * Every new method follows this naming pattern `handle{operationType}` what do you think about it? would you prefer another naming convention?

name seems ok to me. I guess the most important aspect is that it does no longer conflict with *old* ConsumedThing calls.

packages/core/test/ProtocolHelpersTest.ts
relu91: My impression is that it has been resolved...

packages/binding-http/test/http-server-test.ts
relu91: true! I'll remove it

relu91:
Regarding https://github.com/eclipse/thingweb.node-wot/issues/323 I've been doing some tests locally and it seems that we are almost there. To fully verify that is working correctly I need #620 to be merged as it contains relevant changes.

packages/binding-http/test/http-server-test.ts
relu91: Fixed, 👍🏻

danielpeintner:
> Regarding #323 I've been doing some tests locally and it seems that we are almost there. To fully verify that is working correctly I need #620 to be merged as it contains relevant changes.

Okay, Let's merge #620 and revisit this PR.

danielpeintner:
@relu91 let us know once you think the PR is ready

relu91:
ready to go!

danielpeintner (commented):
I added some minor comments. Feel free to handle them as you like.

I did try to run some manual checks like the counter sample
`node packages\cli\dist\cli.js examples\scripts\counter.js`

Once I try to load a simple property via HTTP 
`http://localhost:8080/counter/properties/count`

I get the value on client side but the server crashes 

```
[core/servient] Servient caught Asynchronous script error 'C:\Projects\WoT\thingweb\thingweb.node-wot\examples\scripts\counter.js' and halted with Error [ERR_HTTP_HEADERS_SENT]: Cannot render headers after they are sent to the client
    at new NodeError (node:internal/errors:371:5)
    at ServerResponse.writeHead (node:_http_server:312:13)
    at HttpServer.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot\packages\binding-http\dist\http-server.js:708:17)
```

What bugs me most is that we seems to have such issues that are not properly covered by our tests!

packages/core/src/exposed-thing.ts
```diff
@@ -317,15 +317,20 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
         );
 
         if (this.properties[name]) {
-            let propertyHandler;
-            if (this.propertyHandlers.has(name)) {
-                propertyHandler = this.propertyHandlers.get(name);
-                propertyHandler.observeHandler = handler;
+            if (!this.properties[name].observable) {
+                throw new Error(
+                    `ExposedThing '${this.title}' cannot set observe handler for property '${name}' due to missing observable flag`
```
danielpeintner, line 322: Nit-picking: The observable flag cannot be "missing" since it is there by default.

"... due to conflicting observable flag"
OR
"since the observable flag is et to false"

danielpeintner (commented):
Question: The bug with the counter sample is not fixed yet, right?

relu91:
> Question: The bug with the counter sample is not fixed yet, right?

Checked locally, I can't reproduce it. 🤔

relu91:
> What bugs me most is that we seems to have such issues that are not properly covered by our tests! 

I agree, maybe we can define some integration tests in the CLI package 🤔

danielpeintner:
The latest commit fixes  ERR_HTTP_HEADERS_SENT error.

Having said that, running the counter sample it still fails... somehow an both sides.

### Server ... count is NaN

```
...
Produced counter
counter ready
[binding-http] HttpServer on port 8080 received no Content-Type from [::ffff:192.168.178.24]:51837
Incrementing count from 0 to 1 (with step 1)
[binding-http] HttpServer on port 8080 received no Content-Type from [::ffff:192.168.178.24]:51839
options = {"formIndex":0,"uriVariables":{"step":3}}
Incrementing count from 1 to 4 (with step 3)
[binding-coap] CoapServer on port 5683 received no Content-Format from 192.168.178.24:62027
options = {"formIndex":2,"uriVariables":{"step":{"type":"integer","minimum":1,"maximum":250}}}
Decrementing count from 4 to NaN (with step [object Object])
[binding-coap] CoapServer on port 5683 received no Content-Format from 192.168.178.24:62028
options = {"formIndex":2,"uriVariables":{"step":{"type":"integer","minimum":1,"maximum":250}}}
Decrementing count from NaN to NaN (with step [object Object])
```

### Client 

```
...
count value is 0
count value after increment #1 is 1
count value after increment #2 (with step 3) is 4
Script error: NotSupportedError: Invalid value according to DataSchema
    at InteractionOutput.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\core\dist\interaction-output.js:58:23)
    at fulfilled (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\core\dist\interaction-output.js:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  value: null
```

relu91:
> The latest commit fixes ERR_HTTP_HEADERS_SENT error.
> 
> Having said that, running the counter sample it still fails... somehow an both sides.
> 
> ### Server ... count is NaN
> ```
> ...
> Produced counter
> counter ready
> [binding-http] HttpServer on port 8080 received no Content-Type from [::ffff:192.168.178.24]:51837
> Incrementing count from 0 to 1 (with step 1)
> [binding-http] HttpServer on port 8080 received no Content-Type from [::ffff:192.168.178.24]:51839
> options = {"formIndex":0,"uriVariables":{"step":3}}
> Incrementing count from 1 to 4 (with step 3)
> [binding-coap] CoapServer on port 5683 received no Content-Format from 192.168.178.24:62027
> options = {"formIndex":2,"uriVariables":{"step":{"type":"integer","minimum":1,"maximum":250}}}
> Decrementing count from 4 to NaN (with step [object Object])
> [binding-coap] CoapServer on port 5683 received no Content-Format from 192.168.178.24:62028
> options = {"formIndex":2,"uriVariables":{"step":{"type":"integer","minimum":1,"maximum":250}}}
> Decrementing count from NaN to NaN (with step [object Object])
> ```
> 
> ### Client
> ```
> ...
> count value is 0
> count value after increment #1 is 1
> count value after increment #2 (with step 3) is 4
> Script error: NotSupportedError: Invalid value according to DataSchema
>     at InteractionOutput.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\core\dist\interaction-output.js:58:23)
>     at Generator.next (<anonymous>)
>     at fulfilled (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\core\dist\interaction-output.js:5:58)
>     at processTicksAndRejections (node:internal/process/task_queues:96:5) {
>   value: null
> ```

I "fixed" this too. Basically, the problem was that the new `coap-server` was sending an invalid value as `uriVariables`, therefore I've just changed it to return an empty object. This is not really related to this PR cause looking at the file history it seems that it didn't handle `uriVariables` before (but it was correctly returning `undefined`). Therefore I suggest going on with this PR and creating a new issue about this misbehavior of the `coap-server` implementation.

danielpeintner:
Thanks @relu91.
I created a dedicated CoAP issue, see https://github.com/eclipse/thingweb.node-wot/issues/636.

W.r.t. to the PR it looks good, even though there is still a minor issue... with the CLI.

The client script 
`node packages\cli\dist\cli.js --clientonly examples\scripts\counter-client.js`
runs fine, but fails to properly close ... it _hangs_ in the end.

Shall we merge this PR and look at it in a follow-up PR?

danielpeintner (approved):

relu91:
> The client script
node packages\cli\dist\cli.js --clientonly examples\scripts\counter-client.js
runs fine, but fails to properly close ... it hangs in the end.

yeah, I forgot to mention it! I think is something related to CoAp again... but I'll open an issue to keep track of this!

PR #625, opened 2021-11-11
refactor: re-introduce server test

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/523

### DIFF
--- a/packages/binding-coap/test/coap-server-test.ts
+++ b/packages/binding-coap/test/coap-server-test.ts
@@ -18,7 +18,7 @@
  */
 
 import { suite, test } from "@testdeck/mocha";
-import { expect, should } from "chai";
+import { expect, should, assert } from "chai";
 
 import CoapServer from "../src/coap-server";
 // should must be called to augment all variables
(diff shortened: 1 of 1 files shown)


danielpeintner:
Note: I am not sure what the commented out client test is meant to do?
@relu91 do you know?

relu91:
If I remember correctly it checks that if another coap server is running on the same port it should throw an error. I remember that we solved this issue in https://github.com/mcollina/node-coap, right @JKRhb? Now I'm wondering if it makes sense to do this test on our package... I think it is more a low-level feature.

JKRhb:
> I remember that we solved this issue in https://github.com/mcollina/node-coap, right @JKRhb?

Yeah, we resolved it in https://github.com/mcollina/node-coap/pull/271/, it has to activated by passing a config parameter (`reuseAddr: false`)  to the server, though. I am also not sure if this should be tested here as it is mostly a behavior of the dependency.

danielpeintner:
Exactly, server test can be re-introduced. That's what this PR is doing.
Having said that, there are also client-tests I am not sure about?

https://github.com/eclipse/thingweb.node-wot/blob/dc99f9a848d4d18dbab8045f062b8e104d2a2c65/packages/binding-coap/test/coap-client-test.ts#L44-L101

danielpeintner:
Is it ok to merge (add the server test again) since node-coap has been fixed by now?

JKRhb:
> Is it ok to merge (add the server test again) since node-coap has been fixed by now?

Hmm, one aspect of what @relu91 and I meant above is that the `EADDRINUSE` error is currently not actually being thrown – removing the try-catch-block does not cause the test to fail. For this we need to add an additional parameter to the `createServer` function in `coap-server.ts`.

JKRhb (commented):
If the patch in https://github.com/eclipse/thingweb.node-wot/pull/625#issuecomment-970330832 is applied then these two lines have to removed if I am not mistaken.

Maybe there could also be another assertion added that the `EADDRINUSE` exception is actually being thrown?

danielpeintner:
@JKRhb thanks for you input. Let's wait a bit with this PR since https://github.com/eclipse/thingweb.node-wot/pull/623 also introduces  some *new* tests...

danielpeintner:
@JKRhb do think I shall close this PR. is there any follow-up PR needed?

JKRhb:
Hmm, would you say that the test that has been commented out should be removed? Otherwise I would open a consolidated PR with the suggestions I made above (which should make reviewing them a bit easier).

JKRhb:
I guess one question that needs to be answered in this context is whether it should be allowed to run two server instances in parallel or not.

danielpeintner:
> I guess one question that needs to be answered in this context is whether it should be allowed to run two server instances in parallel or not.

You mean in the *same* instance... with 2 different ports?
I did never try that... neither for other bindings like HTTP...

Running 2 instances separately (with different ports) should just work fine.. or do I miss anything here?

JKRhb:
Oh, no, I actually meant two servers running on the same port. By default, I think this is possible in node-coap but would be disable with the patch I posted above.

danielpeintner:
Ahh, to be honest I don't see benefit in having two servers listening on the same port.. I am surprised this works. Hence, I am OK if this does not work 🙃

@relu91 any opinion... do you have such use-cases ?

relu91:
> Ahh, to be honest I don't see benefit in having two servers listening on the same port.. I am surprised this works. Hence, I am OK if this does not work 🙃
> 
> @relu91 any opinion... do you have such use-cases ?

+1 for throwing when there's already another server listening to the same port. `http` throws therefore I think it is ok to have consistent behavior in node-wot.

danielpeintner:
closed in favor of https://github.com/eclipse/thingweb.node-wot/pull/657

PR #626, opened 2021-11-12
Issue interaction input to readable

danielpeintner:
see https://github.com/eclipse/thingweb.node-wot/issues/482#issuecomment-966497766

### DIFF
--- a/packages/binding-http/test/http-server-test.ts
+++ b/packages/binding-http/test/http-server-test.ts
@@ -131,7 +131,7 @@ class HttpServerTest {
         return httpServer.stop();
     }
 
-    @test async "should change object resource from {} to {'new': true} and try to invoke"() {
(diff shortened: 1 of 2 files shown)


relu91:
I'm not sure if this is really needed after #623

danielpeintner:
> I'm not sure if this is really needed after #623

Great. Anyhow, we should add a test with object/array as the once in this PR to check for it besides simple types.

relu91 (approved):
Ah ok, now I got why we create the PR anyway. Is the name of the test correct? It feels a little bit obscure to me. What about:
> "should serialize objects for actions and properties"

danielpeintner:
> Ah ok, now I got why we create the PR anyway. Is the name of the test correct? It feels a little bit obscure to me. What about:
> 
> > "should serialize objects for actions and properties"

👍

PR #628, opened 2021-11-12
refactor: remove WebSocket from server start since it is not availabl…

danielpeintner:
…e on client side

fixes https://github.com/eclipse/thingweb.node-wot/issues/627


relu91 (approved):
Merging, I'll open an issue as a reminder to re-add WebSockets when it ready (unless we have already one)

PR #634, opened 2021-12-01
chore(coap): update node-coap to version 1.x.x and fix type issues

JKRhb:
This PR updates the `node-coap` dependency to the new major release. There are still a couple of minor issues that have to be resolved first, however (see https://github.com/mcollina/node-coap/pull/322), which is why I opened this PR as a draft for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1104,6 +1104,11 @@
                 "@types/node": "*"
             }
         },
+        "node_modules/@types/capitalize": {
+            "version": "2.0.0",
+            "resolved": "https://registry.npmjs.org/@types/capitalize/-/capitalize-2.0.0.tgz",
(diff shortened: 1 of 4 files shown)


JKRhb:
The PR is now ready :) Unfortunately, I had to introduce a couple of workarounds as `node-coap` uses the (apparently outdated) `stream` module whose Typescript interfaces are incompatible with the regular Node.JS stream interfaces. I will have a look into replacing this dependency to get rid of the type casting.

As a major addition this PR now features the `close` method on the `Agent` which could be a potential fix for #637.

relu91 (approved):
LGTM, reminder open an issue to keep track of the type casting after merge.

danielpeintner (approved):
LGTM

Note1: created https://github.com/eclipse/thingweb.node-wot/issues/642 to keep track of the type-casts

Note2:  #637 is not revolved by this PR 😌

JKRhb:
> LGTM
> 
> Note1: created #642 to keep track of the type-casts

Thank you! :)

> 
> Note2: #637 is not revolved by this PR 😌

Oh, that is unfortunate :( I will have another look into this problem.

PR #635, opened 2021-12-03
Handle security scheme `name` in http binding

### LINKED ISSUE #631: BasicCredential does not let define name for header
In the method sign of class BasicCredential the "name" parameter of the thing description security schema definition specifying the name of the header to be set is ignored and the default _authorization_ is used instead.

``` 
async sign(request: Request): Promise<Request> {
        const result = request.clone();
        result.headers.set(
            "authorization",
            "Basic " + Buffer.from(this.username + ":" + this.password).toString("base64")
        );
        return result;
    }
```

relu91:
With this changes the http-binding now correctly uses the information in `SecurityScheme` to configure a request object. In particular, they focus on handling correctly the parameter `name`. 

Fixes #631 
cc: @fillobotto

### DIFF
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -33,7 +33,7 @@ export class BasicCredential extends Credential {
     /**
      *
      */
-    constructor({ username, password }: BasicCredentialConfiguration, options: BasicSecurityScheme) {
+    constructor({ username, password }: BasicCredentialConfiguration, options?: BasicSecurityScheme) {
         super();
         if (username === undefined || password === undefined || username === null || password === null) {
             throw new Error(`No Basic credentials for Thing`);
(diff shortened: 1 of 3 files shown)


packages/binding-http/src/credential.ts
```diff
@@ -1,3 +1,5 @@
+import { BearerSecurityScheme } from "./../../td-tools/dist/thing-description.d";
+import { BasicSecurityScheme, APIKeySecurityScheme } from "@node-wot/td-tools";
```
danielpeintner, line 2: Minor: imports should come "after" the copyright.

packages/binding-http/src/http-client.ts
```diff
@@ -231,13 +231,18 @@ export default class HttpClient implements ProtocolClient {
         // TODO support for multiple security schemes
```
danielpeintner, line 231: can (should?) this comment be removed?

packages/binding-http/test/credential-test.ts
```diff
@@ -0,0 +1,95 @@
+import { BearerSecurityScheme } from "./../../td-tools/src/thing-description";
+import {
+    BasicCredential,
+    BasicCredentialConfiguration,
+    BearerCredential,
+    BearerCredentialConfiguration,
+    BasicKeyCredential,
+    BasicKeyCredentialConfiguration,
+} from "./../src/credential";
```
danielpeintner, line 9: imports should come after copyright

danielpeintner (commented):
I posted some overview comments only

relu91:
Moved to #638

PR #638, opened 2021-12-09
Handle security scheme `name` in http binding

### LINKED ISSUE #631: BasicCredential does not let define name for header

relu91:
With these changes, the HTTP-binding now correctly uses the information in `SecurityScheme` to configure a request object. In particular, they focus on handling correctly the parameter `name`.

Fixes #631

cc: @fillobotto

### DIFF
--- a/packages/binding-http/test/credential-test.ts
+++ b/packages/binding-http/test/credential-test.ts
@@ -1,12 +1,3 @@
-import { BearerSecurityScheme } from "./../../td-tools/src/thing-description";
-import {
-    BasicCredential,
-    BasicCredentialConfiguration,
-    BearerCredential,
-    BearerCredentialConfiguration,
-    BasicKeyCredential,
-    BasicKeyCredentialConfiguration,
-} from "./../src/credential";
 /********************************************************************************
  * Copyright (c) 2018 - 2020 Contributors to the Eclipse Foundation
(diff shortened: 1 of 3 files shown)


packages/binding-http/test/credential-test.ts
```diff
@@ -0,0 +1,94 @@
+/********************************************************************************
+ * Copyright (c) 2018 - 2020 Contributors to the Eclipse Foundation
```
danielpeintner, line 2: Minor: the years are not correct, It should be 2021 only.

danielpeintner (approved):
LGTM

mentioned one minor issue that will be fixed by the copyright script in the end also.

packages/binding-http/test/credential-test.ts
danielpeintner: Note: let's the copyright script fix the issue..

PR #641, opened 2021-12-14
chore: update copyright year in LICENSE file

JKRhb:
This trivial PR updates the LICENSE file with the year 2021. 

As 2022 is not far away anymore: Maybe this file could also be updated in the copyright header script?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2018 - 2019 Contributors to the Eclipse Foundation
+Copyright (c) 2018 - 2021 Contributors to the Eclipse Foundation
 
 See the NOTICE file(s) distributed with this work for additional
 information regarding copyright ownership.


danielpeintner:
FYI: We usually update copyright years only if the file itself has been updated. Not sure if that makes  sense :-)

The copyright header script essentially also looks at the git history.
Shall we update regardless?

relu91:
It seems that [US regulations](https://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/) just ask for the year of publication..... 

Which are the policies inside the eclipse foundation? I would keep the changes to the source code minimal

relu91 (approved):

JKRhb:
> It seems that [US regulations](https://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/) just ask for the year of publication.....

That would definitely be the most elegant solution.

> 
> Which are the policies inside the eclipse foundation? I would keep the changes to the source code minimal

I just found [this FAQ](https://www.eclipse.org/projects/handbook/#ip-copyright-headers) by the Eclipse Foundation where they state that giving a year range is optional. So I suppose the range could also be removed. They actually use commas for year ranges, by the way ;)

danielpeintner:
2 questions:

1. What should we do with this PR. I am OK to merge it even though I don't think it is actually necessary
2. The more important question to me is whether we should actually change year ranges? It would bring a lot of noise, but could easily be done by fixing the copyright script.
  - `2018 - 2021` to `2018, 2021`

Any comments?

relu91:
> 

My take would be to resolve this PR (merge or not is up to you) and discuss the copyright ranges on an issue.

danielpeintner:
Let's merge this PR and look at the other issue in https://github.com/eclipse/thingweb.node-wot/issues/643

PR #646, opened 2021-12-20
Use `eslint-plugin-notice` for license header policies

### LINKED ISSUE #645: Use eslint plugin to make sure Copyright is there
see https://github.com/eclipse/thingweb.node-wot/pull/644#issuecomment-996729349
> @relu91 
> About this, let [eslint plugin](https://www.npmjs.com/package/eslint-plugin-notice) do the work for us and contributors, shall we? I can volunteer to add this feature :)

relu91:
Fixes #645. Moreover, it fixes additional license headers error leftovers after #641.


danielpeintner (approved):
LGTM 👍

The *old* copyright script did not look into test folders.
BTW, shall we remove the old script? I don't think this is necessary anymore, right?

PR #652, opened 2022-01-24
fix(coap-server): remove type casting to unknown

JKRhb:
After the latest update of node-coap, the type-casting introduces by #634 can now be removed as the `OutgoingMessage` class now provides the correct type information regarding the interfaces it implements.

This PR updates node-coap to the latest version and removes the now obsolete type casting. Resolves https://github.com/eclipse/thingweb.node-wot/issues/642.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
     "requires": true,
     "packages": {
         "": {
-            "name": "thingweb.node-wot",
             "license": "EPL-2.0 OR W3C-20150513",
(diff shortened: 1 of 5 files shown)


JKRhb:
Seems as if I opened this PR a bit too early :/ There are still some typing issues.

danielpeintner (approved):
LGTM

Thanks!

relu91 (approved):
Thank you! well done 👍🏻

Issue #457, opened 2021-08-04
Mqtt implementation needs improvements


relu91:
I've recently worked on #456 and I had the opportunity to have a look into the current implementation of the MQTT binding. I think it can be improved in multiple points, some of which may impact also the binding specification. 

- [x] Missing the ability to self-host the broker. I think that the server implementation should be able to self host the broker not only connect to an external remote node. 
- [x] Missing a correct definition of the configuration parameters. For reference please have a look into the current HTTP implementation where the client has a properly defined interface in the http.ts for the config object.   
- [x] Client implementation lacks the ability to connect using user and password
- [ ] Proper definition of form parameters. Now QoS and retain are not really used in the code. We need to implement a proper model in the binding document. 
- [x] Unit tests are not self-contained but depend upon core libraries. I think we should just test the protocol APIs rather than the full stack in a binding package. Or if we want to test the full stack we should provide smaller tests that just focus on binding APIs. 
- [x] Refactor the expose function in smaller units (but this is a problem of almost every binding). 
- [x] Refactor client implementation to have the same code style of other clients (e.g., it uses `method = () => {}` instead of `public method(){}`)

egekorkan:
Some comments:
- Self-hosting: For the Thing simulator we have, we are using an node based broker called `aedes`. See: https://github.com/tum-esi/shadow-thing/blob/dc5bcbabcb75e1fe8281e3a5bfeafb548c056720/src/cli.ts#L287
- For username/password, you mean Consumer cannot connect right? Also, it does not happen through the servient config like it should

relu91:
After reviewing #562 I added another point to the list above: 
- Refactor client implementation to have the same code style of other clients (e.g., it uses method = () => {} instead of public method(){})

Basically the client uses another style to define class methods: 
https://github.com/eclipse/thingweb.node-wot/blob/82b97b186d3de0d04836f7dc3378b0b7c22580d3/packages/binding-mqtt/src/mqtt-client.ts#L95-L100

danielpeintner:
relates to https://github.com/eclipse/thingweb.node-wot/issues/569

egekorkan:
Regarding the remaining item about qos and retain:
- We should make sure that the Consumer follows the forms requirements. If the form has a specific qos or retain value, they should be copied to the publish request. At this example: https://w3c.github.io/wot-binding-templates/bindings/protocols/mqtt/index.html#example-qos, it is indicating the Consumer to use those parameters.

[...]

PR #656, opened 2022-01-28
Improve mqtt implementation

relu91:
This PRs contains a set of improvements to the current binding MQTT package. It covers most of the changes tracked in #457, in particular with the reference to @fillobotto's [comment](https://github.com/relu91/thingweb.node-wot/pull/22#issue-1084193453):
- [X] Missing the ability to self-host the broker. I think that the server implementation should be able to self host the broker not only connect to an external remote node. 
- [X] Missing a correct definition of the configuration parameters. For reference please have a look into the current HTTP implementation where the client has a properly defined interface in the http.ts for the config object.   
- [X] Client implementation lacks the ability to connect using user and password
- [ ] Proper definition of form parameters. Now QoS and retain are not really used in the code. We need to implement a proper model in the binding document. 
- [X] Unit tests are not self-contained but depend upon core libraries. I think we should just test the protocol APIs rather than the full stack in a binding package. Or if we want to test the full stack we should provide smaller tests that just focus on binding APIs. (In the end we decided to split the tests into two files `.integration` and `.unit`) 
- [X] Refactor the expose function in smaller units (but this is a problem of almost every binding). 
- [x] Refactor client implementation to have the same code style of other clients (e.g., it uses `method = () => {}` instead of `public method(){}`)


packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -410,11 +417,19 @@ export default class MqttBrokerServer implements ProtocolServer {
...
+            }
+
+            if (this.hostedBroker !== undefined) {
+                this.hostedServer.close();
+                this.hostedBroker.close(() => resolve());
+            }
 
-            this.broker.unsubscribe("*");
+            resolve();
         });
```
JKRhb, line 433: As this function is labelled as `async`, is explicitly returning a Promise still necessary?

JKRhb (commented):
Looks good to me! :) See some comments below

packages/binding-mqtt/src/mqtt.ts
```diff
@@ -47,5 +47,23 @@ export class MqttForm extends Form {
...
+    password?: string;
     rejectUnauthorized?: boolean;
 }
+
+export interface MqttBrokerServerConfig {
+    uri: string;
+    user?: string;
+    psw?: string;
+    clientId?: string;
+    protocolVersion?: number;
```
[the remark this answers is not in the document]
egekorkan: nitpicking: 3.1.1 is the actual version and I have no idea if there is a 3.0 or 3.1.0

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -31,51 +34,38 @@ export default class MqttBrokerServer implements ProtocolServer {
...
+            if (config.uri.indexOf("://") === -1) {
+                config.uri = this.scheme + "://" + config.uri;
             }
-            this.brokerURI = uri;
+            this.brokerURI = config.uri;
         }
-
-        if (user !== undefined) {
-            this.user = user;
+        if (config.selfHost) {
```
egekorkan, line 58: in the case of self hosting, the broker uri can be separately provided or the static address in the config can be also used. What is the rationale behind having the two possibilities?

egekorkan (commented):
Thanks for this, I wanted an improved version of this for a while :D I have left some minor comments, here are some other ones:
- Can be in another PR but it would be nice to be able to configure QoS for the entire client as a config parameter. So that the Thing can publish events always with QoS 2. Regarding what can be seen in the TD, it is indeed a topic for the binding templates but I can think of the Thing saying that invoking this action needs QoS 2 so that it is not invoked twice by mistake. The discussion can/should continue somewhere else probably :)
- I have found this at the MQTT Wikipedia: https://en.wikipedia.org/wiki/MQTT#Applications We can add node-wot there :)

relu91:
Thanks, everybody for the reviews, I'll address your comments ASAP. 

> Can be in another PR but it would be nice to be able to configure QoS for the entire client as a config parameter. So that the Thing can publish events always with QoS 2. Regarding what can be seen in the TD, it is indeed a topic for the binding templates but I can think of the Thing saying that invoking this action needs QoS 2 so that it is not invoked twice by mistake. The discussion can/should continue somewhere else probably :)

Good point this is another reason why we didn't fully complete the tasks in #457, it would be better to finalize mqtt protocol binding document and then finish up the implementation.

packages/binding-mqtt/src/mqtt.ts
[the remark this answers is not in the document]
relu91: I've checked mqtt.js, AFIK they are just using the major protocol versions. I've changed the type to only accept 3, 4, or 5. We might need to think more how to handle this in the TD.

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: Fixed using just two promises.

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: Not sure what are you asking. Are you saying that I should peak somehow the `staticAddress` from the CLI config? In this case, it seems that this is handled in the Helpers class and not at the server level (neither HTTP server is using that property).

relu91:
I've should have covered your comments please let me know if there's anything else that I can improve!

packages/binding-mqtt/src/mqtt-broker-server.ts
egekorkan: I didn't know that HTTP was not using it properly. It is just weird that if I have am self hosting the broker and have an HTTP binding as well, I need to write twice my IP address.

egekorkan: This is relevant only for the generated TDs, the request will arrive properly in any case

packages/binding-mqtt/src/mqtt.ts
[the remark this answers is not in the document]
egekorkan: Something really weird is that there is no MQTT spec with version 4, but [mqtt.js even defaults to that version](https://github.com/mqttjs/MQTT.js/blob/7ec4b8fd602e220f50693cb83f082dab764ed3f2/lib/client.js#L28) :D Not sure what to do here since a node-wot user does need to know about mqtt.js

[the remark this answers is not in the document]
relu91: My proposal is to track this missaligment in an issue. Then we can investigate more and understand how we can map mqtt protolVersion onto mqtt.js versions.

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: I think we can resolve this discussion then :)

danielpeintner:
I let @egekorkan and @JKRhb take a look at MQTT and approve since I am not so much in that topic 🙃

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -102,255 +95,278 @@ export default class MqttBrokerServer implements ProtocolServer {
...
-                if (!content) {
+            const observeListener = async (data: Content) => {
+                let content;
+                try {
+                    content = ContentSerdes.get().valueToContent(data, property.data);
+                } catch (err) {
                     console.warn(
                         "[binding-mqtt]",
-                        `HttpServer on port ${this.getPort()} cannot process data for Event ${eventName}`
+                        `MqttServer cannot process data for Property '${propertyName}': ${err.message}`
```
danielpeintner, line 136: @relu91 there is one more HttpServer log that should be fixed in this file.

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -102,255 +95,278 @@ export default class MqttBrokerServer implements ProtocolServer {
...
+        console.debug(
+            "[binding-mqtt]",
+            `MqttBrokerServer at ${this.brokerURI} assigns '${href}' to Event '${eventName}'`
+        );
+
+        const eventListener = async (content: Content) => {
+            if (!content) {
+                console.warn(
+                    "[binding-mqtt]",
+                    `HttpServer on port ${this.getPort()} cannot process data for Event ${eventName}`
```
danielpeintner, line 196: Here.. it needs to be changed.

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -360,35 +376,29 @@ export default class MqttBrokerServer implements ProtocolServer {
                 resolve();
             } else {
                 // try to connect to the broker without or with credentials
-                if (this.psw === undefined) {
+                if (this.config.psw === undefined) {
```
danielpeintner, line 379: Shall we have an external if to check whether this.config exists? Or is is impossible?

```
if (this.config) {
 ...
}
```

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: Thank you! good catch! 👍🏻

packages/binding-mqtt/src/mqtt-broker-server.ts
relu91: you're right `this.config` is initialized [here](https://github.com/eclipse/thingweb.node-wot/pull/656/files#diff-1b0afdf0b85e178b3f8d59ae8dba12e7ee767d9f046f20205a319a2462b16c8dR49) and it might be `undefined`. However, instead of adding the if here I prefer to change the initialization like the following: 
```js
this.config = config ?? {};
```

I think it will work better, won't it?

danielpeintner: fine by me also

relu91:
I resolved the final comments, @danielpeintner @egekorkan @JKRhb going to merge this PR this evening if the tests are green. Please let me know if you find some blocking issues.

PR #658, opened 2022-01-29
style(http-binding): Fix linting issues

JKRhb:
This PR fixes most of the linting issues in the http binding. A few `any`s had to be introduced, however, as the `InMemoryModel` defined for testing does not completely implement an OAuth2 model. This issue should probably be tackled in a follow-up PR, though.

### DIFF
--- a/packages/binding-http/test/http-client-oauth-tests.ts
+++ b/packages/binding-http/test/http-client-oauth-tests.ts
@@ -22,32 +22,34 @@ import { OAuth2SecurityScheme } from "@node-wot/td-tools";
 import InMemoryModel from "./memory-model";
 import { promisify } from "util";
 import { ProtocolHelpers } from "@node-wot/core";
-const OAuthServer = require("express-oauth-server");
-const bodyParser = require("body-parser");
+import { readFileSync } from "fs";
 
-const fs = require("fs");
+import OAuthServer from "express-oauth-server";
+import bodyParser from "body-parser";
+
+import { AuthorizationCodeModel } from "oauth2-server";
 
 @suite("HTTP oauth client implementation")
 class HttpClientOAuthTest {
     private client: HttpClient;
-    static model: any;
+    static model: AuthorizationCodeModel;
     static server: https.Server;
 
     static before(): Promise<void> {
-        const app: any = express();
-        HttpClientOAuthTest.model = new InMemoryModel();
+        const app: express.Express = express();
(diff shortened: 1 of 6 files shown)


danielpeintner (approved):

packages/binding-http/test/http-client-oauth-tests.ts
```diff
@@ -22,41 +22,43 @@ import { OAuth2SecurityScheme } from "@node-wot/td-tools";
...
+import OAuthServer from "express-oauth-server";
+import bodyParser from "body-parser";
+
+import { AuthorizationCodeModel } from "oauth2-server";
 
 @suite("HTTP oauth client implementation")
 class HttpClientOAuthTest {
     private client: HttpClient;
-    static model: any;
+    static model: AuthorizationCodeModel;
```
relu91, line 35:
```suggestion
    static model: InMemoryModell;
```

Should solve also the other `any`s, have you tried it?

relu91 (changes requested):
LGTM, just a suggestion below

packages/binding-http/test/http-client-oauth-tests.ts
JKRhb: I think I tried it out before, as setting the type of `model` to `InMemoryModel` leads to the following errors 😕

```
Type 'InMemoryModel' is not assignable to type 'AuthorizationCodeModel | ClientCredentialsModel | RefreshTokenModel | PasswordModel | ExtensionModel'.
  Property 'verifyScope' is missing in type 'InMemoryModel' but required in type 'RefreshTokenModel'.ts(2322)
index.d.ts(285, 9): 'verifyScope' is declared here.
index.d.ts(170, 9): The expected type comes from property 'model' which is declared here on type 'Options'
```

I think `InMemoryModel` should probably implement/extend one of the mentioned types so that we can get rid of the type casting.

relu91:
After painfully updating the types I got rid of those `any`s :) with this, the `http-binding` should now be eslint approved 🥳

relu91:
@JKRhb may ask you to give the final green light?

relu91 (approved):

JKRhb:
> After painfully updating the types I got rid of those `any`s :) with this, the `http-binding` should now be eslint approved 🥳

Oh, that is awesome! 🎉

> @JKRhb may ask you to give the final green light?

Sure, please go ahead! :)

PR #659, opened 2022-01-29
style(binding-netconf): resolve linting issues

JKRhb:
This PR resolves most of the eslint errors and warnings in the `binding-netconf` package. Unfortunately, a number of `any`s had to be introduced as a number of parameters and types require a bit more work than expected. These should probably be introduced in a follow-up PR.

### DIFF
--- a/packages/binding-netconf/src/async-node-netconf.ts
+++ b/packages/binding-netconf/src/async-node-netconf.ts
@@ -37,15 +37,15 @@ export class Client {
         this.router = null;
     }
 
-    getRouter() {
+    getRouter(): any {
         return this.router;
     }
 
-    deleteRouter() {
+    deleteRouter(): void {
         this.router = null;
     }
 
-    async initializeRouter(host: string, port: number, credentials: any) {
+    async initializeRouter(host: string, port: number, credentials: any): Promise<void> {
         if (this.router && this.router.connected) {
             // close the old one
             this.closeRouter();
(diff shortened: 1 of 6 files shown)


danielpeintner (approved):

packages/binding-netconf/src/async-node-netconf.ts
```diff
@@ -37,15 +37,15 @@ export class Client {
         this.router = null;
     }
 
-    getRouter() {
+    getRouter(): any {
```
relu91, line 40: isn't this a `nodeNetconf.Client` ?

packages/binding-netconf/src/async-node-netconf.ts
```diff
@@ -65,63 +65,58 @@ export class Client {
...
         return new Promise((resolve, reject) => {
             if (payload) {
-                xpath_query = xpath2json.addLeaves(xpath_query, payload);
+                xpathQuery = xpath2json.addLeaves(xpathQuery, payload);
             }
-            const obj_request = xpath2json.xpath2json(xpath_query, NSs);
-            let final_request: any = {};
-            final_request = JSON.parse(JSON.stringify(METHOD_OBJ[method])); // clone the METHOD_OBJ
+            const objRequest = xpath2json.xpath2json(xpathQuery, NSs);
+            let finalRequest: any = {};
```
relu91, line 95: I know that it might get a little verbose but for JSON parsed objects `unknown` it is a better type than `any`

relu91 (changes requested):
netconf package code is pretty hard to fix (eslint speaking), you did a great job. Some minor suggestions below. Not sure if they can improve the situation a little more.

packages/binding-netconf/src/async-node-netconf.ts
JKRhb: Hmm, I think this does not work with the current code as changing the type to `unknown` leads (for instance) to this error:

```
Property 'edit-config' does not exist on type '{}'.ts(7053)
```

packages/binding-netconf/src/async-node-netconf.ts
JKRhb: Hmm, I think the return value of this getter is only the parameters needed to *instantiate* a `nodeNetconf.Client`. Changing line 34 to 

```
private router: nodeNetconf.Client;
```

leads to a number of errors, as this `router` variable is actually passed to a `nodeNetconf.Client` constructor in the `openRouter()` method. Maybe the `router` variable should be renamed to something like `routerParameters` instead and typed with an appropriate interface.

relu91:
After thinking a bit, I would move on since the PR has already improved the code and the remaining `any`s can be fixed later on.

relu91 (approved):

PR #660, opened 2022-01-29
style(binding-modbus): resolve linting issues

JKRhb:
This PR fixes most of the eslint issues in the Modbus binding. There are still a couple of `any`s which will have to be revisited as well as some errors due to the order of interface definitions but those should (once more) be dealt with in a follow-up PR.


danielpeintner (approved):
Thanks for your work @JKRhb !

relu91 (approved):
Thanks! merging right after this comment :)

PR #663, opened 2022-02-05
chore: update wot-typescript-definition

erossignon:
update to latest "wot-typescript-definition" ( version 21) 
and use "pinned" versions to avoid subtle side effects when minor breaking changes are introduced in a patched version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
         "check:versions": "node utils/check_package_version_consistency.js",
         "build:docker": "docker build -t wot-servient .",
         "clean:dist": "npm exec  --workspaces -- npx rimraf tsconfig.tsbuildinfo dist",
+        "update:wot-typescript-definitions": "npx npm-check-updates  -u -f \"wot-typescript-definitions\" --deep",
         "link": "npm link -ws"
     },
     "workspaces": [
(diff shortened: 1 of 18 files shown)


relu91 (approved):
LGTM, package-lock should avoid these "side-effects" for minor releases, but it does not hurt to pin the version.

package.json
```diff
@@ -31,6 +31,7 @@
         "check:versions": "node utils/check_package_version_consistency.js",
         "build:docker": "docker build -t wot-servient .",
         "clean:dist": "npm exec  --workspaces -- npx rimraf tsconfig.tsbuildinfo dist",
+        "update:wot-typescript-definitions": "npx npm-check-updates  -u -f \"wot-typescript-definitions\" --deep",
```
danielpeintner, line 34: I wonder whether we want to have that. @erossignon you added a similar check for opcua in another PR.
@relu91 do you think this is fine?

erossignon: I find this very useful to consistently upgrade a particular module used everywhere in a mono repo.

danielpeintner: I see your arguments

danielpeintner (approved):

PR #665, opened 2022-02-05
chore(core): display info to better understand why schema validation …

erossignon:
this PR adds a tactical message to the console to help investigate why the AJV schema validation has failed.
Prior to this change, there was no way to detect why the system was throwing an exception. 

I am happy to replace this with a better mechanism, as console.log is not always welcome in production code. 
Please advice.

### DIFF
--- a/packages/binding-opcua/src/opcua-client.ts
+++ b/packages/binding-opcua/src/opcua-client.ts
@@ -245,7 +245,7 @@ export default class OpcuaClient implements ProtocolClient {
             };
             const result = await this.session.call(methodToCall);
 
-            console.log("[binding-opcua]", "[invoke]", result.toString());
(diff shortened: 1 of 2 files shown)


relu91 (changes requested):
I think having this log message could help to understand what is going on. I would approve of this change. However, please follow the [logging guidelines](https://github.com/eclipse/thingweb.node-wot#logging) and don't forget to sign your commits 😃 . Thank you for your contribution!

relu91 (changes requested):
From logging guidelines: 
> 2. Avoid to use info and log in packages other than the cli package.

I think the log messages you added should be logged under `console.debug` .

relu91 (approved):

danielpeintner (approved):

PR #667, opened 2022-02-05
Update node opcua

erossignon:


package.json
```diff
@@ -25,6 +25,7 @@
         "lint": "npm run lint --silent --workspaces --if-present",
         "lint:fix": "npm run lint:fix --silent --workspaces --if-present",
         "format": "prettier --write . && npm run format --silent --workspaces --if-present",
+        "update:opcua": "npx -y npm-check-updates -u --deep -f \"node-opcua*\" && npm install --workspaces",
```
danielpeintner, line 28: Mhh, needed/wanted ?

erossignon: Yes, in the new binding there will be a dozen of node-opcua-* modules to update, this script will become handy.

danielpeintner: I see  (https://github.com/eclipse/thingweb.node-wot/pull/663#pullrequestreview-874492744)

danielpeintner (approved):

package.json
relu91: Can't this be moved to the `binding-opcua` `package.json`?

erossignon: I'll drop this PR , I don't thnk it is needed anymore as the new binding will overwriteit.

PR #674, opened 2022-02-15
Thing model helpers

relu91:
This PR adds a set of helpers functions to handle [Thing Models](https://w3c.github.io/wot-thing-description/#thing-model). The changes might be a little intimidating but the core contribution is contained `thing-model-helpers.ts`. There we added three public functions:
1. `public static isThingModel(_data: unknown): _data is ThingModel`: simply checks that the data is "castable" to a ThingModel. This is useful when loading TMs from files or raw json strings. Its main benefit is to automatically cast _data to ThingModel and do a runtime check. 
2. `public static validateThingModel(data: ThingModel): { valid: boolean; errors: string }`: validate a ThingModel instance accordingly to W3C ThingModel json schema
3. `public async getPartialTDs(model: unknown, options?: CompositionOptions): Promise<ExposedThingInit[]>`: the main function of the helper class. It basically instantiates a Thing Description from a model. It returns an array because in the case of composed Thing Models we choose to use the approach of linking to other sub TDs, instead of flattening everything. Notice that model is `unknown` to easily plug in raw json data. 
4. `public static getModelVersion(data: ThingModel): string`: utility function to retrieve the version. 
5.  `public async fetchModel(uri: string): Promise<ThingModel> `: download a ThingModel

For further explanation about how to use these new functions, I suggest giving a look at `ThingModelHelperCompositionTest.ts` and `ThingModelHelperTest.ts`. 

We tried to follow the directions described in the spec please advise if you find anything off: I'm looking forward for your feedback! 

Open issues:
- I would prefer to have the validation functions inside the td-tools package. Consequently, I'd like to rename it wot-tools to capture the fact that contains functions to handle also TMs and in the future maybe other utilities. 
- What I don't like too much about this implementation is the dependency on `Servient`. The servient is needed for fetching linked TMs. However, in the future, we can define a `Resolver` interface and break the direct dependency with  `Servient`. 
- the `fetchModel` function contains a side-effect that records the URLs for dependency cycle detection. I would say that it would be better to remove this side-effect and handle the cycle detection differently. A solution could be TM hashing, this would be independent of the TM URL but it makes the instantiation algorithm more CPU intensive. 
- The tool does not handle placeholders that contains objects

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12015,6 +12015,12 @@
             "resolved": "https://registry.npmjs.org/wot-thing-description-types/-/wot-thing-description-types-1.1.0-27-September-2021.tgz",
             "integrity": "sha512-DprWwVZKnEJSS5qNI0NHJSLYv6tmIkA3wSXjl9OskAXFMnQpxTwnlNT960f341XFDJF9pUBfi7nsD+7tnaFlvA=="
         },
+        "node_modules/wot-thing-model-types": {
+            "version": "1.1.0-3-February-2022",
+            "resolved": "https://registry.npmjs.org/wot-thing-model-types/-/wot-thing-model-types-1.1.0-3-February-2022.tgz",
+            "integrity": "sha512-IYetSr4Z7DmXESMg2OKDyIPTRwGSciIMuJBPyABNptNHodbwcMHx/48GO0kx6TpsDy9gBYkGO1SFXvt1/NZoMw==",
+            "dev": true
+        },
         "node_modules/wot-typescript-definitions": {
             "version": "0.8.0-SNAPSHOT.21",
             "resolved": "https://registry.npmjs.org/wot-typescript-definitions/-/wot-typescript-definitions-0.8.0-SNAPSHOT.21.tgz",
@@ -12704,6 +12710,7 @@
                 "typescript": "4.4.3",
                 "typescript-standard": "^0.3.36",
                 "wot-thing-description-types": "^1.1.0-26-July-2021a",
+                "wot-thing-model-types": "^1.1.0-3-February-2022",
                 "wot-typescript-definitions": "^0.8.0-SNAPSHOT.21"
             }
         },
@@ -13691,6 +13698,7 @@
                 "vm2": "^3.9.6",
                 "web-streams-polyfill": "^3.0.1",
                 "wot-thing-description-types": "^1.1.0-26-July-2021a",
+                "wot-thing-model-types": "1.1.0-3-February-2022",
                 "wot-typescript-definitions": "^0.8.0-SNAPSHOT.21"
             }
         },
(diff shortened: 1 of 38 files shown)


danielpeintner:
I did not fully review the PR but I have some very *generic* questions / comments.

* I would have expected the ThingModel tooling under `td-tools`, similar to validating TDs. 
  Is the main issue the dependency on `Servient` as mentioned (which may go away as you described) or are there other parts the NEED `core` tooling? My feeling (so far) is, that `core` *starts* once we have already a proper TD.
* I saw you implemented your *own*  placeholder replacer? Seems okay, on the other hand there is a library that we could use..
  https://www.npmjs.com/package/json-placeholder-replacer. Did you consider that?

relu91:
Thank you for the feedback! Good points:
> Is the main issue the dependency on Servient as mentioned (which may go away as you described) or are there other parts the NEED core tooling? My feeling (so far) is, that core starts once we have already a proper TD.

No other dependencies it is just the fetching of ThingModels that needs a `Servient` instance. I agree that it would be better to move everything in td-tools. Should we do the change that I described above in this PR? I would opt for doing that right after this one. 

> I saw you implemented your own placeholder replacer? Seems okay, on the other hand, there is a library that we could use..
https://www.npmjs.com/package/json-placeholder-replacer. Did you consider that?

The replacement logic is pretty simple so far and the library seems not well supported (the last code commit is two years old and it has only 6 stars). Nonetheless, if you have some strong points for having that we can plug that in!

danielpeintner:
> > Is the main issue the dependency on Servient as mentioned (which may go away as you described) or are there other parts the NEED core tooling? My feeling (so far) is, that core starts once we have already a proper TD.
> 
> No other dependencies it is just the fetching of ThingModels that needs a `Servient` instance. I agree that it would be better to move everything in td-tools. Should we do the change that I described above in this PR? I would opt for doing that right after this one.

I think either way is fine. In the end I would really like to see it in td-tools 👍

> > I saw you implemented your own placeholder replacer? Seems okay, on the other hand, there is a library that we could use..
> > https://www.npmjs.com/package/json-placeholder-replacer. Did you consider that?
> 
> The replacement logic is pretty simple so far and the library seems not well supported (the last code commit is two years old and it has only 6 stars). Nonetheless, if you have some strong points for having that we can plug that in!

Maybe I would feel more comfortable if we have placeholder tests for composed types also... I think this may cause some burdens in the end. I leave it up top you.

danielpeintner:
> > Maybe I would feel more comfortable if we have placeholder tests for composed types also
> 
> Do you mean `objects` as values in the placeholder map?

Yes, correct.
I think this should be possible also. Or am I wrong?

relu91:
Based on the discussion in https://github.com/w3c/wot-thing-description/issues/1399, I added this point to the open issues:
> - The tool does not handle placeholders that contains objects

My preference is to merge this PR as it is and handle the issues above in smaller chunks. @danielpeintner what do you think?

danielpeintner:
> My preference is to merge this PR as it is and handle the issues above in smaller chunks. @danielpeintner what do you think?

Fine by me. I created https://github.com/eclipse/thingweb.node-wot/issues/680 to keep track of the discussion. Feel free to add aspects I might have forgotten.

danielpeintner (approved):

PR #677, opened 2022-02-17
downgrade to windows-2019 instead of latest

erossignon:
- nodejs 12 will be phased out in April 2022
     but github action environement "windows-latest" has
     introduced a breaking change causing a built failure.
     This PR forces a former windows image to be used
     until we phase out nodejs 12 support.


relu91 (approved):
Sounds good! thank you.

PR #682, opened 2022-02-24
Improve readme of all packages

egekorkan:
So I wanted to fix the language highlighting of the README of the MQTT package and in the end, I ended up going through all the packages. A few changes can be opinionated. What I have done in general:
- Make sure that code snippets have a language (having js after the triple ` )
- URI Scheme of each protocol is mentioned in the beginning
- Fix tabs to spaces in code snippets
- At the end of each readme has a clickable link that points to the github home page of node-wot (important since packages published to npm see the package readme and they are not aware of the entire project)
- Added Wikipedia links to protocols (most of them at least)
- Added Binding Template link to the beginning. Putting this in the beginning seems a bit weird, maybe.
- Fixed typos and English issues

Some notes:

- I think that each protocol should have a simple overview section in the beginning, a bit like [firestore](https://github.com/egekorkan/thingweb.node-wot/tree/master/packages/binding-firestore) has. 
- The examples in the readmes are old. While I was doing this, https://github.com/eclipse/thingweb.node-wot/pull/669 was still not merged so I did not do this yet.
- I think that each readme should contain a TD that can be consumed by the client of this protocol. Given that all protocols at least have the client-side, this sounds reasonable since TDs can be written by hand.

### DIFF
--- a/packages/binding-mqtt/README.md
+++ b/packages/binding-mqtt/README.md
@@ -1,5 +1,13 @@
 # Eclipse Thingweb node-wot MQTT Binding
 
+W3C Web of Things (WoT) Protocol Binding for [MQTT](https://en.wikipedia.org/wiki/MQTT).
+This package uses [mqtt](https://www.npmjs.com/package/mqtt) as a low level library for MQTT.
+W3C WoT Binding Template for MQTT can be found [here](https://w3c.github.io/wot-binding-templates/bindings/protocols/mqtt/index.html).
+
+## Protocol specifier
+
+The protocol prefix handled by this binding is `mqtt://` or `mqtts://`.
+
 ## Getting Started
 
 In the following examples it is shown how to use the MQTT binding of node-wot.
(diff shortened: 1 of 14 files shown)


packages/binding-mqtt/README.md
```diff
@@ -1,5 +1,13 @@
 # Eclipse Thingweb node-wot MQTT Binding
```
relu91, line 1: This title here is a little bit different from the others, should we align it?

relu91 (approved):
Ok for merging, just a minor comment that can be addressed later on.

packages/binding-mqtt/README.md
egekorkan: Fixed it now in the last commit

danielpeintner (approved):
LGTM, thanks!

JKRhb:
Great improvements, thank you @egekorkan!

Issue #640, opened 2021-12-13
Evaluate `fast-uri` package


relu91:
Head over to https://github.com/fastify/fast-uri. It supports also IPV6 something that we are looking for in our applications.

PR #686, opened 2022-02-25
Fix browser example

relu91:
This PR introduces back the browser bundle among the supported packages. A couple of notes:
1. I discovered that `node-fetch` when used in browsers defaults to native `fetch` implementation this means the http client was sending `ReadableStreams` (instead of `Readable` from NodeJS) instance to the core. The core is expecting that every binding respect the contract of the `Content` interface which requires a NodeJS `Readable`. This PR changes `binding-http` to convert the stream before sending it the the `core`
2. `browserify` is using an old package for replacing nodejs `url` package. This creates issues when directly importing `url`. I discovered that using the global `URL` variable bypass `browserify` and it works well in browsers too. This PRs adds this workaround. Maybe #640 is another possible solution
3. I improved a little bit the example code and updated it
4. I fixed a bug when handling HTTP log polling requests from events. The servent was not handling head requests properly.

### DIFF
--- a/examples/browser/counter.js
+++ b/examples/browser/counter.js
@@ -44,12 +44,12 @@ function showInteractions(thing) {
             dtItem.appendChild(link);
             document.getElementById("properties").appendChild(dtItem);
             document.getElementById("properties").appendChild(ddItem);
-
+            console.log("property: " + td.properties[property]);
             dtItem.onclick = (click) => {
                 thing
                     .readProperty(property)
-                    .then((res) => {
-                        ddItem.textContent = res;
+                    .then(async (res) => {
+                        ddItem.textContent = await res.value();
                     })
                     .catch((err) => window.alert("error: " + err));
             };
@@ -103,7 +103,7 @@ function showInteractions(thing) {
             document.getElementById("events").appendChild(item);
 
             eventSubscriptions[evnt] = false;
-
+            let subscription;
             checkbox.onclick = (click) => {
                 if (
                     document.getElementById(evnt).checked &&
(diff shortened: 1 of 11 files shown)


examples/browser/counter.js
```diff
@@ -44,12 +44,12 @@ function showInteractions(thing) {
             dtItem.appendChild(link);
             document.getElementById("properties").appendChild(dtItem);
             document.getElementById("properties").appendChild(ddItem);
-
+            console.log("property: " + td.properties[property]);
```
relu91, line 47: Just noticed this left over, I'll update the pr after CI gives green light

danielpeintner:
Thanks @relu91 👍 

Note: No review, I still need to take a look. I just want to mention that there are 2 more "browser" examples we need to work on
* index
* smart-coffee-machine

We can tackle them in another PR. As you like @relu91?

relu91:
Oh yeah, you're right I think we can do that in another PR!

danielpeintner (changes requested):
I did run into some issues on my side
* a global `npm run build` does NOT create the dist folder for browser-bundle.
One needs to walk into packages\browser-bundle and call 
`npm run build` there to get `wot-bundle.min.js`
* When trying to run the counter.html I get several errors and the UI does not show
  - Uncaught TypeError: class heritage _$core_31.default is not an object or null
  - Uncaught ReferenceError: Wot is not defined

grafik

relu91:
> a global npm run build does NOT create the dist folder for browser-bundle.
One needs to walk into packages\browser-bundle and call
npm run build there to get wot-bundle.min.js

Nice catch I didn't check for that sorry. Since the `browser-bundle` is not a TS project I've added the relevant commend in `build`.

> When trying to run the counter.html I get several errors and the UI does not show
Uncaught TypeError: class heritage _$core_31.default is not an object or null
Uncaught ReferenceError: Wot is not defined

It seems that is a `tinify` error... I temporally remove it but I think we should file an issue so that in the future we can understand what's going on.

danielpeintner:
> It seems that is a `tinify` error... I temporally remove it but I think we should file an issue so that in the future we can understand what's going on.

https://github.com/eclipse/thingweb.node-wot/issues/690 created

danielpeintner (approved):
Looks good!

Minor: do you want to remove the console.log ?

PR #689, opened 2022-02-28
docs: fix copyright year based on git log

danielpeintner:


relu91:
Looks good! I don't think the es-lint plugin does the same check. It would be awesome to have this integrated in our CI pipeline so that we don't have to run it periodically.

relu91 (approved):

danielpeintner:
> Looks good! I don't think the es-lint plugin does the same check. It would be awesome to have this integrated in our CI pipeline so that we don't have to run it periodically.

I agree, but for the time being it seems easier for me to run it manually from time top time 🙃

PR #691, opened 2022-03-02
add code of conduct and security MD

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/688

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,10 +10,5 @@ The thingweb team will mark them with the according [security](https://github.co
 
 ## Disclosure
 
-Disclosure is initially limited to the reporter and all Eclipse Committers,
-but is expanded to include other individuals, and the general public.
-The timing and manner of disclosure is governed by the
-[Eclipse Security Policy](https://www.eclipse.org/security/policy.php).
-
 Publicly disclosed issues are listed on the
 [Disclosed Vulnerabilities Issues Page](https://github.com/eclipse/thingweb.node-wot/labels/security).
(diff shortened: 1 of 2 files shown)


danielpeintner:
Note: I added the code of conduct "as is" from https://github.com/eclipse-ee4j/.github and simplified the security file a lot.

SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+
+Vulnerabilities can be reported via [GitHub issues](https://github.com/eclipse/thingweb.node-wot/issues).
+The thingweb team will mark them with the according [security](https://github.com/eclipse/thingweb.node-wot/labels/security) label.
+
+## Disclosure
+
+Disclosure is initially limited to the reporter and all Eclipse Committers,
+but is expanded to include other individuals, and the general public.
+The timing and manner of disclosure is governed by the
+[Eclipse Security Policy](https://www.eclipse.org/security/policy.php).
```
relu91, line 16: I think this statement is not valid since we support only GitHub issues. Github issues are public which means that we can't limit the disclosure to just Eclipse Commiters.

relu91 (changes requested):
I agree with the changes but I've found a problem in the security document.

SECURITY.md
danielpeintner: Good point!
I removed the disclosure limitation (since we act in public).

I wonder whether we need the whole section. Personally I don't think it hurts to list it again.
Opinions?

relu91 (approved):
I'm ok for keeping the section.

PR #695, opened 2022-03-11
Support `uriVariables` at the root level

### LINKED ISSUE #670: Top level URI Variables
assuming https://github.com/w3c/wot-thing-description/pull/1383 gets approved we need to be able to handle *global* `uriVariables` also.

relu91:
Fix #670


danielpeintner:
Minor: coding style check fails

danielpeintner (changes requested):
I wanted to update the counter example, see https://github.com/eclipse/thingweb.node-wot/commit/f68c03144d068333637063f460a006d1285c488e

I assume using global uriVariables should work but with this PR it does no longer work... can you check?

POST http://192.168.178.24:8080/counter/actions/increment?step=2 should increment by 2 .. but it does not.

relu91:
> I assume using global uriVariables should work but with this PR it does no longer work... can you check?

I should update the counter example with your changes and understand why is not working anymore, right? Will do!

danielpeintner:
> I should update the counter example with your changes and understand why is not working anymore, right? Will do!

Yes, updating the counter with global uriVariables in this PR might be good 🙃

Somehow the variables are not passed along to the handler

packages/binding-coap/src/coap-server.ts
```diff
@@ -562,6 +580,48 @@ export default class CoapServer implements ProtocolServer {
...
+                } else {
+                    params[queryKey] = queryValue;
+                }
+            }
+        });
+
+        return params;
+    }
+
     private isEmpty(obj: Record<string, unknown>) {
```
erossignon, line 625: isEmpty doesn't depend on this and can be put outside the class as a simple function to avoid prototype pollution.

packages/binding-coap/src/coap-server.ts
```diff
@@ -562,6 +580,48 @@ export default class CoapServer implements ProtocolServer {
         res.end("Not Found");
     }
 
+    private parseUrlParameters(
```
erossignon, line 583: can be static

packages/binding-http/src/http-server.ts
```diff
@@ -487,7 +487,11 @@ export default class HttpServer implements ProtocolServer {
         }
     }
 
-    private parseUrlParameters(url: string, uriVariables: { [key: string]: TD.DataSchema }): Record<string, unknown> {
+    private parseUrlParameters(
```
erossignon, line 490: DRY: this is a duplication of coap-server#parseUrlParameter ?  may be a shared utility method is needed instead.

erossignon, line 490: DUplication of code

erossignon (changes requested):

danielpeintner (approved):
LGTM now!
Thanks!

I will add the revised example with global uriVariables in another PR.

PR #696, opened 2022-03-11
Issue 692

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/692

### DIFF
--- a/examples/scripts/countdown.js
+++ b/examples/scripts/countdown.js
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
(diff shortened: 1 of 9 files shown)


danielpeintner:
@egekorkan @relu91  can someone check what I did.
I tried to add tests for the various cases.

Note: I also updated the examples so that for the TestFest both URLs get reporte

danielpeintner:
Note: The idea of this PR is that if there is no "wot/td" context I add both ("https://www.w3.org/2019/wot/td/v1" and "https://www.w3.org/2022/wot/td/v1.1")
IF there is one I leave it as is.

egekorkan (approved):

relu91 (approved):
LGTM; just one question do we have the requirement to put the context link always as the first entry?

PR #698, opened 2022-03-14
avoid issues with no data for event

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/697

### DIFF
--- a/examples/browser/counter.js
+++ b/examples/browser/counter.js
@@ -113,7 +113,7 @@ function showInteractions(thing) {
                     eventSubscriptions[evnt] = true;
                     thing
                         .subscribeEvent(evnt, async function (data) {
-                            console.log("Data:" + (await data.value()));
+                            console.log('Event "' + evnt + '"');
                             updateProperties();
                         })
                         .then((sub) => {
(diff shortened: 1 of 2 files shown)


danielpeintner:
Not sure why the pipeline fails.

I don't think this has to do with my  changes which do not affect any code that is tested...

examples/browser/counter.js
```diff
@@ -113,7 +113,7 @@ function showInteractions(thing) {
                     eventSubscriptions[evnt] = true;
                     thing
                         .subscribeEvent(evnt, async function (data) {
-                            console.log("Data:" + (await data.value()));
+                            console.log('Event "' + evnt + '"');
```
relu91, line 116: This should output just the `InteractionOutput` object, is it really what we want? In principle the previous code was correct and was showing how to properly handle event payloads. I think this new code it is a little bit misleading for new users. Another option is to add as payload  `null` so that the example will still work and you can use `await data.value()` here.

danielpeintner: Yes, either we report null OR we leave the console log as is.. it does only print the event name.

relu91: Oh sorry, I've overlooked I was thinking that you were printing `data`. Ok for merging then!

relu91 (approved):

PR #709, opened 2022-03-18
Solve Eslint issues with circular definitions

JKRhb:
This PR should resolve the first point of #534. The typescript-specific rule `@typescript-eslint/no-use-before-define` is added as an replacement for `no-use-before-define` which gets rid of some false positives. In the remaining cases, the issues are resolved by reordering definitions so that they are only used after they have been introduced.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,8 @@ module.exports = {
         ],
         "workspaces/no-relative-imports": "error",
         "@typescript-eslint/no-unused-vars": "off", // or "@typescript-eslint/no-unused-vars": "off",
+        "no-use-before-define": "off",
+        "@typescript-eslint/no-use-before-define": ["error"],
         "unused-imports/no-unused-imports": "error",
         "unused-imports/no-unused-vars": [
             "warn",
--- a/packages/binding-http/src/oauth-token-validation.ts
+++ b/packages/binding-http/src/oauth-token-validation.ts
@@ -51,21 +51,30 @@ interface TokenInformation {
     client_id?: string;
 }
 
-export default function (method?: Method): EndpointValidator {
-    if (!method || !method?.name) {
-        throw new Error("Undefined oauth token validation method");
-    }
-
-    switch (method.name) {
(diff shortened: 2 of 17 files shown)


relu91 (changes requested):
Looks good but you introduced a bunch of errors in the core package too... Since it was hard to get everything correct I would prefer to:
- leave everything as it is
- OR fix every error that this new change introduce

relu91:
Oh, it is still in draft sorry :) I guess you are going to solve the additional errors :)

JKRhb:
> Oh, it is still in draft sorry :) I guess you are going to solve the additional errors :)

I just converted it to one as I noticed the additional errors 😅 I'll try to resolve the new problems and then we'll hopefully get rid of this issue for good :)

JKRhb:
The errors should be resolved now :) Had to apply a lot of reordering, though :/

JKRhb:
Hmm, I think the CI pipeline needs to be restarted, not sure why it failed

relu91 (approved):
LGTM, let's wait for the CI

danielpeintner (approved):
LGTM even though there was a lot of re-ordering which makes it difficult to review 🙃

Do you think we should update prettierrc.json in this PR or a separate one?
As you like!

> BTW, I noticed that in [.prettierrc.json](https://github.com/eclipse/thingweb.node-wot/blob/59fe3264bb4cfa19a3c8cddb0f1c68b4685a4a47/.prettierrc.json) we still use es5. Shall we change that too?
see https://github.com/eclipse/thingweb.node-wot/search?q=ES5

relu91:
> Do you think we should update prettierrc.json in this PR or a separate one?
@JKRhb can you do this here?

JKRhb:
> > Do you think we should update prettierrc.json in this PR or a separate one?
> 
> @JKRhb can you do this here?

Hmm, I think I would open another PR for that, as the change actually results in 65 changed files :/

JKRhb:
I opened #711 for the updates to `prettierrc.json` :)

PR #710, opened 2022-03-18
fix(examples): fix oauth2 exposer example

JKRhb:
This PR resolves another eslint issue by updating the `wot-thing-description-types` dependency and removing the `@ts-ignore` comment from the OAuth2 exposer example.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14337,9 +14337,9 @@
             "dev": true
         },
         "node_modules/wot-thing-description-types": {
-            "version": "1.1.0-27-September-2021",
-            "resolved": "https://registry.npmjs.org/wot-thing-description-types/-/wot-thing-description-types-1.1.0-27-September-2021.tgz",
-            "integrity": "sha512-DprWwVZKnEJSS5qNI0NHJSLYv6tmIkA3wSXjl9OskAXFMnQpxTwnlNT960f341XFDJF9pUBfi7nsD+7tnaFlvA=="
+            "version": "1.1.0-09-February-2022",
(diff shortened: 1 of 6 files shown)


relu91:
Well done I think we should remember when we update `wot-typescript-definitions` like here https://github.com/eclipse/thingweb.node-wot/pull/707 . However, the lastest version is from Febrary: [1.1.0-09-February-2022](https://www.npmjs.com/package/wot-thing-description-types/v/1.1.0-09-February-2022) where are you using an old one?

relu91:
p.s. @danielpeintner what bugs me is that the latest version of wot-typescript-defintions is from 2 days ago, whereas wot-thing-description-types is from February. something is off... isn't it?

danielpeintner:
> However, the lastest version is from Febrary: [1.1.0-09-February-2022](https://www.npmjs.com/package/wot-thing-description-types/v/1.1.0-09-February-2022) where are you using an old one?

Good point. We should use "1.1.0-09-February-2022"

> p.s. @danielpeintner what bugs me is that the latest version of wot-typescript-defintions is from 2 days ago, whereas wot-thing-description-types is from February. something is off... isn't it?

I don't think so.

wot-typescript-definitions@0.8.0-SNAPSHOT.22 was published by my recently since we forgot to push changes on NPM.
wot-thing-description-types@1.1.0-09-February-2022 matches with github [history](https://github.com/w3c/wot-scripting-api/commits/main/typescript/thing-description/package.json)

danielpeintner (changes requested):
Please use 1.1.0-09-February-2022

JKRhb:
> Please use 1.1.0-09-February-2022

Done :) Thank you for noticing!

relu91:
and with this we are eslint errors free🎉🎉 merging.

PR #662, opened 2022-02-01
Feature/new binding opcua

erossignon:

PR #711, opened 2022-03-18
style: change trailingComma option to "all" and reformat codebase

JKRhb:
This PR removes the `es5` value from the `trailingComma` option of pettier and changes it to `all`, adding trailing commas to function and method parameter lists. The codebase is reformatted accordingly.


relu91 (approved):

erossignon:
Can we merge this PR after merging the new opcua binding which now contains a dozen of commits from October 2021=> 
This would avoid merging issues , thanks

JKRhb:
> Can we merge this PR after merging the new opcua binding which now contains a dozen of commits from October 2021=> This would avoid merging issues , thanks

Sure! I'll rebase once #662 is merged :)

danielpeintner:
An alternative would be to use `trailingComma": "none"` OR to go with the implicit default by removing the option completely.
I think the default would be "es5" but we do not mention it in our file (see https://prettier.io/docs/en/options.html#trailing-commas)

Doing so would not result in any other changes..

Any thoughts, opinions ?

danielpeintner:
@relu91 do you have any preference w.r.t. _trailingComma_ style?

relu91:
tbh I don't... the default seems a reasonable assumption... but I don't have a strong opinion!

JKRhb:
Then I think we can just leave it as is :) If we think the style should be adjusted at a later point in time, we can simply update the config and re-run prettier. I would then simply close the PR :)

Issue #516, opened 2021-09-20
Reenable failing for linting Action 


JKRhb:
In #512 the linting Action was configured to succeed even if there are any linting issues present to avoid a failure in every CI run. Once all linting issues have been resolved, this change should be reverted.

relu91:
Fixed in #712

PR #712, opened 2022-03-18
chore: let CI fail on linting error

JKRhb:
As all eslint errors have now been fixed, we can actually let the CI fail if there should be any linting errors, asserting better code quality :)


relu91:
Let's go! Fix #516 🥳

PR #714, opened 2022-03-19
style(binding-netconf): address linting issues

JKRhb:
This PR applies some refactorings to the netconf binding, resolving a number of eslint warnings. There are now six warnings remaining in the `netconf-codec.ts` and the `xpath2json.ts` files which seem to require more substantial changes/type checking to be fixed. These should probably be dealt with in a follow-up PR.

### DIFF
--- a/packages/binding-netconf/src/codecs/netconf-codec.ts
+++ b/packages/binding-netconf/src/codecs/netconf-codec.ts
@@ -15,6 +15,12 @@
 
 import * as TD from "@node-wot/td-tools";
 import Url from "url-parse";
+import { DataSchemaValue } from "wot-typescript-definitions";
+
+interface PayloadNamespaces {
+    payload: unknown;
+    namespaces: Record<string, string>;
+}
 
 /** default implementation offering JSON de-/serialisation */
 export default class NetconfCodec {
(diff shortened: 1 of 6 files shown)


packages/binding-netconf/src/codecs/netconf-codec.ts
```diff
@@ -70,32 +75,30 @@ export default class NetconfCodec {
...
-    valueToBytes(value: any, schema: TD.DataSchema, parameters?: { [key: string]: string }): Buffer {
+    valueToBytes(value: unknown, schema: TD.DataSchema, parameters?: { [key: string]: string }): Buffer {
         // console.debug("NetconfCodec serializing", value);
         let body = "";
         if (value !== undefined) {
             const NSs = {};
-            // let leaf = value.leaf;
-            let leaf = <any>schema;
-            leaf = leaf.forms[0].href.split("/").splice(-1, 1); // take the first one, since there is no difference for the leaf
+            let leaf = schema.forms[0].href.split("/").splice(-1, 1); // take the first one, since there is no difference for the leaf
```
JKRhb, line 101: This part of the method seems to be broken at the moment as it only works with property affordances and not with generic dataschemas.

packages/binding-netconf/src/codecs/netconf-codec.ts
```diff
@@ -164,27 +172,12 @@ export default class NetconfCodec {
...
-        console.debug("[binding-netconf]", obj);
-        for (const k in obj) {
-            if (Object.prototype.hasOwnProperty.call(k)) {
-                // recursive call to scan property
-                mapJsonToArray(obj[k]);
-            }
-        }
-    } else {
-        // not an Object so obj[k] here is a value
-        console.debug("[binding-netconf]", obj);
```
JKRhb, line 188: I removed this function as it isn't used anywhere and does not return anything.

PR #716, opened 2022-03-20
chore: replace nyc with c8, update `package-lock.json`

JKRhb:
This PR should fix https://github.com/eclipse/thingweb.node-wot/issues/616 by replacing nyc with the native coverage tool c8 :) By using c8, it seems as if Typescript is handled out of the box so `@istanbuljs/nyc-config-typescript` could be removed which caused the warning regarding the outdated `uuid` dependency.

The dependencies in `package-lock.json` are also updated, addressing a couple of security warnings.


relu91 (approved):
Well done!

danielpeintner (approved):

PR #718, opened 2022-03-21
fix: remove internal keys from getThingDescription() call

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/705


danielpeintner:
Note: Maybe it makes sense to add "_" before internal keys
`propertyHandlers` -> `_propertyHandlers`
to avoid possible collisions?

Any thoughts?

relu91 (approved):
yes, prepend "_" would be a nice idea.

danielpeintner:
Note: I prefixed internal variables to avoid possible TD term collisions

PR #721, opened 2022-03-28
fix: examples in READMEs for v0.8

danielpeintner:


danielpeintner:
TODO MQTT update is still missing

@JKRhb the CoAP `example-client.js` hangs when reaching out to "coap://plugfest.thingweb.io:5683/testthing".
I believe the code is correct.. just not sure whether this is an issue of the server part?
I don't have good CoAP tools around but I tried another CoAP client library and seem to have the same issue. Hence to me it seems like a server issue? @JKRhb Do you know?

danielpeintner:
> isn't this caused by the reasons described in [my comment](https://github.com/eclipse/thingweb.node-wot/issues/637#issuecomment-1070881519)?

Maybe I was not very clear... it _hangs_ from the beginning on after the first GET.. NO result is reported.

Note1: I did try locally and then it reports values. It seems it has to do again with reaching out to the web again.
Note2: The PR https://github.com/eclipse/thingweb.node-wot/pull/720 is integrated. So this is not the issue. Maybe the Eclipse vserver?
Note3: Attached the test-file I am using.

```
// example-client.js
Servient = require("@node-wot/core").Servient;
CoapClientFactory = require("@node-wot/binding-coap").CoapClientFactory;

Helpers = require("@node-wot/core").Helpers;

servient.addClientFactory(new CoapClientFactory(null));

wotHelper
    .fetch("coap://plugfest.thingweb.io:5683/testthing") // coap://localhost:5683/testthing
    .then(async (td) => {
        // using await for serial execution (note 'async' in then() of fetch())
        try {
            const thing = await WoT.consume(td);

            // read property
            const read1 = await thing.readProperty("string");
            console.log("string value is: ", await read1.value());
        } catch (err) {
            console.error("Script error:", err);
        }
    })
    .catch((err) => {
        console.error("Fetch error:", err);
    });
```

relu91 (approved):
LGTM! thanks! 🥳  I'd merge this, should we open an issue for the hanging problem?

danielpeintner:
> LGTM! thanks! 🥳 I'd merge this, should we open an issue for the hanging problem?

Created https://github.com/eclipse/thingweb.node-wot/issues/724 to keep track

fatadel:
@relu91 @danielpeintner 
I haven't really tested but I was exploring the COAP example in readme and saw there

`servient.addClientFactory(new CoapClientFactory(null));`

How is this legit if it passes null as CoapServer, which was working before but in v8 the field is not nullable anymore?

PR #723, opened 2022-03-28
Move vm from core to cli package

### LINKED ISSUE #377: Removing VM from core package
Even though I had discussed it with @danielpeintner and @relu91 I have realized that I did not open an issue about it.

In the core module and still in the 0.8.x branch at the servient (https://github.com/eclipse/thingweb.node-wot/blob/4c9152cb7bb8f10957b1ac9efa24c6d323dbe5c0/packages/core/src/servient.ts#L16) the VM module is included and the run script is also there. I think that this functionality should be moved to the CLI since it is not about the core servient functionality. It also causes issues in webpack where we have to manually remove the includes in where node-wot is installed at node_modules at https://github.com/tum-esi/wade

relu91:
Being close to our big release I thought that it makes sense to improve node-wot architecture with this substantial refactor. This PR moves the VM dependence from `core` to `cli`. The direct consequence of this change is that now the `core` module is leaner and it has not node only dependencies. Moreover, we can better serve users that want to use node-wot as a simple library (embedded in their applications) cause in 99% of cases they don't need the scripting runtime support. 

Fix #377

### DIFF
--- a/packages/cli/test/RuntimeTest.ts
+++ b/packages/cli/test/RuntimeTest.ts
@@ -22,26 +22,30 @@
 
 import { suite, test } from "@testdeck/mocha";
 import { should, assert } from "chai";
+import DefaultServient from "../src/cli-default-servient";
 
-import Servient from "../src/servient";
 import fs from "fs";
 // should must be called to augment all variables
 should();
 
-@suite("the runtime of servient")
+@suite("Test suite for script runtime")
 class WoTRuntimeTest {
-    static servient: Servient;
+    static servient: DefaultServient;
 
     static WoT: typeof WoT;
 
     exit: (code?: number) => never;
 
-    static before() {
+    static async before() {
         // We need to disable this check for killing servient logs
-        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        /* eslint-disable @typescript-eslint/no-empty-function */
         console.error = () => {};
-        this.servient = new Servient();
-        console.log("before starting test suite");
+        console.debug = () => {};
+        console.warn = () => {};
+        console.info = () => {};
+        /* eslint-enable @typescript-eslint/no-empty-function */
+        this.servient = new DefaultServient(true);
+        await this.servient.start();
     }
 
     beforeEach() {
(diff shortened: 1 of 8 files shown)


packages/cli/test/RuntimeTest.ts
```diff
@@ -22,26 +22,30 @@
...
 
     static WoT: typeof WoT;
 
     exit: (code?: number) => never;
 
-    static before() {
+    static async before() {
         // We need to disable this check for killing servient logs
-        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        /* eslint-disable @typescript-eslint/no-empty-function */
```
danielpeintner, line 41: Can this disable comment be removed?

packages/cli/test/RuntimeTest.ts
```diff
@@ -22,26 +22,30 @@
...
         console.error = () => {};
-        this.servient = new Servient();
-        console.log("before starting test suite");
+        console.debug = () => {};
+        console.warn = () => {};
+        console.info = () => {};
+        /* eslint-enable @typescript-eslint/no-empty-function */
```
danielpeintner, line 46: Maybe not, since it just uses another comment style */ vs //

relu91: We have two choices:
- `console.debug = () => { /* nop */ }`
- `/* eslint-disable @typescript-eslint/no-empty-function */` with the corresponding `/* eslint-enable @typescript-eslint/no-empty-function */` 

Personally, I found the first a workaround as it does not clearly explain why we are adding a comment in an empty function.

relu91: we can also change a little bit [eslint configuration](https://eslint.org/docs/rules/no-empty-function#:~:text=Rule%20Details,if%20it%20contains%20a%20comment.&text=Examples%20of%20correct%20code%20for,)%20%7B%20%2F%2F%20do%20nothing.%20%7D) to allow empty arrow functions.

danielpeintner: I am fine as is. I was just wondering.

danielpeintner (approved):
LGTM

I also did a quick check locally with the CLI and the browser samples 👍

Issue #680, opened 2022-02-21
Thing Model - Next steps


danielpeintner:
based on discussions in https://github.com/eclipse/thingweb.node-wot/pull/674 a collection of ideas/plans

* move Thing Model code to `td-tools`
  * use resolver for  `Servient` dependency?
  * rename `td-tools` to `wot-tools` (DP: Not sure if this is needed)
* use https://www.npmjs.com/package/json-placeholder-replacer for placeholder code ?
* test composed types (e.g., objects and/or arrays) in tests for placeholder

relu91:
I would add another point:
- Add `implements(td: ThingDescription, tm: ThingModel): boolean` function to check if a `td` correctly implements a `tm`.

danielpeintner:
based on https://github.com/eclipse/thingweb.node-wot/pull/719#issuecomment-1080477578 I added Todo: test thing-model-helpers in a browser

relu91:
I changed the points to checkboxes and checked the first point 👍🏻

PR #725, opened 2022-03-29
Thing model

lukesmolo:
This should implement the second and third points of the todo list for issue #680

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6194,6 +6194,15 @@
             "dev": true,
             "license": "MIT"
         },
+        "node_modules/json-placeholder-replacer": {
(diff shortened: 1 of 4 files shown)


relu91 (changes requested):
PR looks good but it seems that package-lock wasn't updated. Moreover, I'm curious what happens when you try to fill with an array something like this: "hello {{variable}}". placeHolderReplacer throws? I think we should have a test for this condition.

relu91 (approved):
Talked with @lukesmolo offline. I've added a small test just to be sure if we ever change the placeholder library we keep consistency. When CI is green we can merge!

PR #726, opened 2022-03-29
CLI: pass port as argument

### LINKED ISSUE #117: Problem creating 2 wot servers (nice to have)
when creating two servers after each other, 2nd one gives problem saying port is used,
I know it can be done by changing default port 8080 to sth else and creating 2nd one or creating it at the same time with 2 files in argument
But it will be nice to have having an argument like "--httpport 8081" or in small version "-hp 8081" to assign port when creating the server 
example command: wotservient sth.js -hp 8081

relu91:
This PR introduces a new cmd line option to pass the port and override default configurations (or the config file). Furthermore, it improves a little bit the `cli.ts` style code. 

Fix #117

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -288,6 +288,7 @@ Options:
   -f,   --configfile <file>         load configuration from specified file
   -h,   --help                      show this help
         --[protocol].port           specify the port to expose the server for a specific protocol
+                                    (examples: --http.port 8888, --coap.port 3333)
 
 wot-servient.conf.json syntax:
 {
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -71,7 +71,7 @@ const overrideConfig = function (conf: any): Promise<unknown> {
         conf = conf ?? {};
 
         servientPorts.forEach((port, protocol) => {
-            if (port !== 0) {
+            if (port !== undefined) {
                 if (!(protocol in conf)) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli.ts
```diff
@@ -246,16 +277,17 @@ If one or more SCRIPT is given, these files are loaded instead of the directory.
...
+  -c,   --clientonly                do not start any servers
+                                    (enables multiple instances without port conflicts)
+  -cp,  --compiler <module>         load module as a compiler
+                                    (The module must export a create function which returns
                                     an object with a compile method)
-  -f,  --configfile <file>         load configuration from specified file
-  -h,  --help                      show this help
+  -f,   --configfile <file>         load configuration from specified file
+  -h,   --help                      show this help
+        --[protocol].port           specify the port to expose the server for a specific protocol
```
danielpeintner, line 290: Shall we add an example in the description like the following

`--http.port 8081`

At first it wasn't very clear to me how to pass the argument
I had to look at the regex 🙃

danielpeintner (approved):
Great addition 👍

As mentioned below .. maybe a sample might be useful

packages/cli/src/cli.ts
relu91: Good point! I will add an example!

relu91: Done ✔

danielpeintner (approved):

PR #727, opened 2022-03-31
chore: increase version to 8.0.1

danielpeintner:


danielpeintner:
@relu91 others
shall we merge this PR "as is" and integrate the [CoAP](https://github.com/eclipse/thingweb.node-wot/issues/728) and [OPC-UA](https://github.com/eclipse/thingweb.node-wot/issues/729) fixes in a follow-up PR?

danielpeintner:
> I would not merge this PR without fixing the 2 issues first, commenting tests out is not a good idea;

@erossignon Thanks for the fix!

I do have some questions now
* how could that happen?
  - shouldn't minor version changes be backwards compatible?
  - why does `npm install` not refresh all dependencies (needs to remove `package-lock.json` first?) 
* how can we prevent this from happening again (create a *fresh* `package-lock.json` file from time to time?)
* might users of node-wot encounter issues with the recent release `v0.8.0`? Shall we release a bug-fix release soon?

relu91:
> shouldn't minor version changes be backwards compatible?

IMHO, they should but not everybody follows this convention 😢 

> why does npm install not refresh all dependencies (needs to remove package-lock.json first?)

Maybe [npm-update](https://docs.npmjs.com/cli/v7/commands/npm-update)  what we are looking for. 

> how can we prevent this from happening again (create a fresh package-lock.json file from time to time?)

my takeaway from this experience is that _before_ release we should update all the dependencies to the latest version. In my understanding dependants libraries will not follow the package-lock constraints and if they install node-wot right after the release they will download the latest version of our dependencies. 

> might users of node-wot encounter issues with the recent release v0.8.0? Shall we release a bug-fix release soon?
 
 I did some tests locally and I didn't encounter any problems. But I'm not against to have more frequent releases 👍🏻

relu91 (approved):

danielpeintner:
I suggest to wait for https://github.com/mcollina/node-coap/pull/333 to land. This would bring as back to "normal" again

danielpeintner:
Note: since we are introducing some *breaking* changes in https://github.com/eclipse/thingweb.node-wot/pull/731 @relu91 is suggesting to jump to v0.9.0
see https://github.com/eclipse/thingweb.node-wot/pull/731#discussion_r844984781

JKRhb:
https://github.com/mcollina/node-coap/pull/333 has now been merged :) Hope everything is now working again.

danielpeintner:
I reverted https://github.com/eclipse/thingweb.node-wot/commit/897d3dd32f6d6959a159d79ef9da8fe7e4198210 and used the latest coap version with new issues.

see https://github.com/eclipse/thingweb.node-wot/issues/728#issuecomment-1102303593

Hence, I suggest to merge this PR as is and go on with CoAP issues in https://github.com/eclipse/thingweb.node-wot/issues/728

JKRhb:
> I reverted [897d3dd](https://github.com/eclipse/thingweb.node-wot/commit/897d3dd32f6d6959a159d79ef9da8fe7e4198210) and used the latest coap version with new issues.
> 
> see [#728 (comment)](https://github.com/eclipse/thingweb.node-wot/issues/728#issuecomment-1102303593)
> 
> Hence, I suggest to merge this PR as is and go on with CoAP issues in #728

That sounds good :) I am going to investigate the new (pretty strange) issue in the meantime

PR #731, opened 2022-04-04
refactor: remove obsolete observe/subscribe code

danielpeintner:


danielpeintner:
@relu91 is there anything that prevents us from removing these old code parts ?

packages/core/src/exposed-thing.ts
```diff
@@ -316,18 +141,30 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
-            this.events[eventName] = newEvent;
+            const act: TD.ThingAction = this.actions[actionName];
+            if (act.safe === undefined || typeof act.safe !== "boolean") {
+                act.safe = false;
+            }
+            if (act.idempotent === undefined || typeof act.idempotent !== "boolean") {
+                act.idempotent = false;
+            }
         }
+        // event defaults -> none
```
relu91, line 167: Shouldn't we handle defaults in `td-parse` ?

packages/core/src/exposed-thing.ts
```diff
@@ -635,331 +465,6 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
-            if (this.events[name]) {
-                // let sub: Subject<any> = this.events[name].getState().subject;
-                // sub.unsubscribe(); // XXX causes loop issue (see browser counter example)
-                console.debug("[core/exposed-thing]", `ExposedThing '${this.title}' unsubscribes from event '${name}'`);
-            } else {
-                reject(new Error(`ExposedThing '${this.title}', no event found for '${name}'`));
-            }
-        });
-    }
-
```
relu91, line 962: These methods were all public. In not against removing them, but in practice this change causes a public API change which means that the next version will be a 0.9.x if we care about semver.

relu91 (approved):
I think we can safely remove those obsolete methods. I left some minor comments below

packages/core/src/exposed-thing.ts
danielpeintner: Mhh, good point... internally they were net referenced anymore... and incrementing to 0.9. seems like an overkill to me 🙃

relu91: yeah I know, but it is just a number 😄 I think we should start issuing releases more often. Take for example the caop library. If they had some release process like us we would still have a lot of bugs.

packages/core/src/exposed-thing.ts
relu91: > To be honest... I think we can safely remove them.. but wasn't sure what to do.
Shall we remove it.. and look into it again IF we run into any issue?

+1

packages/core/src/exposed-thing.ts
danielpeintner: I see your point. Let's discuss this in https://github.com/eclipse/thingweb.node-wot/pull/727#issuecomment-1092634350

packages/core/src/exposed-thing.ts
danielpeintner: @relu91 please have a look at the additional commit

relu91 (approved):

PR #732, opened 2022-04-05
Use *new* statistics links

danielpeintner:
old link/domain does not work any longer


egekorkan:
By the way, I get the following from shields.io, not sure why we were doing the complicated way npm

danielpeintner:
It is not about the image.. the shields.io image is still fine.
It's is about the link when you click on it.

relu91 (approved):

Issue #177, opened 2020-02-25
Unusual behavior when configuring HTTPS from config file


egekorkan:
Summary: We need strict validation of the security-related configuration in the config file passed to wot-servient CLI. The code expects a very strict config file where empty or wrong fields are not filled by defaults or any error reported, even with log level equal to debug (even for info level, there should be error messages).

Long version:
I want to have basic auth on an HTTP Thing that I instantiate via wot-servient CLI and by supplying a config file. I write the following config files and observe the following behavior. Only the relevant parts of the config file are shown:

- Forgetting `serverKey` or `serverCert`
```json
"http":{
        "port":8080,
        "serverKey":"/home/eko/git/thingweb.node-wot/examples/security/privatekey.pem"
    }
```
An HTTP server is started, without giving any information. This happens due to [this line](https://github.com/eclipse/thingweb.node-wot/blob/350fcf7cdf18e8e983cbaee8312c0240bce344a3/packages/binding-http/src/http-server.ts#L67).

- Forgetting `security`
```json
    "http":{
        "port":8080,
        "serverKey":"/home/eko/git/thingweb.node-wot/examples/security/privatekey.pem",
        "serverCert":"/home/eko/git/thingweb.node-wot/examples/security/certificate.pem"
    }
```
HTTPS server is instantiated but no need to use username password. This is intended but again no information is given.

- No credentials or wrong id
```jsonc
    "http":{
        "port":8080,
        "serverKey":"/home/eko/git/thingweb.node-wot/examples/security/privatekey.pem",
        "serverCert":"/home/eko/git/thingweb.node-wot/examples/security/certificate.pem"
    },
    "credentials": { //this whole block can be removed for the same behavior
        "urn:myThing": { //id in the counter.js is set to urn:uuid:counter
            "username": "node-wot", // username key (not the value) can be changed to another string for the same behavior
            "password": "hello" // password key (not the value) can be changed to another string for the same behavior
        }
    }
```
This is the most obscure case since:
1. Credentials object must be added
2. The id of thing from the script must be copy pasted. If the script does not have id, which is optional, a random id is generated.
3. `username` and `password` keys must be put correctly and not `identity` and `psk`. So the programmer must know what keys should be for the given security scheme above.

If all these and also the ones above, are not done perfectly, a Thing is created with an unknown username-password.

danielpeintner:
[...]

PR #733, opened 2022-04-05
feat: cli to accept servient config through file/.env/cli-arguments

fillobotto:
Under the advice of @relu91, I worked on improving the CLI package.
The reason of this PR is mainly to allow CLI to accept servient configuration parameters in three different ways (in order of priority):
- Parameters passed as CLI arguments
- Dotenv (.env) file
- Physical JSON configuration file

Also, the old cli arguments parsing method has been replaced and now it's using [commander.js](https://github.com/tj/commander.js) library.

### How to pass parameters as CLI arguments
Using CLI in this way
`cli.js script-name.js -p http.port:=8080 mqtt.port:=8081`
is equivalent of passing a JSON config file made in the following way
`{ http: { port: 8080 },  mqtt: { port: 8081 } }`


packages/cli/src/cli.ts
```diff
@@ -21,48 +21,215 @@ import DefaultServient from "./cli-default-servient";
...
+  CLIENTONLY      : boolean setting if no servers shall be started (default=false)
+  STATIC          : string with hostname or IP literal for static address config
+  RUNSCRIPT       : boolean to activate the 'runScript' Action (default=false)
+  HPORT           : integer defining the HTTP listening port
+  PROXY           : object with "href" field for the proxy URI,
+                                "scheme" field for either "basic" or "bearer", and
+                                corresponding credential fields as defined below
+  ALLOW           : boolean whether self-signed certificates should be allowed
+  BROKER-URL      : URL to an MQTT broker that publisher and subscribers will use
+  BROKER-UNIQUEID : unique id set by mqtt client while connecting to broker
```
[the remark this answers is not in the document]
egekorkan: Also what type? I see string in the schema

packages/cli/src/cli.ts
```diff
@@ -21,68 +21,216 @@ import DefaultServient from "./cli-default-servient";
...
+        THING_ID2: {
+            "username": USERNAME,
+            "password": PASSWORD
+        }
+    }
+}
 
+wot-servient.conf.json fields:
+  ---------------------------------------------------------------------------
+  All entries in the config file structure are optional
```
egekorkan, line 95: this contradicts with the schema

packages/cli/src/wot-servient-schema.conf.json
```diff
@@ -0,0 +1,99 @@
...
+        "http": {
+            "$id": "#root/http",
+            "type": "object",
+            "properties": {
+                "port": {
+                    "$id": "#root/http/port",
+                    "type": "integer"
+                },
+                "proxy": {
+                    "$id": "#root/http/proxy",
```
egekorkan, line 36: this can be made specific I think

packages/cli/src/wot-servient-schema.conf.json
```diff
@@ -0,0 +1,99 @@
...
+                    "type": "object",
+                    "properties": {}
+                },
+                "allowSelfSigned": {
```
egekorkan, line 40: serverKey and serverCert are missing

egekorkan: and they have a cross dependency

egekorkan (changes requested):
Thank you first of all for the schema! This was mentioned by me quite some time ago at https://github.com/eclipse/thingweb.node-wot/issues/177 but I have definitely forgotten about it (please close that issue and [this](https://github.com/eclipse/thingweb.node-wot/issues/353) afterwards btw)

I have some change requests. Most are small things to fix. However, given the issues I have pointed out in #177, it would be good to test the config file as well.

It might be worth it to check https://github.com/eclipse/thingweb.node-wot/issues/214 can be solved here.

fillobotto:
Based on your suggestions, I managed to make the validation schema more accurate

packages/cli/src/wot-servient-schema.conf.json
egekorkan: I sadly do not have one but [here](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-http/src/http-server.ts#L98) you can see that they are both needed if one is needed

relu91:
What concerns me is that the CI is failing cause it cannot find the package-lock. I'll try to understand what's going on...

relu91:
I fixed the issue with the CI, but sadly I introduced commit dfb7555 in the history. I hope is not a big deal. I also did some minor changes and improvements to the help text:
```
Usage: wot-servient [options] [files...]


Run a WoT Servient in the current directory.


Arguments:
  files                             script files to execute. If no script is given, all .js files in the current directory are loaded. If one or more 
                                    script is given, these files are loaded instead of the directory.

Options:
  -v, --version                     display node-wot version
  -i, --inspect [host]:[port]       activate inspector on host:port (default: 127.0.0.1:9229)
  -ib, --inspect-brk [host]:[port]  activate inspector on host:port (default: 127.0.0.1:9229)
  -c, --client-only                 do not start any servers (enables multiple instances without port conflicts)
  -cp, --compiler <module>          load module as a compiler
  -f, --config-file <file>          load configuration from specified file (default: "wot-servient.conf.json")
  -p, --config-params <param...>    override configuration paramters [key1:=value1 key2:=value2 ...] (e.g. http.port=8080)
  -h, --help                        show this help

wot-servient.conf.json syntax:
{
    "servient": {
        "clientOnly": CLIENTONLY,
        "staticAddress": STATIC,
        "scriptAction": RUNSCRIPT
    },
    "http": {
        "port": HPORT,
        "proxy": PROXY,
        "allowSelfSigned": ALLOW
    },
    "mqtt" : {
        "broker": BROKER-URL,
        "username": BROKER-USERNAME,
        "password": BROKER-PASSWORD,
        "clientId": BROKER-UNIQUEID,
        "protocolVersion": MQTT_VERSION
    },
    "credentials": {
        THING_ID1: {
            "token": TOKEN
        },
        THING_ID2: {
            "username": USERNAME,
            "password": PASSWORD
        }
    }
}

wot-servient.conf.json fields:
  CLIENTONLY      : boolean setting if no servers shall be started (default=false)
  STATIC          : string with hostname or IP literal for static address config
  RUNSCRIPT       : boolean to activate the 'runScript' Action (default=false)
  HPORT           : integer defining the HTTP listening port
  PROXY           : object with "href" field for the proxy URI,
                                "scheme" field for either "basic" or "bearer", and
                                corresponding credential fields as defined below
  ALLOW           : boolean whether self-signed certificates should be allowed
  BROKER-URL      : URL to an MQTT broker that publisher and subscribers will use
  BROKER-UNIQUEID : unique id set by MQTT client while connecting to the broker
  MQTT_VERSION    : number indicating the MQTT protocol version to be used (3, 4, or 5)
  THING_IDx       : string with TD "id" for which credentials should be configured
  TOKEN           : string for providing a Bearer token
  USERNAME        : string for providing a Basic Auth username
  PASSWORD        : string for providing a Basic Auth password
  ---------------------------------------------------------------------------

Environment variables must be provided in a .env file in the current working directory.

Example:
VAR1=Value1
VAR2=Value2
```

relu91:
I have a couple of ideas about to further improving the code and functionalities (I got a few bugs while trying) but I'd like to tackle them in another PR.

packages/cli/src/cli.ts
```diff
@@ -21,68 +21,215 @@ import DefaultServient from "./cli-default-servient";
...
+    .option("-cp, --compiler <module>", "load module as a compiler")
+    .option(
+        "-f, --config-file <file>",
+        "load configuration from specified file",
+        parseConfigFile,
+        "wot-servient.conf.json"
+    )
+    .option(
+        "-p, --config-params <param...>",
+        "override configuration paramters [key1:=value1 key2:=value2 ...] (e.g. http.port=8080)",
```
fillobotto, line 179: @relu91 can you also update this line? because I previously used this notation `key1:=value1` but then the example right after is `http.port=8080` instead of `http.port:=8080`

danielpeintner:
How to you plan to move on with this PR? Ready to be reviewed / merged.
Improvements done in follow-up PRs like mentioned by @relu91

relu91:
yeah sorry I forgot to merge this :) let's move on!

PR #734, opened 2022-04-06
fix: use proper handleUnobserveProperty call instead of event

danielpeintner:
seems like a copy & paste error


relu91 (approved):
Right! 👍🏻

PR #735, opened 2022-04-06
added tuya binding with security and codec management

SeppiaBrilla:
Hello everyone,
I've developed a "binding" (codec + credential class) that allows Node-Wot to interact with Tuya's smart things.  
The binding is only client-side and it needs a slightly custom TD in order to work: a custom field ("tuya:PropertyName") in the schema that corresponds to the name used by tuya Apis to identify the property and two different forms for reading and writing (with an Url formed this way: https://openapi.{tuyaRegion}.com/v1.0/iot-03/devices/{deviceID} as a base Uri + /status (read) + /commands (write)).
The credential class implements the custom security protocol used by tuya Apis, it is a bit complicated and it needs a lot of data: it uses a secret key to sign a header field composed of the public key from tuya, a token, the current time, the method an ashed body and the other headers. [Tuya documentation](https://developer.tuya.com/en/docs/cloud)
I hope this can help, 
Alessio

### DIFF
--- a/packages/binding-http/src/codecs/tuya-codec.ts
+++ b/packages/binding-http/src/codecs/tuya-codec.ts
@@ -1,0 +1,48 @@
+/********************************************************************************
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the W3C Software Notice and
+ * Document License (2015-05-13) which is available at
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+
+import { ContentCodec } from "@node-wot/core";
+import * as TD from "@node-wot/td-tools";
+import { DataSchemaValue } from "wot-typescript-definitions";
+
+export default class HttpTuyaCodec implements ContentCodec {
+    getMediaType(): string {
+        return "application/json+tuya";
+    }
+
+    bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
+        const parsedBody = JSON.parse(bytes.toString());
+        if (!parsedBody.success) throw new Error(parsedBody.msg ? parsedBody.msg : JSON.stringify(parsedBody));
+        for (const key in parsedBody.result) {
+            if (parsedBody.result[key].code === schema["tuya:PropertyName"]) {
+                return parsedBody.result[key].value;
+            }
+        }
+        throw new Error("Property not found");
+    }
+
+    valueToBytes(value: unknown, schema: TD.DataSchema, parameters?: { [key: string]: string }): Buffer {
+        const obj = {
+            commands: [
+                {
+                    code: schema["tuya:PropertyName"],
+                    value: value,
+                },
+            ],
+        };
+        const body = JSON.stringify(obj);
+        return Buffer.from(body);
+    }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -3606,6 +3606,14 @@
                 "url": "https://github.com/sponsors/sindresorhus"
             }
         },
+        "node_modules/decode-uri-component": {
+            "version": "0.2.0",
+            "resolved": "https://registry.npmjs.org/decode-uri-component/-/decode-uri-component-0.2.0.tgz",
+            "integrity": "sha1-6zkTMzRYd1y4TNGh+uBiEGu4dUU=",
+            "engines": {
+                "node": ">=0.10"
+            }
+        },
         "node_modules/decompress-response": {
             "version": "3.3.0",
             "resolved": "https://registry.npmjs.org/decompress-response/-/decompress-response-3.3.0.tgz",
@@ -5288,6 +5296,14 @@
             },
             "engines": {
                 "node": ">=8"
+            }
+        },
+        "node_modules/filter-obj": {
+            "version": "1.1.0",
+            "resolved": "https://registry.npmjs.org/filter-obj/-/filter-obj-1.1.0.tgz",
+            "integrity": "sha1-mzERErxsYSehbgFsbF1/GeCAXFs=",
+            "engines": {
+                "node": ">=0.10.0"
             }
         },
         "node_modules/finalhandler": {
@@ -10174,6 +10190,23 @@
                 "url": "https://github.com/sponsors/ljharb"
             }
         },
+        "node_modules/query-string": {
+            "version": "7.1.1",
+            "resolved": "https://registry.npmjs.org/query-string/-/query-string-7.1.1.tgz",
+            "integrity": "sha512-MplouLRDHBZSG9z7fpuAAcI7aAYjDLhtsiVZsevsfaHWDS2IDdORKbSd1kWUA+V4zyva/HZoSfpwnYMMQDhb0w==",
+            "dependencies": {
+                "decode-uri-component": "^0.2.0",
+                "filter-obj": "^1.1.0",
+                "split-on-first": "^1.0.0",
+                "strict-uri-encode": "^2.0.0"
+            },
+            "engines": {
+                "node": ">=6"
+            },
+            "funding": {
+                "url": "https://github.com/sponsors/sindresorhus"
+            }
+        },
         "node_modules/querystring": {
             "version": "0.2.0",
             "resolved": "https://registry.npmjs.org/querystring/-/querystring-0.2.0.tgz",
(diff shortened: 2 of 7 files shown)


egekorkan:
Thanks for the contribution! Please make sure that you sign the Eclipse agreement (see https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md). Also, some changes to the readme of HTTP binding would be nice to show how this is used. I did not do a real review but some quick comments:
- Custom security scheme name is a bit tricky since TD spec sort of does not allow that. Is there no way to describe it with the current bearer scheme? 
- I did not see a TD yet but if you are adding protocol-specific information outside of the forms, it is a bad design choice for TDs. You wrote a change to the schema so that is what I have assumed.

SeppiaBrilla:
Hi, 

- It is not possible to use any other method since it is a custom security scheme, I called it a "Custom bearer" but in reality the token is just a small part of the needed data: is also necessary a private key assigned by Tuya on their platform and also to sign the entire request (body included) in the "sign" field. I've tried to squeeze it into some other method but I haven't found a way.
- I need a custom field in the Dataschema passed to the bytesToValue and valueToBytes functions. I've used the same method used by the opcua codec.

P.s. sorry, I forgot to sign the agreement. It should be signed now

JKRhb:
> * Custom security scheme name is a bit tricky since TD spec sort of does not allow that. Is there no way to describe it with the current bearer scheme?

There is an example for a custom security scheme in the TD specification, though: https://w3c.github.io/wot-thing-description/#adding-security-schemes I guess you would technically need to prefix it (and add a context extension) for it to be a valid scheme?

egekorkan:
In theory that's how you'll do new schemes. Just that the json schema for the td doesn't allow such an extension now, so some consumers will not accept it

SeppiaBrilla:
> > * Custom security scheme name is a bit tricky since TD spec sort of does not allow that. Is there no way to describe it with the current bearer scheme?
> 
> There is an example for a custom security scheme in the TD specification, though: https://w3c.github.io/wot-thing-description/#adding-security-schemes I guess you would technically need to prefix it (and add a context extension) for it to be a valid scheme?

so if I change the securityscheme in something like 
`TuyaCustomBearerSecurityScheme extends SecurityScheme {
    scheme: "Tuya:CustomBearer";
    key: string;
    secret: string;
    baseUri: string;
}`
It should be possible to add a "Tuya" value in the context and it should work, am I right?


> In theory that's how you'll do new schemes. Just that the json schema for the td doesn't allow such an extension now, so some consumers will not accept it

Well... Since this binding is only client-side it would work only on node-Wot and any other consumer won't accept it either way or am I missing something?

packages/binding-http/src/codecs/https-codecs.ts
```diff
@@ -0,0 +1,51 @@
...
+                  code:schema["tuya:PropertyName"],
+                  value: value
+              }
+          ]
+        };
+        const body = JSON.stringify(obj);
+        return Buffer.from(body);
+    }
+
+}
```
relu91, line 51: Can you rename this file to: `tuya-codec`?

packages/binding-http/src/codecs/https-codecs.ts
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+
+ import { ContentCodec } from "@node-wot/core";
+ import * as TD from "@node-wot/td-tools";
+ import { DataSchemaValue } from "wot-typescript-definitions";
+
+ export default class HttpTuyaCodec implements ContentCodec {
+    getMediaType(): string {
+        return "application/json+Tuya";
```
relu91, line 22:
```suggestion
        return "application/json+tuya";
```

packages/binding-http/src/codecs/https-codecs.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
+        let parsedBdoy;
+        parsedBdoy = JSON.parse(bytes.toString());
+        if(!parsedBdoy.success) throw new Error(parsedBdoy.msg);
```
relu91, line 28: what happens if the payload is invalid? like if someone is returning { "hello" : "world"} the error will contain un `undefined` message. Can we improve the error message in such cases?

packages/binding-http/src/codecs/https-codecs.ts
```diff
@@ -0,0 +1,51 @@
...
+ import { DataSchemaValue } from "wot-typescript-definitions";
+
+ export default class HttpTuyaCodec implements ContentCodec {
+    getMediaType(): string {
+        return "application/json+Tuya";
+    }
+
+    bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
+        let parsedBdoy;
+        parsedBdoy = JSON.parse(bytes.toString());
```
relu91, line 27:
```suggestion
        parsedBody = JSON.parse(bytes.toString());
```
`Bdoy` ? do you mean Body?

packages/td-tools/src/thing-description.ts
```diff
@@ -278,6 +278,13 @@ export interface OAuth2SecurityScheme extends SecurityScheme {
     scopes?: Array<string>;
 }
 
+export interface TuyaCustomBearerSecurityScheme extends SecurityScheme{
+    scheme: "TuyaCustomBearer";
+    key:string;
+    secret:string;
+    baseUri:string;
+}
+
```
relu91, line 287: I think you mixed up the concepts of `SecurityScheme` and `Credentials`. The `SecurityScheme` contains only information about the Security protocol. In other words, it tells consumers WHAT to do to correctly authenticate with the remote WebThing. On the other hand, `Credentials` contains sensitive information needed by the protocol to successfully authenticate. For example, `username` and `password`. Remember also that the `SecurityScheme` is contained in the TD and therefore is public. I should never put secrets inside a TD. 

Therefore, I suggest moving `key` and `secret` inside a credential object.

relu91 (changes requested):
Sorry for the previous comments I mistakenly clicked on add review comment instead of creating review 😄 . 

First of all, @SeppiaBrilla for your efforts! I think this is a valuable contribution to node-wot and to the standard. I left some minor change requests can you please cover them ?

Besides that, I think the implementation is good enough for what we have now in the spec. However, I'll open issues to address some of the concerns raised during the review:
- One issue in the TD to discuss SecurityScheme extensions in the JSON schema
- Discuss in the security task force how the Tuya security scheme can be mapped. I found the [official documentation](https://developer.tuya.com/en/docs/iot/singnature?id=Ka43a5mtx1gsc) and it seems to me that we might be able to cover this special case with the combination of `alg` and `format` parameters of the `BearerSecurityScheme`. I am not sure though and I want to discuss it with the experts.

danielpeintner:
> P.s. sorry, I forgot to sign the agreement. It should be signed now

ECA is approved now. Thanks!

packages/binding-http/src/codecs/https-codecs.ts
SeppiaBrilla: Since the msg field is mandatory in case something went wrong (success is false) I haven't considered such a path. However, I think I could return the entire body (stringified) in that case.

SeppiaBrilla:
Hi, 
Thanks for your feedback, your time and sorry for the many changes required (it's my first PR in an open-source project).
I should have fixed all the problems with my latest commit. Please let me know if there's anything else I should address.

relu91 (approved):
Looks good, couple of things that I noticed:
- There are some unwanted changes to log messages 
- Too many unwanted changes on the package-lock.json

I would go over and merge it but only after I will fix the package-lock.json 👍🏻

packages/td-tools/src/thing-description.ts
```diff
@@ -278,6 +278,11 @@ export interface OAuth2SecurityScheme extends SecurityScheme {
     scopes?: Array<string>;
 }
 
+export interface TuyaCustomBearerSecurityScheme extends SecurityScheme {
```
danielpeintner, line 281: Should we add a note here "why" it is here / needed

danielpeintner: Can we move it into the http binding package ?

packages/binding-http/src/codecs/tuya-codec.ts
```diff
@@ -0,0 +1,51 @@
+/********************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
```
danielpeintner, line 2: nit: years need to be updated

SeppiaBrilla:
Hi, I've moved the "TuyaCustomBearerSecurityScheme" in the http.ts file along with a brief comment that explains why is it needed and I've fixed the year in the codec file

danielpeintner (approved):
LGTM.

Can you resolve the conflict for package-lock.json

SeppiaBrilla:
Hi, 
I've minimized the change on the package-lock file to the minimum I could (the two packages I've added are the new ones I need) but still result in a conflict. Since I don't have write access to the repository I can not see what the problem is and fix it.

danielpeintner:
I suggest to merge the current master branch into your branch. 
Resolving merge conflicts with package-lock might be difficult but I think marking it as resolved... remove the file and re-creating it with `npm install` should do the job..

@relu91  can you support?

relu91:
I'm currently at a conference, but I will resolve the conflicts myself when I'll have a couple of minutes to look into this! Don't worry 👍🏻

relu91:
Should be ready to go. What I did:
- fixed package-lock
- fixed eslint errors
- squashed the commit history in on simple commit

PR #739, opened 2022-04-19
chore(binding-opcua): update packages

erossignon:


danielpeintner (changes requested):
PR itself looks good. 

We just have a conflict since I merged another PR with increasing the version number..

danielpeintner (approved):

PR #742, opened 2022-04-20
Fix CoapClientFactory constructor, clean up examples

JKRhb:
This PR fixes an issue raised in https://github.com/eclipse/thingweb.node-wot/issues/741. To make testing the CoAP examples easier, I included them as actual JavaScript files while also cleaning them up a bit more.


relu91 (changes requested):
We have a dedicated examples project for these cases. I would keep this consistent with the current structure.

JKRhb:
I now moved the examples to the dedicated project. The only downside I see here is that it is a bit more difficult to test/run them this way.

JKRhb:
Hmm. I also noticed that building the examples project is currently not possible, as it seems:

```
$ npm run build

> @node-wot/examples@0.8.1 build
> tsc -b

src/security/oauth/exposer.ts:38:17 - error TS2345: Argument of type '{ "@context": string; title: string; id: string; securityDefinitions: { oauth2_sc: { scheme: string; flow: string; authorization: string; token: string; scopes: string[]; }; }; security: string[]; actions: { ...; }; }' is not assignable to parameter of type '{ [x: string]: unknown; id?: string; title?: string; titles?: { [x: string]: string; }; properties?: { [x: string]: { [x: string]: unknown; "@type"?: TypeDeclaration; description?: string; descriptions?: { [x: string]: string; }; ... 22 more ...; required?: string[]; }; }; ... 15 more ...; "@context"?: [] | ... 1 mo...'.
  Types of property 'securityDefinitions' are incompatible.
    Type '{ oauth2_sc: { scheme: string; flow: string; authorization: string; token: string; scopes: string[]; }; }' is not assignable to type '{ [x: string]: { [x: string]: unknown; "@type"?: TypeDeclaration; description?: string; descriptions?: { [x: string]: string; }; proxy?: string; scheme?: "nosec"; } | { [x: string]: unknown; "@type"?: TypeDeclaration; ... 4 more ...; oneOf?: string[]; } | ... 6 more ... | { ...; }; }'.
      Property 'oauth2_sc' is incompatible with index signature.
        Type '{ scheme: string; flow: string; authorization: string; token: string; scopes: string[]; }' is not assignable to type '{ [x: string]: 
unknown; "@type"?: TypeDeclaration; description?: string; descriptions?: { [x: string]: string; }; proxy?: string; scheme?: "nosec"; } | { [x: string]: unknown; "@type"?: TypeDeclaration; ... 4 more ...; oneOf?: string[]; } | ... 6 more ... | { ...; }'.
          Type '{ scheme: string; flow: string; authorization: string; token: string; scopes: string[]; }' is not assignable to type '{ [x: string]: unknown; "@type"?: TypeDeclaration; description?: string; descriptions?: { [x: string]: string; }; proxy?: string; scheme?: "oauth2"; authorization?: string; token?: string; refresh?: string; scopes?: string | string[]; flow?: "code"; }'.
            Types of property 'scheme' are incompatible.
              Type 'string' is not assignable to type '"oauth2"'.

38     WoT.produce(td).then((thing) => {
                   ~~


Found 1 error.
```

danielpeintner:
If we (re-)introduce having binding examples.. I think we should do that for other "core" bindings as well..

i.e., add the sample scripts from HTTP etc.

What do you think?

relu91:
Weird I remember solving that issue a while ago, any clue why before your changes we didn't get that error?

JKRhb:
It also seemed familiar to me, maybe it was a different example, though? Or was it in the http-binding tests?

relu91:
> It also seemed familiar to me, maybe it was a different example, though? Or was it in the http-binding tests?

It should compile. [Client](https://github.com/w3c/wot-scripting-api/blob/main/typescript/thing-description/thing-description.d.ts#L473) is a valid value

JKRhb:
> It should compile. [Client](https://github.com/w3c/wot-scripting-api/blob/main/typescript/thing-description/thing-description.d.ts#L473) is a valid value

Oh, you're right, maybe I had an outdated version. It seems like the `ExposedThingInit` type annotation was missing, though. I updated the PR accordingly :)

danielpeintner (approved):
I am OK with the PR.

Adding http examples may be done in a follow-up PR

JKRhb:
@danielpeintner Do you think this one could also be merged? Or should @relu91 have another look?

relu91 (approved):
Let's go!

PR #744, opened 2022-04-23
chore(deps): bump coap to 1.0.8

### LINKED ISSUE #637: Counter client example hangs after execution
It seems that there is a resource leak in the counter example. My first impression is that CoAp client does not get properly cleaned up after the interaction with the remote thing. Maybe integrating https://github.com/mcollina/node-coap/pull/317 would help to solve the problem. 

### How to reproduce
Run `node packages/cli/dist/cli.js examples/scripts/counter.js` and in another shell `node packages/cli/dist/cli.js -c examples/scripts/counter-client.js`. The client application hangs without properly exiting.

### LINKED ISSUE #724: CoAP example hangs after first GET
see https://github.com/eclipse/thingweb.node-wot/pull/721#issuecomment-1080433488 

---------

Maybe I was not very clear... it hangs from the beginning on after the first GET.. NO result is reported.

Note1: I did try locally and then it reports values. It seems it has to do again with reaching out to the web again.
Note2: The PR https://github.com/eclipse/thingweb.node-wot/pull/720 is integrated. So this is not the issue. Maybe the Eclipse vserver?
Note3: Attached the test-file I am using.

```
// example-client.js
Servient = require("@node-wot/core").Servient;
CoapClientFactory = require("@node-wot/binding-coap").CoapClientFactory;

Helpers = require("@node-wot/core").Helpers;

servient.addClientFactory(new CoapClientFactory(null));

wotHelper
    .fetch("coap://plugfest.thingweb.io:5683/testthing") // coap://localhost:5683/testthing
    .then(async (td) => {
        // using await for serial execution (note 'async' in then() of fetch())
        try {
            const thing = await WoT.consume(td);

            // read property
            const read1 = await thing.readProperty("string");
            console.log("string value is: ", await read1.value());
        } catch (err) {
            console.error("Script error:", err);
        }
    })
    .catch((err) => {
        console.error("Fetch error:", err);
    });
```

### LINKED ISSUE #728: CoAP causes build issues
In https://github.com/eclipse/thingweb.node-wot/pull/727 I wanted to update all version numbers from `0.8.0` to `0.8.1`.

Doing so I removed `package-lock.json` file and recreated it calling `npm install`.
Unfortunately this caused issues in binding-coap.

I wonder also why we use out-dated libraries.
e.g., "node-coap-client": "1.0.8" vs "node-coap-client": "2.0.2"

@JKRhb do you have a clue why this issue is happening?

grafik

JKRhb:
Resolves #728, resolves #724. Probably also resolves #637?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -802,7 +802,6 @@
             "version": "5.0.2",
             "resolved": "https://registry.npmjs.org/@types/bl/-/bl-5.0.2.tgz",
             "integrity": "sha512-V4g3uJIfBHeDd/35QTPOujJ4+viJJVtNwC2LmBUZeXGSGL60R5iTsBEZ9Nh+wP3asMOA/LEFHxmKT6JzK+Vd0A==",
-            "dev": true,
             "dependencies": {
                 "@types/node": "*",
                 "@types/readable-stream": "*"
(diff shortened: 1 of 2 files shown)


JKRhb:
One check needs to be restarted

danielpeintner (approved):

PR #746, opened 2022-04-27
chore(binding-http): add HTTP examples from README as files

danielpeintner:
follow-up PR for https://github.com/eclipse/thingweb.node-wot/pull/742

### DIFF
--- a/packages/examples/src/bindings/http/example-client.ts
+++ b/packages/examples/src/bindings/http/example-client.ts
@@ -1,0 +1,42 @@
+/********************************************************************************
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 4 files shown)


danielpeintner:
Note: I noticed that `npm run build` for `packages/examples` is not run by `npm run build` in root folder (meaning that there can be code that does not compile).
Is there any reason for that? Shall we change that? I would say so...
@relu91 @JKRhb @erossignon

JKRhb:
I think there were some issues initially with the examples package that have been resolved in the meantime, so I would also be in favor of including the package in the build script :)

danielpeintner:
Since we got 👍's I added `packages/examples` again to build path. I think this PR is now ready for review...

danielpeintner:
Since we got 👍's I added `packages/examples` again to build path. I think this PR is now ready for review...

relu91 (approved):

PR #748, opened 2022-05-02
fix: value() might also return undefined if no value is expected

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/747

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -16,7 +16,7 @@ import * as util from "util";
 import * as WoT from "wot-typescript-definitions";
 import { ContentSerdes } from "./content-serdes";
 import { ProtocolHelpers } from "./core";
-import { DataSchemaError, NotReadableError, NotSupportedError } from "./errors";
+import { DataSchemaError, NotSupportedError } from "./errors";
 import { Content } from "./protocol-interfaces";
 import Ajv from "ajv";
 
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -63,9 +63,9 @@ export class InteractionOutput implements WoT.InteractionOutput {
         // the value has been already read?
         if (this.parsedValue) return this.parsedValue as T;
 
-        // is data schema valid?
+        // any data (schema)?
         if (!this.schema) {
-            throw new NotReadableError("Cannot read value: missing schema");
+            return undefined; // no value expected
         }
 
         const validate = ajv.compile<T>(this.schema);
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -63,35 +63,34 @@ export class InteractionOutput implements WoT.InteractionOutput {
         // the value has been already read?
         if (this.parsedValue) return this.parsedValue as T;
 
-        // any data (schema)?
-        if (!this.schema) {
-            return undefined; // no value expected
-        }
-
-        const validate = ajv.compile<T>(this.schema);
-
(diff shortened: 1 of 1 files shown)


egekorkan:
> Even if I agree that these changes are a solution for https://github.com/eclipse/thingweb.node-wot/issues/747, I am not convinced that assuming an undefined schema is equal to "no-content". Maybe I missed some discussion on the TD repository, do you reach a resolution on this problem? 
> 
> 
> 
> Why don't we try to parse the value and return what the protocol is giving us? does it create problems down the road? 

I agree with this approach as well. There can be different valid cases of not defining a schema:
- laziness of the Thing developer (which should penalise the Consumer developer any further by node-wot)
- the schema defined by `additionalResponses` that node-wot does not support yet (I think)
- The payload is not describable by a JSON Schema due to different payload formats
- Similar to the above, the payload is described by terms in the forms

danielpeintner:
Valid points. I updated the PR so that
* IF there is a schema it will be used to validate
* IF there is *no* schema it tries to report what the binding gives us (or `undefined`)

packages/core/src/interaction-output.ts
```diff
@@ -63,35 +63,34 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
         }
-        // read fully the stream
 
+        // read fully the stream
         const data = await ProtocolHelpers.readStreamFully(this.content.body);
         this.dataUsed = true;
         this.buffer = data;
+
         // call the contentToValue
         const value = ContentSerdes.get().contentToValue({ type: this.content.type, body: data }, this.schema);
```
relu91, line 78: One concern: if `this.schema` is nullable but in `contentToValue` it is not... how TS compiler accept this call without warning you?

relu91 (approved):
As I expressed below there might be some problems. For example, some binding uses the schema to parse the payload correctly... Anyhow, I think we can create an issue and fix it later!

packages/core/src/interaction-output.ts
danielpeintner: No TS compiler warning but I don't think the compiler can figure it out in this case..

We might want to add a note to the `ContentSerdes` interface methods a la 
"Note: schema parameter might be null/undefined"

https://github.com/eclipse/thingweb.node-wot/blob/3803354edafd9def2f90553f7ce9ef2eeba37ab0/packages/core/src/content-serdes.ts#L29-L30

The alternative would be to change the signature to mark schema as optional.

`bytesToValue(bytes: Buffer, schema?: DataSchema, parameters? ...)`

Not sure though whether changing the interface (at this point in time) makes sense. Any proposal? 

Question: What is the *best* way to go
1. Just add a note without changing the signature
2. change signature of interface
3. Do 2. it in a follow-up PR (with other breaking changes)

Note: I can check the current implementations (and update the PR accordingly) whether they properly deal with schema being null/undefined. There are some that fail like OctetstreamCodec (need to check others).

JKRhb: Making the schema optional sounds like a good idea, I also implemented it this way in `dart_wot` (see [here](https://github.com/namib-project/dart_wot/blob/812bddc568be58d4566acbbeb6f6b2cf66a93ad5/lib/src/core/content_serdes.dart#L211)). If no schema is passed, then the payload is simply returned and not validated.

danielpeintner: Makes sense. I created a new issue https://github.com/eclipse/thingweb.node-wot/issues/753 and I suggest we deal with it in a follow-up PR...

PR #750, opened 2022-05-05
Cleanup td definititions (v2)

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/529

replaces https://github.com/eclipse/thingweb.node-wot/pull/550

### DIFF
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -14,6 +14,7 @@
  ********************************************************************************/
 
 import * as WoT from "wot-typescript-definitions";
+import * as TDT from "wot-thing-description-types";
 
 import { Subject } from "rxjs/Subject";
 
@@ -39,8 +40,10 @@ import {
 } from "./protocol-interfaces";
 
 export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
-    security: Array<string>;
-    securityDefinitions: { [key: string]: TD.SecurityType };
+    security: string | [string, ...string[]];
+    securityDefinitions: {
+        [key: string]: TDT.SecurityScheme;
+    };
 
     id: string;
     title: string;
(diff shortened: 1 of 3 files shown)


relu91 (commented):
Seems fine overall but I want to test a couple of things. One question that I have is about keeping most of the types in td-tools. Shouldn't we just get rid of them? Do you have any particular reason for keeping them?

danielpeintner:
> One question that I have is about keeping most of the types in td-tools. Shouldn't we just get rid of them? Do you have any particular reason for keeping them?

My thinking was that some might be *breaking* changes. For example let's assume we no longer use `Form` as we used to do it might break things since its used in interfaces.
Hence, I was reluctant to change a lot ... and suggest to do that in a future 0.9.x release. For the time being I would go for a 0.8.1 bugfix release...

Do you feel we should apply the changes earlier (in this PR)?

packages/core/src/exposed-thing.ts
```diff
@@ -131,7 +134,8 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
                 }
             }
             if (!languageSet) {
-                arrayContext.push({
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                (arrayContext as any).push({
```
relu91, line 138: Did a couple of tests and indeed my suspects were confirmed. TS correctly warns you about something nasty happening in this line. Basically, `arrayContext` at this point might be [] (for the typings point of view). Now to correctly satisfy the type you would first push a W3CUrlContext string and then your `@language` but TS gets confused cause it can't really tell if the array is empty or it has already the W3CUrlContext string. Sadly, checking for a not-empty array don't solve the issue (either TS is not smart enough or I'm missing something). So if we are already sure that there's already a W3CUrlContext inside thing["@context"] we can solve the error like this:
```ts
(arrayContext as Exclude<typeof arrayContext, []>).push({
                    "@language": TD.DEFAULT_CONTEXT_LANGUAGE,
                });
```

This would solve the TS error and removes the need of a cast to any. On the other hand, I wonder: should we check that the array is not empty and then push the correct string inside? Or at this point we already checked that? Also is this addition still relevant? do we need a default lang tag?

relu91, line 138:
```suggestion
(arrayContext as Exclude<typeof arrayContext, []>).push({
                    "@language": TD.DEFAULT_CONTEXT_LANGUAGE,
                });
```

relu91 (changes requested):
I had a concern about "any" cansting and I have a couple of comments below. I think we should solve it in this PR.

packages/core/src/exposed-thing.ts
danielpeintner: Looks good. Thanks!

relu91 (approved):
Good to go now!

PR #752, opened 2022-05-11
Exposed template update 0.8.x

danielpeintner:
based on https://github.com/eclipse/thingweb.website/pull/36


danielpeintner:
@egekorkan @FadySalama please review.

I tried to simplify the README by removing all XX to YY lines since this become an update hell 🙃

egekorkan (approved):

egekorkan:
> @egekorkan @FadySalama please review.
> 
> 
> 
> I tried to simplify the README by removing all XX to YY lines since this become an update hell 🙃

Looks good to me, also ok to remove the line numbers. One thing that I am not sure if it is needed anymore: Do we need to add ajv as a dependency and isn't it the case that the inputs are automatically validated?

danielpeintner:
> One thing that I am not sure if it is needed anymore: Do we need to add ajv as a dependency and isn't it the case that the inputs are automatically validated?

Good point.  You are right, there is no need for ajv anymore. Having said that, ajv is still in the template code provided by @FadySalama. We can do this change in a follow-up PR where the actual code is updated also. Does this make sense?

danielpeintner:
I created https://github.com/eclipse/thingweb.node-wot/issues/754 as a follow-up issue

PR #756, opened 2022-05-13
refactor: make schema in ContentSerdes optional

danielpeintner:
Note: TD might not contain schema information

fixes https://github.com/eclipse/thingweb.node-wot/issues/753

### DIFF
--- a/packages/core/src/codecs/octetstream-codec.ts
+++ b/packages/core/src/codecs/octetstream-codec.ts
@@ -52,7 +52,7 @@ export default class OctetstreamCodec implements ContentCodec {
 
     bytesToValue(
         bytes: Buffer,
-        schema: DataSchema,
+        schema?: DataSchema,
         parameters: { [key: string]: string | undefined } = {}
     ): DataSchemaValue {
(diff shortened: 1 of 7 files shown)


relu91:
I would put on hold and wait for #758

relu91:
Fixed merge conflicts. I think we can go ahead 👍🏻 .

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -66,7 +66,7 @@ export default class OctetstreamCodec implements ContentCodec {
         }
 
         let dataLength = bytes.length;
-        let dataType: string = schema.type;
+        let dataType: string = schema ? schema.type : "undefined";
```
JKRhb, line 69: I think assignments like these could be replaced like so:

```suggestion
        let dataType: string = schema ?? "undefined";
```

danielpeintner (commented):
I am fine with and without the proposed change by @JKRhb

JKRhb:
> @relu91 @JKRhb do you have strong preferences? I am inclined to just move on with this PR...

Sure, feel free to merge from my part :)

Issue #758, opened 2022-05-16
`tsconfig.json` too benevolent


relu91:
This comment got me suspicious https://github.com/eclipse/thingweb.node-wot/pull/748#discussion_r871093794. `tsc` should be able to identify such trivial cases pretty well, so I further investigated. 

What I found is bad... we allowed the whole codebase to be not **strictly** checked. In other words, the root `tsconfig.json` contains `strict: false`. I think this got undercover when we unified all the tsconfigs in one using the composite build. 

Packages to be fixed:
- [x] binding-coap
- [x] binding-modbus
- [x] browser-bundle
- [x] binding-file
- [x] binding-mqtt
- [x] cli
- [x] binding-netconf
- [x] core
- [x] binding-opcua
- [x] examples
- [x] binding-websockets
- [x] td-tools
- [x] binding-http
- [x] binding-mbus
- ~~binding-fujitsu~~
- ~~binding-firestore~~
- ~~binding-firestore-browser-bundle~~

danielpeintner:
Mhh, this is really bad. I quickly checked and there are many issues / warning 😌.
However, I suggest to clean-up the entire repo and all packages.. maybe tackle it package by package.. step by step

relu91:
> I suggest to clean-up the entire repo and all packages.. maybe tackle it package by package.. step by step

yeah, I agree. I'll add some checkboxes to keep track of the progress

relu91:
@JKRhb and @egekorkan since we are having a good sprint here, do you have time to tackle `binding-coap` and `binding-mqtt` ?

relu91:
Binding netconf solved with #1071

relu91:
We the final merge of #1059 we should now have the global strict set to true.

Issue #738, opened 2022-04-19
*breaking* changes - next version 0.8.1 or jump to v0.9.0


danielpeintner:
Note: since we are introducing some *breaking* changes in https://github.com/eclipse/thingweb.node-wot/pull/731 @relu91 is suggesting to jump to v0.9.0
see https://github.com/eclipse/thingweb.node-wot/pull/731#discussion_r844984781

EDIT: One downside is a bit that we loose the connection between  "wot-typescript-definitions" versioning

_Originally posted by @danielpeintner in https://github.com/eclipse/thingweb.node-wot/issues/727#issuecomment-1092634350_

danielpeintner:
My take is that since we only remove deprecated methods I personally would go for version 0.8.1.

What do others think?

egekorkan:
From my point of view, this is rather a small update (just removing the consumed thing methods right?) thus more inclined for 0.8.1 but this would, of course, contradict with semver. There have been previous cases of semver getting broken, like removing writeproperty from exposed thing but then bringing it back (couldn't find the issue). One thing that would actually bother me would be that outsiders getting confused since the perceived change from 7->8 and 8->9 is the same when seen from the version number but the implementation change is not, especially now since 8 is just published. For the migration guides, we would then need to directly say 0.9 and also deprecate 0.8 from npm. Also, all the marketing posts we have done would look outdated now :/

To summarize:
- I am sort of ok to go for 0.9 if we deprecate 0.8 from npm
- I am ok to go for 0.8.1 while putting a note that removing the inherited functions was a breaking change but we did it quickly after 0.8.0 and would affect only a small number of people

danielpeintner:
closing since
* [v0.8.1](https://github.com/eclipse/thingweb.node-wot/releases/tag/v0.8.1) has been released
* see also https://github.com/eclipse/thingweb.node-wot/issues/770

PR #759, opened 2022-05-17
chore: regenerate package-lock.json for next release

danielpeintner:
I regenerated package-lock to get ready for next release `0.8.1`.

Does this sound reasonable for everyone? We have many minor fixes/improvements and firestore binding is back also.

This makes me think we should publish an update...
@relu91 @sebastiankb @egekorkan

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5734,18 +5734,6 @@
             "integrity": "sha1-l4hXRCxEdJ5CBmE+N5RiBYJqvYA=",
             "dev": true
         },
-        "node_modules/detect-libc": {
-            "version": "1.0.3",
-            "resolved": "https://registry.npmjs.org/detect-libc/-/detect-libc-1.0.3.tgz",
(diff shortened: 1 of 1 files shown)


relu91 (approved):
Ok for releasing 0.8.1 but please check https://github.com/eclipse/thingweb.node-wot/issues/738

relu91:
On a second thought should we wait for #758?

danielpeintner:
> Ok for releasing 0.8.1 but please check #738

I talked with @egekorkan and @sebastiankb  and we were thinking that sticking to 0.8.1 is better for now.

> On a second thought should we wait for #758?

I don't think this is a blocker and can follow also in 0.8.2.
Moreover, I don't think this will happen within some days. 

Hence , I suggest to publish NOW and publish again if we have some updates.

relu91:
>I talked with @egekorkan and @sebastiankb and we were thinking that sticking to 0.8.1 is better for now.

yes please check my latest comment. let's merge this and release

PR #762, opened 2022-05-17
Proposal for a new logging API

JKRhb:
As I mentioned in https://github.com/eclipse/thingweb.node-wot/issues/17#issuecomment-927098851, I see some potential for improvement of the current logging solution as a lot of debug messages are added to the console by default. There is of course the possibility to redefine the functions on the `console` object, but I don't think that this is the most elegant solution and also removes the capability of printing anything to the console at all.

In this PR, I therefore wanted to propose a different solution which uses a `LogLevel` enum and a `log` function which only prints the logging statement if the `LogLevel` is high enough (otherwise it gets ignored). The log level can be set globally using a `setLogLevel` function. I also introduced individual logging functions for each log level so that the actual logging provider could be replaced (for example, with functions from the `debug` package). For now, I only refactored the `ConsumedThing` class to demonstrate how the new API could look in practice.

I am looking forward to your feedback :)

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -14,6 +14,7 @@
  ********************************************************************************/
 
 import { ConsumedThing as IConsumedThing, InteractionInput, Subscription } from "wot-typescript-definitions";
+import { log, LogLevel } from "./logger";
 
 import * as TD from "@node-wot/td-tools";
 
@@ -28,6 +29,8 @@ import UriTemplate = require("uritemplate");
 import { InteractionOutput } from "./interaction-output";
 import { FormElementEvent, FormElementProperty } from "wot-thing-description-types";
 import { ThingInteraction } from "@node-wot/td-tools";
+
+const debugPrefix = "core/consumed-thing";
 
 enum Affordance {
     PropertyAffordance,
@@ -142,7 +145,7 @@ class InternalPropertySubscription extends InternalSubscription {
         if (!form) {
             throw new Error(`ConsumedThing '${this.thing.title}' did not get suitable form`);
         }
-        console.debug("[core/consumed-thing]", `ConsumedThing '${this.thing.title}' unobserving to ${form.href}`);
+        log(LogLevel.Debug, debugPrefix, `ConsumedThing '${this.thing.title}' unobserving to ${form.href}`);
         await client.unlinkResource(form);
         this.active = false;
     }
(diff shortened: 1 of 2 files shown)


JKRhb:
Made some adjustments which should simplify the API a bit more.

danielpeintner:
I need to play with it a bit more but it looks promising.
So the idea is to to add logger.ts _once_ in the base package (which would be `td-tools`) and use it in all other packages.

I am wondering, shall it be possible to set different log levels based on the package?
This could be achieved by specifying the log level based on prefix ..

JKRhb:
> I am wondering, shall it be possible to set different log levels based on the package?
> This could be achieved by specifying the log level based on prefix ..

Oh, that is a very good idea! I will add that as an optional argument to the PR.

JKRhb:
> > I am wondering, shall it be possible to set different log levels based on the package?
> > This could be achieved by specifying the log level based on prefix ..
> 
> Oh, that is a very good idea! I will add that as an optional argument to the PR.

I incorporated the suggestion and also added some documentation :)

JKRhb:
Simplified some parts of the code and added the possibility to define your own custom logging and formatting functions, making the API a bit more versatile.

relu91:
Thanks, @JKRhb for your contribution! as always they are very welcomed! Sadly, even though I agree that the current logging system is far from perfect, I wouldn't re-invent the wheel by creating yet another logging utility class. In my opinion, the best option to improve our logging system is to use [debug](https://www.npmjs.com/package/debug). I already used it in other projects and I found it very convenient. Other pros:
- Well-known 43K of users
- Well supported (last published 2 months ago)
- Support for browser
- Support for ENV configuration
- Cool formatters
- Colors
- Namespaces

JKRhb:
Thank you for your feedback, @relu91! I agree, using debug is probably the best solution :) I just wanted to give some input for the discussion as I saw a lot of potential for improvement and got the impression that the discussion was leaning towards not using an external library. However, using the debug library seems like a good practice to me.

Do you want to make a proposal using the debug library instead? Or should I update the PR? In the latter case, I unfortunately might not find that much time to do so in the upcoming weeks.

relu91:
> Do you want to make a proposal using debug instead? Or should I update the PR? In the latter case, I unfortunately might not find that much time to do so in the upcoming weeks.

If @danielpeintner is ok I would close this PR and create a new one with the proposal 👍🏻 maybe just focused on core and cli ... but it depends on how much time do you have!

danielpeintner:
Fine my be 👍

relu91:
as discussed I'm closing the PR

PR #767, opened 2022-05-20
First steps in fixing #758

relu91:
I've started the long journey in the pursuit of a type-safe code base. This PR is still a work in progress since I have still to solve the issues caused by my changes in the other packages. My plan in this PR is to have td-tools and core strictly checked. The good news is that they already are, BUT on the way, I had to make some choices that influenced the other packages. To fully close this PR I will fix the errors in the other packages to have the CI/CD work again. Meanwhile, I'd ask you to review core and td-tools because I'll try to not touch those anymore. 

Let me know if you see something off! 👍🏻


packages/core/test/ClientTest.ts
```diff
@@ -189,9 +189,13 @@ class TDClientFactory implements ProtocolClientFactory {
...
 class TrapClient implements ProtocolClient {
-    private trap: (...args: unknown[]) => Content | Promise<Content>;
-
-    public setTrap(callback: (...args: unknown[]) => Content | Promise<Content>) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private trap: (...args: any[]) => Content | Promise<Content> = () => ({
+        type: "application/json",
+        body: Readable.from(Buffer.from("")),
+    });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
JKRhb, line 197: :)

```suggestion

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```

packages/core/src/consumed-thing.ts
```diff
@@ -452,7 +474,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
             throw new Error(`ConsumedThing '${this.title}' has no links for this interaction`);
         }
 
-        let form: TD.Form;
+        let form;
```
JKRhb, line 477: Can the type be implied here? Or is it set to an implicit `any`?

packages/core/src/consumed-thing.ts
```diff
@@ -491,7 +513,8 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
                     "[core/consumed-thing]",
                     `ConsumedThing '${this.title}' chose cached client for '${schemes[cacheIdx]}'`
                 );
-                client = this.getClients().get(schemes[cacheIdx]);
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- if cacheIdx is valid, then clients *contains* schemes[cacheIdx]
+                client = this.getClients().get(schemes[cacheIdx])!;
```
JKRhb, line 517: Would it make sense to throw an error here instead?

packages/core/src/consumed-thing.ts
```diff
@@ -659,7 +683,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
 
         const content = await client.invokeResource(form, input);
         // infer media type from form if not in response metadata
-        if (!content.type) content.type = form.contentType;
+        if (!content.type) content.type = form.contentType ?? "application/json";
```
JKRhb, line 686:
```suggestion
        if (content.type == null) {
            content.type = form.contentType ?? "application/json";
        }
```

packages/core/src/consumed-thing.ts
```diff
@@ -765,7 +790,8 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         await client.subscribeResource(
             form,
             (content) => {
-                if (!content.type) content.type = form.contentType;
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- tsc get confused when nullables are to listeners lambdas
+                if (!content.type) content.type = form!.contentType ?? "application/json";
```
JKRhb, line 794:
```suggestion
                if (content.type == null) {
                    content.type = form!.contentType ?? "application/json";
                }
```

packages/core/src/content-serdes.ts
```diff
@@ -162,16 +171,16 @@ export class ContentSerdes {
...
             console.debug("[core/content-serdes]", `ContentSerdes serializing to ${contentType}`);
-            const codec = this.codecs.get(mt);
             bytes = codec.valueToBytes(value, schema, par);
         } else {
             console.warn(
                 "[core/content-serdes]",
                 `ContentSerdes passthrough due to unsupported serialization format '${contentType}'`
             );
-            bytes = Buffer.from(value.toString());
+            bytes = Buffer.from(value === null ? "" : value.toString());
```
JKRhb, line 183: Is it intended to only check for `null` and not also for `undefined` (using `value == null`) here?

packages/core/src/exposed-thing.ts
```diff
@@ -95,7 +95,15 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
                 return this.subjectTDChange;
             };
         })().getSubject;
-
+        // The init object might still have undefined values, so initialize them here.
+        // TODO: who checks that those are valid?
+        this.id = thingModel.id || "";
+        this.title = thingModel.title || "";
```
JKRhb, line 101:
```suggestion
        this.id = thingModel.id ?? "";
        this.title = thingModel.title ?? "";
```

packages/core/src/protocol-helpers.ts
```diff
@@ -57,14 +63,10 @@ export default class ProtocolHelpers {
...
     ): void {
-        if (
-            form &&
-            tdTemplate &&
-            tdTemplate.properties &&
-            tdTemplate.properties[propertyName] &&
-            tdTemplate.properties[propertyName].forms
-        ) {
-            for (const formTemplate of tdTemplate.properties[propertyName].forms) {
+        if (form && tdTemplate.properties?.[propertyName]) {
```
JKRhb, line 66: Do we actually need to check for `form` here?

packages/core/src/protocol-helpers.ts
```diff
@@ -57,14 +63,10 @@ export default class ProtocolHelpers {
...
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we know that tdTemplate.properties[propertyName] is not null
+            const property = tdTemplate.properties[propertyName]!;
```
JKRhb, line 68: Hmm, should the property itself be passed as a parameter instead?

packages/core/src/protocol-helpers.ts
```diff
@@ -57,14 +63,10 @@ export default class ProtocolHelpers {
...
+            for (const formTemplate of property.forms ?? []) {
```
JKRhb, line 69: I think we could also get rid of the eslint disable comment above by using something like this:

```suggestion
            for (const formTemplate of property?.forms ?? []) {
```

packages/core/src/protocol-helpers.ts
```diff
@@ -84,14 +86,10 @@ export default class ProtocolHelpers {
...
-            tdTemplate &&
-            tdTemplate.actions &&
-            tdTemplate.actions[actionName] &&
-            tdTemplate.actions[actionName].forms
-        ) {
-            for (const formTemplate of tdTemplate.actions[actionName].forms) {
+        if (form && tdTemplate.actions?.[actionName]) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we know that tdTemplate.actions[actionName] is not null
+            const action = tdTemplate.actions[actionName]!;
+            for (const formTemplate of action.forms ?? []) {
```
JKRhb, line 92: Same here.

```suggestion
            const action = tdTemplate.actions[actionName];
            for (const formTemplate of action?.forms ?? []) {
```

packages/core/src/protocol-helpers.ts
```diff
@@ -111,14 +109,10 @@ export default class ProtocolHelpers {
...
-            tdTemplate &&
-            tdTemplate.events &&
-            tdTemplate.events[eventName] &&
-            tdTemplate.events[eventName].forms
-        ) {
-            for (const formTemplate of tdTemplate.events[eventName].forms) {
+        if (form && tdTemplate.events?.[eventName]) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we know that tdTemplate.events[eventName] is not null
+            const event = tdTemplate.events[eventName]!;
+            for (const formTemplate of event.forms ?? []) {
```
JKRhb, line 115:
```suggestion
            const event = tdTemplate.events[eventName];
            for (const formTemplate of event?.forms ?? []) {
```

packages/core/src/protocol-helpers.ts
```diff
@@ -360,14 +386,17 @@ export default class ProtocolHelpers {
         if (interaction.forms !== undefined && finalFormIndex === -1) {
             if (operationName !== undefined) {
```
JKRhb, line 387: Maybe this allows for getting rid of the comments below?

```suggestion
        if (interaction.forms != null && finalFormIndex === -1) {
            if (operationName != null) {
```

packages/core/src/servient.ts
```diff
@@ -48,12 +48,18 @@ export default class Servient {
         // initializing forms fields
         thing.forms = [];
         for (const name in thing.properties) {
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
JKRhb, line 52: Is this intentional? If I understood it correctly, this would lead to the line not being transpiled at all.

packages/core/src/servient.ts
```diff
@@ -70,7 +76,7 @@ export default class Servient {
     }
 
     public addThing(thing: ExposedThing): boolean {
-        if (thing.id === undefined) {
+        if (thing.id === undefined || thing.id === null || thing.id.length === 0) {
```
JKRhb, line 79: As mentioned above, `== null` checks for both `undefined` and `null` at the same time.

```suggestion
        if (thing.id == null || thing.id.length === 0) {
```

packages/core/src/wot-impl.ts
```diff
@@ -73,7 +90,7 @@ export default class WoTImpl {
             );
             return newThing;
         } catch (err) {
-            throw new Error("Cannot consume TD because " + err.message);
+            throw new Error("Cannot consume TD because " + (err instanceof Error ? err.message : err));
```
JKRhb, line 93:
```suggestion
            throw new Error(`Cannot consume TD because ${err instanceof Error ? err.message : err}`);
```

packages/core/src/wot-impl.ts
```diff
@@ -101,23 +118,12 @@ export default class WoTImpl {
                     throw new Error("Thing already exists: " + newThing.title);
                 }
             } catch (err) {
-                reject(new Error("Cannot produce ExposedThing because " + err.message));
+                reject(new Error("Cannot produce ExposedThing because " + (err instanceof Error ? err.message : err)));
```
JKRhb, line 121:
```suggestion
                reject(new Error(`Cannot produce ExposedThing because ${err instanceof Error ? err.message : err}`));
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -239,7 +239,7 @@ export class ThingModelHelpers {
                                 console.debug("[td-tools]", "http fetched:", parsedData);
                                 resolve(parsedData);
                             } catch (e) {
-                                console.error(e.message);
+                                console.error("[td-tools]", (e as Error).message);
```
JKRhb, line 242: Maybe we could also call `e.toString()` here.

JKRhb (commented):
Thank you for your hard work! See a couple of comments/suggestions below :)

packages/core/src/consumed-thing.ts
danielpeintner: I am not sure if the suggested changes is identical. JS `null` and `undefined` lead to the same boolean while the suggestion is purely about `null`.

JKRhb: I also only realized it due to stricter null checks in the linter for node-code, where this check is the only comparison which allows for using `==` instead of `===` :)

JKRhb: (Overall, the handling of nullish values is definitely not the most intuitive language feature of JS...)

packages/core/src/consumed-thing.ts
relu91: The thing is that at the point we are sure that this.getClients().get(schemes[cacheIdx])! will not be undefined. Check line 508 cacheIdx will be not be -1 if this.getClients() contains that scheme.

packages/core/src/consumed-thing.ts
relu91: Generally, I avoid using `==` just because it confuses people. I'm wondering why should we use your proposal?

packages/core/src/content-serdes.ts
relu91: So if we just trust the tsc compiler we know that value is not undefined, therefore it does not make sense. Some people say that in general is not guaranteed that people will use your library from TS and therefore you should take into account corner cases that are covered by the compiler. My problem here is that we don't have a policy. So if we check for undefined here, then we have to do it everywhere basically.... not sure if is something that we want.

packages/core/src/consumed-thing.ts
JKRhb: Oh, sorry, I think I might have internalized [this eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md) (which is currently enforced in node-coap) a bit too much. I thought it was a requirement of a stricter tsconfig, but that is not actually the case.

The rationale for using `foo == null` (which is only supposed to be used for this particular null check, otherwise using `==` is discouraged for good reasons) is that it potentially contributes to more type safety, as it does not become true for the empty string, empty arrays and objects, and 0 (in contrast to `!foo`). However, it does potentially require more accompanying type checks to not let these value falls through (which is actually a potential source of new bugs, as I had to experience).

The change was just a suggestion, though, so we can simply mark the thread as resolved :) However, for the future it might be worth considering to have even stricter linting rules to enforce additional type safety.

packages/core/src/consumed-thing.ts
JKRhb: Same here, could simply be resolved.

packages/core/src/content-serdes.ts
JKRhb: Hmm, there is a check above, though, which logs a warning if the value is `undefined`. To be consistent, that check would need to be removed as well?

My suggestion here would be to simply use the following:

```suggestion
            bytes = Buffer.from(value?.toString() ?? "");
```

This way, the code is a bit more concise and `undefined` is also covered (which would otherwise cause an error, as `toString()` seems to not exist on `undefined`, in contrast to other data types).

relu91:
@JKRhb and @danielpeintner I'm starting working on this again. I'm not sure about what I need to address first. What do you prefer between:
1. Resolve @JKRhb review comments/suggestions
2. Fix other packages

danielpeintner:
I think we should  focus on `core` and `td-tools` for now.
Other packages may also follow in future PRs.

relu91:
so you are ok even if the other packages have building errors? (my personal take is to go with this strategy)

packages/core/src/servient.ts
JKRhb: Oh, sorry, you are right, feel free to mark this as resolved :)

danielpeintner:
> so you are ok even if the other packages have building errors? (my personal take is to go with this strategy)

One the short run either way is fine for me.
Before merging this PR we should have a clean build. I guess, there is no question with that regard.

What I think is a good way forward is to properly fix `core` and `td-tools` with strict rules in this PR.
The other packages can be fixed so that it compiles. In follow-up PRs we can fix any other binding/package with strict rules in place...

I don't know whether that makes sense for everybody?

relu91:
> Before merging this PR we should have a clean build. I guess, there is no question with that regard.

Ok then for a clear build I have to change also non-core packages... let's hope it is not a massive change! But to simplify the review process I would first fix @JKRhb's comments.

packages/core/src/content-serdes.ts
relu91: I might have agreed with @JKRhb's suggestion but then I thought more about the logic. IMHO is wrong... The else branch is activated if no codec is found and it defaults to simple sending value.toString or "". But value.toString() is not always correct. For example for objects it returns: [object Object] . Why should we send this to the wire? 🤔 So the resolve this conversation I would like to open an issue instead of fixing this detail. thumbs up if you agree

danielpeintner (approved):
Good work @relu91 👍

Looks good!

Since I don't have much time before the vacation and I think it would be good to merge it soon I tried to look as good as possible.

Note: The PR introduces some more // @ts-ignore we might want to look later.

BTW, see question below w.r.t. to killme.js

killme.js
```diff
@@ -0,0 +1,28 @@
+const delay = (ms) => {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+};
```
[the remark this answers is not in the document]
relu91: Oops added unintentionally, it was there to test #771. I'll remove it and then merge!

relu91:
File removed, merging.

PR #768, opened 2022-05-23
feat(binding-coap)!: use new binding-template

JKRhb:
This PR implements a first version of the [CoAP Binding-Template](https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/index.html) in the `binding-coap` package. While all current vocabulary terms are included, not all of them are handled yet (especially the qblockwise and hop-limit parameters).

Due to the changed prefix and a new vocabulary term for the methods, this PR should actually constitute a breaking change. Therefore, it might be included in the 0.9.0 release?

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -26,7 +26,7 @@ import { Subscription } from "rxjs/Subscription";
 import * as TD from "@node-wot/td-tools";
 
 import { ProtocolClient, Content, ContentSerdes, createLoggers } from "@node-wot/core";
-import { CoapForm, CoapOption, CoapMethodName, isSupportedCoapMethod, isValidCoapMethod } from "./coap";
+import { BlockSize, blockSizeToOptionValue, CoapForm, CoapMethodName } from "./coap";
 import CoapServer from "./coap-server";
 import { Readable } from "stream";
 import {
@@ -218,51 +218,55 @@ export default class CoapClient implements ProtocolClient {
         return options;
     }
 
-    private determineRequestMethod(formMethod: CoapMethodName, defaultMethod: CoapMethodName): CoapMethodName {
-        if (isSupportedCoapMethod(formMethod)) {
-            return formMethod;
-        } else if (isValidCoapMethod(formMethod)) {
-            // TODO: Merge with condition above when a new version of node-coap is
-            //       released
-            debug(`Method ${formMethod} is not supported yet. Using default method ${defaultMethod} instead.`);
-        } else {
-            debug(`Unknown method ${formMethod} found. Using default method ${defaultMethod} instead.`);
+    private setBlockOption(req: OutgoingMessage, optionName: "Block1" | "Block2", blockSZX?: BlockSize): void {
+        if (blockSZX == null) {
+            return;
         }
 
-        return defaultMethod;
+        try {
+            const block2OptionValue = blockSizeToOptionValue(blockSZX);
+            req.setOption(optionName, block2OptionValue);
+        } catch (e) {
+            warn(e.toString());
+        }
+    }
+
+    private getRequestParamsFromForm(form: CoapForm, defaultMethod: CoapMethodName, observe = false) {
+        // TODO: Add qblock parameters and hop limit, once implemented in node-coap
+
+        const method = form["cov:method"] ?? defaultMethod;
+        debug(`CoapClient got Form "method" ${method}`);
+
+        const contentFormat = form["cov:contentFormat"] ?? form.contentType ?? "application/json";
+        debug(`"CoapClient got Form 'contentType' ${contentFormat} `);
+
+        const accept = form["cov:accept"];
+        if (accept != null) {
+            debug(`"CoapClient determined Form 'accept' ${accept} `);
+        }
+
+        return {
+            ...this.uriToOptions(form.href),
+            method,
+            observe,
+            contentFormat,
+            accept,
+        };
+    }
+
+    private applyFormDataToRequest(form: CoapForm, req: OutgoingMessage) {
+        const blockwise = form["cov:blockwise"];
+        if (blockwise != null) {
+            this.setBlockOption(req, "Block2", blockwise["cov:block2SZX"]);
+            this.setBlockOption(req, "Block1", blockwise["cov:block1SZX"]);
+        }
     }
 
     private generateRequest(form: CoapForm, defaultMethod: CoapMethodName, observable = false): OutgoingMessage {
-        const options: CoapRequestParams = this.uriToOptions(form.href);
+        const requestParams = this.getRequestParamsFromForm(form, defaultMethod, observable);
 
-        if (form["cov:methodName"] != null) {
-            const formMethodName = form["cov:methodName"];
-            debug(`CoapClient got Form "methodName" ${formMethodName}`);
-            options.method = this.determineRequestMethod(formMethodName, defaultMethod);
-        } else {
-            options.method = defaultMethod;
-        }
-
-        options.observe = observable;
-
-        const req = this.agent.request(options);
-
-        // apply form data
-        if (typeof form.contentType === "string") {
-            debug("CoapClient got Form 'contentType'", form.contentType);
-            req.setOption("Accept", form.contentType);
-        }
-        if (Array.isArray(form["cov:options"])) {
-            debug("CoapClient got Form 'options'", form["cov:options"]);
-            const options = form["cov:options"] as Array<CoapOption>;
-            for (const option of options) {
-                req.setOption(option["cov:optionName"], option["cov:optionValue"]);
-            }
-        } else if (typeof form["cov:options"] === "object") {
-            warn("CoapClient got Form SINGLE-ENTRY 'options'", form["cov:options"]);
-            const option = form["cov:options"] as CoapOption;
-            req.setHeader(option["cov:optionName"], option["cov:optionValue"]);
-        }
+        const req = this.agent.request(requestParams);
+        this.applyFormDataToRequest(form, req);
 
         return req;
     }
--- a/packages/binding-coap/src/coap.ts
+++ b/packages/binding-coap/src/coap.ts
@@ -14,8 +14,6 @@
  ********************************************************************************/
 
 import { Form } from "@node-wot/td-tools";
-import { OptionName } from "coap-packet";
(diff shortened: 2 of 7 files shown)


relu91:
> Due to the changed prefix and a new vocabulary term for the methods, this PR should actually constitute a breaking change. Therefore, it might be included in the 0.9.0 release?

> Having said that, we really need a "better" versioning strategy.. w.r.t. to node-wot core but also w.r.t. to the bindings

My understanding is that we are not following semver right now. So we can keep the current rationale that we "manually" decide to switch from 0.x.0 to 0.y.0 . IMHO the right way of fixing this is to set a plan for 1.0.0 and declare that we are in a "beta" stage were we don't follow semver in the README

packages/binding-coap/test/coap-client-test.ts
```diff
@@ -117,7 +117,7 @@ class CoapClientTest {
         const coapClient = new CoapClient(coapServer);
         const form: CoapForm = {
             href: `coap://127.0.0.1:${port2}`,
-            "cov:methodName": "GET",
+            "coap:method": "GET",
```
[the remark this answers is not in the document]
JKRhb: Hmm, good question. The [HTTP-in-RDF](https://www.w3.org/TR/HTTP-in-RDF10/) document uses `http` as a prefix. However, the TD document itself [uses `htv`](https://w3c.github.io/wot-thing-description/#http-binding-assertions), so I suppose the default prefix could only be changed in TD 2.0?

relu91 (approved):

packages/binding-coap/test/coap-client-test.ts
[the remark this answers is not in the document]
danielpeintner: @egekorkan do you know more?

danielpeintner (commented):
I am not sure if we should approve this PR as is.

@egekorkan mentions in https://github.com/eclipse/thingweb.node-wot/pull/768#discussion_r881801127 he would like to see the `cov` prefix...

egekorkan:
Actually I would propose to wait for the editor's draft to become a working draft before going for a major (0.9.0) release since there might be breaking changes in the binding document

danielpeintner:
The [CoAP Binding-Template](https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/index.html) document seems to use `cov` prefix again.
Anyhow, it is still not _properly_ published ...

// For the time being I converted the PR to draft.

packages/binding-coap/test/coap-client-test.ts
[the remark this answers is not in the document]
JKRhb: Changed the prefix back to `cov:`

JKRhb:
> The [CoAP Binding-Template](https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/index.html) document seems to use `cov` prefix again. Anyhow, it is still not _properly_ published ...
> 
> // For the time being I converted the PR to draft.

👍

In the meantime, I updated the PR to the draft's latest version and rebased it to the master branch, resolving the conflicts.

egekorkan:
FYI: I am planning to get some significant work done on the binding templates and aim for a WD release of the currently well written ones (not hue or xml or json etc.)

packages/binding-coap/src/coap.ts
```diff
@@ -29,16 +27,30 @@ export * from "./coap-client";
...
+
+export type BlockSizeOptionValue = 0 | 1 | 2 | 3 | 4 | 5 | 6;
+
+export interface BlockWiseTransferParameters {
+    "cov:block2SZX"?: BlockSize;
+    "cov:block1SZX"?: BlockSize;
 }
 
-export type CoapMethodName = "GET" | "POST" | "PUT" | "DELETE" | "FETCH" | "PATCH" | "iPATCH";
+export interface QBlockWiseTransferParameters {
```
ektrah, line 41: `QBlockWiseTransferParameters` and `BlockWiseTransferParameters` have been merged in the latest draft, so there's just `BlockWiseTransferParameters` for both now.

packages/binding-coap/src/coap.ts
```diff
@@ -64,3 +76,22 @@ export function isValidCoapMethod(methodName: CoapMethodName): methodName is Coa
...
+ *
+ * Maps the values 16, 32, 64, 128, 256, 512, and 1024 to 0, 1, 2, 3, 4, 5, and 6.
+ * Other values will be rounded to the nearest value accepted by CoAP.
+ *
+ * @param blockSize The original block size value.
+ * @returns A mapped block size value usable by CoAP.
+ */
+export function blockSizeToOptionValue(blockSize: BlockSize): BlockSizeOptionValue {
+    // TODO: Check if there is a more elegant solution.
+    const blockSizeOptionValue = Math.floor(Math.sqrt(blockSize) - 4);
```
ektrah, line 91: [Math.log2](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/log2)?

ektrah (commented):
Looks quite good overall; two small comments below.

I think the `form.contentType` to `Content-Format` conversion needs quite a bit more thinking in the spec. Apart from that, I don't expect any major changes at this point.

packages/binding-coap/src/coap.ts
JKRhb: Thanks, adjusted in 501929a0bd76c10d66f3609b35ea70c8689f04d2 :)

packages/binding-coap/src/coap.ts
```diff
@@ -29,16 +27,25 @@ export * from "./coap-client";
...
+
+export interface BlockWiseTransferParameters {
+    "cov:block2SZX"?: BlockSize;
+    "cov:block1SZX"?: BlockSize;
+}
+
 export class CoapForm extends Form {
-    public "cov:methodName"?: CoapMethodName;
-    public "cov:options"?: Array<CoapOption> | CoapOption;
+    public "cov:method"?: CoapMethodName;
```
JKRhb, line 42: Regarding `cov:method`: As the HTTP binding is using `htv:methodName` (see [here](https://w3c.github.io/wot-binding-templates/bindings/protocols/http/#http-vocabulary-terms)), I was wondering if we should also keep `cov:methodName` for the sake of consistency.

danielpeintner: Consistency is important!
I hope the binding documents do align there as well... which does not seem to be the case.
* https://w3c.github.io/wot-binding-templates/bindings/protocols/http/index.html#http-vocabulary-terms
* https://w3c.github.io/wot-binding-templates/bindings/protocols/coap/index.html#requests-responses

@egekorkan

ektrah: The naming of `cov:method` is consistent with [`modbus:function`](https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/#form-terms) and [`mqv:controlPacket`](https://w3c.github.io/wot-binding-templates/bindings/protocols/mqtt/#controlpacket), so I don't see any need for change here.

packages/binding-coap/src/coap.ts
JKRhb: You're right, that would actually be unintuitive. I converted the function's implementation in 26f53d026af6ed72e3e6dd60a2ecccbf4fe7fad6 to a switch statement, capturing only the allowed values. Hope that it hasn't become a bit too verbose this way.

packages/binding-coap/src/coap-client.ts
```diff
@@ -231,11 +238,47 @@ export default class CoapClient implements ProtocolClient {
...
+    }
+
+    private applyFormData(form: CoapForm, req: OutgoingMessage) {
+        // TODO: Add qblock parameters and hop limit, once implemented in node-coap
+        // TODO: Better differentiate between content type and accept
+
+        const contentFormat = form.contentType;
+        if (contentFormat != null) {
+            debug(`"CoapClient got Form 'contentType' ${contentFormat} `);
+            req.setOption("Content-Format", contentFormat);
```
ektrah, line 261: Let's say there is a property that looks like this:

```json
{
    "op": "readproperty",
    "cov:method": "GET",
}
```

Wouldn't this code set the `Content-Format` option in the GET request to `50`, implying erroneously that the GET request has a payload?

ektrah: Also, since the `contentType` member of a form is mandatory with default value "application/json", I guess the code should look like this:

```diff
-const contentFormat = form.contentType;
-if (contentFormat != null) {
+const contentFormat = form.contentType || "application/json";
 req.setOption("Content-Format", contentFormat);
```

This would mean that *every* request (regardless of method) would have to have a payload, no?

JKRhb: Thank you for the feedback :) I added a couple of updates to the PR. Currently, however, there are a few limitations in node-coap which prevent me from using the numeric values from `cov:contentFormat` and `cov:accept`. I'll try to add these features to node-coap in the upcoming days, then hopefully this PR will be able to be finished :)

JKRhb: I updated node-coap in the meantime, so this thread can be considered resolved :)

JKRhb:
> @JKRhb I assume this is now ready for review, right?

Exactly :)

relu91 (approved):
Thank you @JKRhb! From a quick review, the changes look good. Just a minor comment below and one question:
Regarding [this line](https://github.com/eclipse/thingweb.node-wot/pull/768/files#diff-3f331255e93fb9e3e714530c8787b51ab7615c6cb3ad33a16d1e753d5e5833a7R240) does this means that the coap agent is responsible to translate `contentType` to its CoAP equivalent?

JKRhb:
> Thank you @JKRhb! From a quick review, the changes look good. Just a minor comment below and one question: Regarding [this line](https://github.com/eclipse/thingweb.node-wot/pull/768/files#diff-3f331255e93fb9e3e714530c8787b51ab7615c6cb3ad33a16d1e753d5e5833a7R240) does this means that the coap agent is responsible to translate `contentType` to its CoAP equivalent?

Thank you for the feedback :) The current procedure is as follows: If `form["cov:contentFormat"]` is defined, then the client will simply use the numeric value. Otherwise, it will use the string representation in the `contentType` field or `application/json` as a fallback and try to map it to a Content-Format, based on the (currently somewhat flawed) logic implemented in node-coap.

relu91:
Ok! thanks again :) let's merge this.

PR #773, opened 2022-05-26
WIP: Add proposal for new discovery API

JKRhb:
This PR contains a potential proposal for a new discovery. It is still a work in progress and will be updated in accordance with the discussion in https://github.com/w3c/wot-scripting-api/issues/364.


relu91:
Thank you, Jan! I know that this is still a WIP but I think I saw two issues already:
1. *node-fetch* as a dependency of the core module: I don't want this, I like that the core module is protocol-agnostic as much as possible. Staying protocol agnostic with Discovery is a little bit tricky but my idea was to extend the core-protocol interface to allow also calls for discovery. WoTHelpers fetch method does something in that direction, but I know is far to be complete. We also experiment with the "old" discovery interface internally but sadly I can't show you the code 😢 
2. Examples in core: I think we discussed already that examples should stay on the example project. Did we change this policy?

JKRhb:
Thank you for the feedback, @relu91 :) I am aware of these two issues and the commit that contains them is supposed to be dropped later, node-fetch and the example are only added for illustration purposes. In fact, I already have an idea for a protocol agnostic implementation which I will add later :)

Sorry, I probably should have made these points clearer in the description of the PR :/ However, I really appreciate the feedback :)

JKRhb:
> 1. Staying protocol agnostic with Discovery is a little bit tricky but my idea was to extend the core-protocol interface to allow also calls for discovery.

This is now also the direction I took in my latest commits :) I am not sure about the semantics yet and the design of the API (not to mention that the implementation itself is also still WIP), but the proof of concept for HTTP already works, at least :) Looking forward to having more discussion on this topic.

JKRhb:
I updated the PR with a new approach I found experimenting with Discovery in `dart_wot` and added basic support for CoAP. The new API now delegates the processing of the TD to the `ThingDiscoveryImpl` while the actual binding implementations now simply return `Content`. There are still a couple of quirks and issues (e.g., the ContentSerdes cannot handle `application/td+json` at the moment), but in general, I think this points in the right direction.

One thing I implemented in `dart_wot` I couldn't get to work in Typescript/Javascript yet is the usage of AsyncIterators/Streams at the binding level, which are passed up to the WoT runtime. This is especially useful for discovery from CoRE Resource Directories and/or from multicast. I will do some more research in this regard.

JKRhb:
I think this PR can be considered superseded by #864 in the meantime :) I will therefore close this one in order to clean up the PR tab a bit.

PR #774, opened 2022-05-26
docs: change/align flag names

danielpeintner:
clientonly to client-only
configfile to config-file

fixes https://github.com/eclipse/thingweb.node-wot/issues/772


relu91 (approved):
LGTM

PR #775, opened 2022-05-26
fix: observable property

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/764

### DIFF
--- a/examples/scripts/smart-coffee-machine-client.js
+++ b/examples/scripts/smart-coffee-machine-client.js
@@ -68,6 +68,8 @@ WoTHelpers.fetch("http://127.0.0.1:8080/smart-coffee-machine").then(async (td) =
             // But, of course, could have a much more sophisticated handler
             log("outOfResource event:", await data.value());
         });
+        // fire property change for maintenanceNeeded
+        await thing.writeProperty("servedCounter", 1001);
     } catch (err) {
         console.error("Script error:", err);
     }
--- a/examples/scripts/smart-coffee-machine.js
+++ b/examples/scripts/smart-coffee-machine.js
@@ -254,7 +254,7 @@ Assumes one medium americano if not specified, but time and mode are mandatory f
                 // (the notify function here simply logs a message to the console)
                 notify(
                     "admin@coffeeMachine.com",
-                    `maintenanceNeeded property has changed, new value is: ${servedCounter}`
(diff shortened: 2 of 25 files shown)


relu91 (approved):

danielpeintner:
I updated the TS definitions and added code for `emitPropertyChange()`;

Having said that, it does still not work as excepted. To show the issue I updated the smart-coffee-machine example to trigger `emitPropertyChange()` which fails now with 

```
 Server error: Internal Server Error
    at HttpClient.checkFetchResponse (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-http\dist\http-client.js:290:19)
    at LongPollingSubscription.<anonymous> (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-http\dist\subscription-protocols.js:42:54)
    at fulfilled (C:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-http\dist\subscription-protocols.js:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

@relu91 maybe you have a better hint/idea...

relu91 (commented):
Found issues (both in the scripting API spec and node-wot). Here is a summary:
- Node-wot is failing because is expecting "something to send" over the network, but actually, we are not returning anything (null). 
- Checking the spec we are actually prescribing that we should send the new value of the Property over the network -> node-wot needs to be updated
- Reading the [spec algorithm](https://www.w3.org/TR/wot-scripting-api/#the-emitpropertychange-method) we are mentioning to use also the observeHandler to get the new value of the property and using the readHandler only as a fallback. However, this is a little bit counterintuitive if compared to what we are doing for Events. I belive that a regular user would expect that the observeHanlder is used when someone "observes" a property not when we want to send the new value. We might need some renaming/refactoring
- in node-wot emitProperty change in sync but in the spec is actually async (and it needs to be async because we must call readHandler)

relu91:
I would try to solve the issue above in the next week so that we can merge it when @danielpeintner is back 😃 .

relu91:
Issue #775 should now be solved, however, we need to keep track of the issue in the scripting API because the current implementation uses some assumptions that might be different from the consensus. I'll open an issue after merging this.

PR #779, opened 2022-05-28
Fix oAuth 2.0 in http binding

### LINKED ISSUE #769: OAuth example does not work as intended
The OAuth example in [examples/security/oauth](https://github.com/eclipse/thingweb.node-wot/tree/master/examples/security/oauth) is not working as intended, and I'm wondering how this example could ever have worked.

First of all, I think you messed up the terms `grants`, `grant_types`, and `scope`/`scopes`. The [`memory-model.js`](https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/memory-model.js) file on [line 29](https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/memory-model.js#L29) specifies the `grants` field as:
https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/memory-model.js#L29
The docs for the used [oauth2-server](https://github.com/oauthjs/node-oauth2-server) plugin state that `client.grants` should return `grant_types` such as `client_credentials` or `authorization_code`. The corresponding docs are located at the _Return value:_ section of the `getClient(clientId, clientSecret, [callback])` method [here](https://oauth2-server.readthedocs.io/en/latest/model/spec.html#getclient-clientid-clientsecret-callback). Maybe [this line](https://github.com/oauthjs/node-oauth2-server/blob/6d4c98794bc024a8cf93cf9e79f92f93766da9f4/lib/handlers/token-handler.js#L215) of the oauth2-server package helps to understand the issue.

Additionally, the statement in the [README](https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/README.md) to change the `grants` field by removing the `limited` scope is not working for the same reason:
https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/README.md?plain=1#L39-L40

In short, this means that you need to restructure your memory-model like this:
```
...
this.clients = [
  {
    clientId: "node-wot",
    clientSecret: "isgreat!",
    redirectUris: [""],
    grants: ["client_credentials"],
    scopes: ["observe", "control", "admin"],
  },
];
...
```
...and your [oauth-server](https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/server.js) like this:

```
...
app.use("/introspect", (req, res) => {
  const token = res.locals.oauth.token;
  console.log("Token was", token ? "Ok" : "not Ok");
  res.json({
    active: !!token,
    scope: token.client.scopes.join(" "),
    client_id: token.client.clientId,
  }).end();
});
...
```

... and your [exposer](https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/examples/security/oauth/exposer.js#L19-L27) like this:

```
...
securityDefinitions: {
  oauth2_sc: {
    scheme: "oauth2",
    flow: "client",
    authorization: "https://example.com/authorization",
    token: "https://localhost:3000/token",
    scopes: ["observe", "control", "admin"],
  },
},
...
```

...to make the example work.

I suggest using other scopes than `client` to prevent confusion with the `client_credentials` grant. Indeed, it is anyways confusing because the OAuth standard is using `scope` to indicate the token scope/scopes and the libraries or the WoT standard use the keyword `scopes`.

Maybe I have some time on the weekend to create an Eclipse account for contributing.

PS: I know you marked the `affordance-level` scopes (_scope-values associated with a form_) in the [specification](https://www.w3.org/TR/wot-thing-description11/#form) as a `TODO`:
https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/packages/binding-http/src/http-server.ts#L428-L432
However, the current implementation of the scope validation here:
https://github.com/eclipse/thingweb.node-wot/blob/98a271052a2dd84afa71d8e73079d85a2449ac73/packages/binding-http/src/oauth-token-validation.ts#L135-L142
does not check the scope field. The corresponding _Introspection_ standard in [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2) states that the `scope` field is optional. Moreover, the returned scopes from different resource servers may differ. If the exposer requires a specific scope value and one of the resource servers omits the scope field, device access is unintentionally granted because the validation simply skips scope comparison if the `scope` field is missing.

relu91:
Fixes #769. @estallio can I ask you to review it?

### DIFF
--- a/examples/security/oauth/README.md
+++ b/examples/security/oauth/README.md
@@ -37,7 +37,7 @@ oAuth token was Ok!
 ```
 
 This confirm that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
-to `./memory-model.js` and try to remove the string `"limited"` from the grants. Run again the example and you will see that the action is not executed and an error is returned by the client.
+to `./exposer.js` and try to remove the string `"user"` from the grants. Run again the example and you will see that the action is not executed and an error is returned by the client.
 
 ## Where is a TS version?
 
--- a/examples/security/oauth/README.md
+++ b/examples/security/oauth/README.md
@@ -36,7 +36,7 @@ you should see the following line at the end of consumer log:
 oAuth token was Ok!
 ```
 
-This confirm that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
+This confirms that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
 to `./exposer.js` and try to remove the string `"user"` from the grants. Run again the example and you will see that the action is not executed and an error is returned by the client.
 
 ## Where is a TS version?
--- a/packages/binding-http/src/oauth-token-validation.ts
+++ b/packages/binding-http/src/oauth-token-validation.ts
@@ -130,6 +130,16 @@ export class EndpointValidator extends Validator {
         if (!validationResult.active) {
             return false;
         }
+        if (scopes.length === 0) {
+            // We don't require any particular scope; accept the token
+            return true;
+        }
+
+        if (!validationResult.scope) {
+            // If the token doesn't have any scope and we already know that scopes.length > 0,
+            // then the token is not valid
+            return false;
+        }
 
         // Check if the token's scopes are allowed by the Thing Descriptor
         if (validationResult.scope) {
--- a/packages/binding-http/src/oauth-token-validation.ts
+++ b/packages/binding-http/src/oauth-token-validation.ts
@@ -142,14 +142,12 @@ export class EndpointValidator extends Validator {
         }
 
         // Check if the token's scopes are allowed by the Thing Descriptor
-        if (validationResult.scope) {
-            const tokenScopes = validationResult.scope.split(" ");
-            const validScope = tokenScopes.some((tokenScope) => {
-                return scopes.some((thingScope) => tokenScope === thingScope);
-            });
+        const tokenScopes = validationResult.scope.split(" ");
+        const validScope = tokenScopes.some((tokenScope) => {
+            return scopes.some((thingScope) => tokenScope === thingScope);
+        });
 
-            if (!validScope) return false;
-        }
+        if (!validScope) return false;
 
         // Check if the client was allowed in the servient configuration file
         if (validationResult.client_id && !validationResult.client_id.match(clients)) {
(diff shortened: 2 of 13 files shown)


estallio:
The changes look good and the example works as intended now :thumbsup:
I have commented one line that seems unnecessary depending on your code policy.
However, if there is anything else I can do - just let me know :slightly_smiling_face:

examples/security/oauth/README.md
```diff
@@ -37,7 +37,7 @@ oAuth token was Ok!
 ```
 
 This confirm that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
```
danielpeintner, line 39: This confirmS

packages/examples/src/security/oauth/README.md
```diff
@@ -40,7 +40,7 @@ oAuth token was Ok!
 ```
 
 This confirm that the oAuth flow was completed successfully. Now you can have fun revoking the access to the consumer script. Go
```
danielpeintner, line 42: confirmS

danielpeintner (approved):
LGTM.

see 2 minor nits...

danielpeintner:
> I have commented on one line that seems unnecessary, depending on your code policy.

@estallio the comment did not make it through.. somehow..

packages/binding-http/src/oauth-token-validation.ts
```diff
@@ -130,6 +130,16 @@ export class EndpointValidator extends Validator {
...
+        }
+
+        if (!validationResult.scope) {
+            // If the token doesn't have any scope and we already know that scopes.length > 0,
+            // then the token is not valid
+            return false;
+        }
 
         // Check if the token's scopes are allowed by the Thing Descriptor
         if (validationResult.scope) {
```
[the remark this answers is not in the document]
relu91: Yeah you're right I was in an hurry and I forget to delete it 👍🏻

danielpeintner (approved):

PR #780, opened 2022-06-02
fix: return the object as described in the TD

MartinMeyer1:
The json codec returned the value corresponding to the key "value" of the readed property, instead of the full json object as expected and described in the TD.

Signed-off-by: MartinMeyer1 <meyer.mart@outlook.com>

### DIFF
--- a/packages/core/src/codecs/json-codec.ts
+++ b/packages/core/src/codecs/json-codec.ts
@@ -51,7 +51,7 @@ export default class JsonCodec implements ContentCodec {
                 throw err;
             }
         }
-		
+
         return parsed;
     }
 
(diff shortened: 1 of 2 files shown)


MartinMeyer1:
I can't get why this doesn't work on windows...

danielpeintner (approved):
LGTM.

by now all implementations should no longer use the "wrapper"

danielpeintner:
@MartinMeyer1 can you merge in the changes from master...

danielpeintner (changes requested):
I noticed we have the same wrapper in another codec.

https://github.com/eclipse/thingweb.node-wot/blob/8f30e313d100e19bccd31446ee7fd0013c30e9d4/packages/core/src/codecs/netconf-codec.ts#L47-L52

Can you fix this also as part of this PR. Thanks!

danielpeintner (approved):
Looks good.

Thanks @MartinMeyer1

PR #782, opened 2022-06-06
test: use lower port numbers for CoAP server tests

### LINKED ISSUE #781: CoAP CI failures
Not sure if it is just a random failure or something we overlooked:
image

It seems on mac sometimes we are receiving a earlier than expected (or a different error type)

JKRhb:
Hopefully fixes #781.


danielpeintner (approved):
LGTM.

Let's give it a try whether it solves the issue we were seeing...

PR #783, opened 2022-06-06
refactor(core): use debug module for logging

JKRhb:
As discussed in #762, this PR introduces the `debug` module for logging in the core package. This cleans up the test logs a lot and makes node-wot more usable for production, IMHO.


JKRhb:
Hmm. Not sure what causes the error here. I think the failed tests can simply be restarted?

packages/core/src/codecs/netconf-codec.ts
```diff
@@ -16,14 +16,18 @@
...
+
 /** default implementation offering JSON de-/serialisation */
 export default class NetconfCodec implements ContentCodec {
     getMediaType(): string {
         return "application/netconf";
     }
 
     bytesToValue(bytes: Buffer, schema: DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
-        // console.debug(`NetconfCodec parsing '${bytes.toString()}'`);
+        // log(`NetconfCodec parsing '${bytes.toString()}'`);
```
danielpeintner, line 30: General question. Shall we "re-activate" those commented-out log messages

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -123,7 +127,7 @@ export default class OctetstreamCodec implements ContentCodec {
 
                         // warn about numbers being too big to be represented as safe integers
                         if (!Number.isSafeInteger(result)) {
-                            console.warn("[core/octetstream-codec]", "Result is not a safe integer");
+                            log("Result is not a safe integer");
```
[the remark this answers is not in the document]
danielpeintner: I agree, we should make clear whether we need this granularity.

I personally think *some* logging like the example above "Result is not a safe integer" as a warning is important always.
Some more fine-grained logging like the example below.. "NetconfCodec parsing ..." is something I would like to turn on only if I would like to better understand what is going on.

Does that make sense?

packages/core/src/codecs/netconf-codec.ts
JKRhb: Added in f8d1f0767fa2a0e03de18457cc33bb7ab667d825

packages/core/src/codecs/octetstream-codec.ts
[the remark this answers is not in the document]
JKRhb: Hmm, would it make sense to create a `warn` alias for `log` for now to preserve the original differentiation between log levels and decide on the details in a follow-up PR? Or do you have a concrete proposal for how to proceed here?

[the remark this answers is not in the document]
relu91: > Hmm, would it make sense to create a warn alias for log for now to preserve the original differentiation between log levels and decide on the details in a follow-up PR? Or do you have a concrete proposal for how to proceed here?

I ok with this plan

JKRhb:
Thank you, @relu91! I am currently a bit occupied until Friday, but after then I will start to update the PR immediately :)

packages/core/src/codecs/octetstream-codec.ts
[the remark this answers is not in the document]
JKRhb: Sorry for the delay! I now updated the PR accordingly, looking forward to your feedback :)

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -17,6 +17,11 @@ import { ContentCodec } from "../content-serdes";
 import { DataSchema, DataSchemaValue } from "wot-typescript-definitions";
 import { getFloat16, setFloat16 } from "@petamoriken/float16";
 
+import debug from "debug";
+
+const logDebug = debug("core/octetstream-codec");
+const logWarn = logDebug;
+
```
danielpeintner, line 24: Since I am not familiar with the debug library I was wondering how "we" or "users" of node-wot can configure to log only warnings/errors besides *all* debug statements?

Moreover, wouldn't it make sense to add a prefix to make clear which kind of log statement it is? Like the following?
const logDebug = debug("[debug] core/octetstream-codec");
const logWarn = debug("[warn] core/octetstream-codec");

JKRhb: Hmm, maybe we should actually consider using [debug-level](https://www.npmjs.com/package/debug-level) instead then, as it would give us both capabilities and more.

danielpeintner: Makes sense to me.
@relu91?

relu91: Some concerns about `debug-level`:
- It is not that popular
- It seems to have just one maintainer
- It has quite a few dependencies for a logging library
- It depends directly on debug but it implements its own logic (I actually don't understand why the dev choose this path)
- I think we can use namespaces to achieve "almost" the same behaviour example below

```js
const debug = debug("core:octetstream-codec:debug");
const info = debug("core:octetstream-codec:info");
```

Notice that the *well-known* separator is `:` as document [here](https://www.npmjs.com/package/debug#Conventions). Enable or disable level nails down to correctly set up the environment variable:
```bash
DEBUG=core:*:info # any info core message
DEBUG=core:* # any core message
```
We sort of lose di ability to say I want ALL the messages from WARN to INFO but I think is not really so problematic, because you can still do this:
```bash
DEBUG=core:*:info, core:*:warn # any core info and warn messages
```

Therefore I would still go for the `debug` library, what do you think?

danielpeintner: This path is fine by me also. We just need to carefully document this or point to a proper documentation so that people know how to configure debugging properly.

JKRhb: Thank you, @relu91, explanation of the correct usage of the debug module! I now applied the separator/namespace syntax and also did some refactoring, locating the functions for creating the loggers again in a `loggers.ts` file.

The functions defined here allow you to either create a logging function for one specific log level (using, for example, `createDebugLogger` or `createErrorLogger`) or create multiple loggers at once, like so: 

```ts
const { debug, warn } = createLoggers("core", "servient");
```

When creating loggers, the first argument is always the package name and the second one is the namespace for the current file, the two of which get separated by a colon when creating the `Debugger` object as you described. The respective log level (e.g., `debug`) is always appended at the end, so that the full namespace of the `debug` logger above would become `core:servient:debug`, while the full `warn` namespace would be `core:servient:warn`.

I generally have a good feeling about this new solution, and I am looking forward to your feedback :)

relu91 (changes requested):
Thanks! I think we are heading in the right direction 🥳 two minor things and then from my point of view we can merge:
1. In `debug.ts` I suggest using variadic arguments so that user can use more than two nesting levels (e.g. `createLoggers("core", "contet-encoders", "json") `) 
2. I would prepend always the string `node-wot` so that we will not conflict with other libraries.

JKRhb:
> Thanks! I think we are heading in the right direction 🥳 two minor things and then from my point of view we can merge:
> 
> 1. In `debug.ts` I suggest using variadic arguments so that user can use more than two nesting levels (e.g. `createLoggers("core", "contet-encoders", "json") `)
> 2. I would prepend always the string `node-wot` so that we will not conflict with other libraries.

Thank you, @relu91, these are very good additions :) I added them in the latest commits.

JKRhb:
I took the freedom of already squashing the commits :) Let me know, however, if you see anything that could still be improved.

relu91 (approved):
🥳 Thank you! I think we can go ahead, we should remember the update the README.md but we can wait until we finish the migration to the new configuration in all bindings.

danielpeintner (approved):

Issue #785, opened 2022-06-12
Writing to properties fails in browser


fatadel:
Create `foo.html` with the following code:

```
<script src="https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@latest/dist/wot-bundle.min.js"></script>
<script>
    const servient = new Wot.Core.Servient();
    servient.addClientFactory(new Wot.Http.HttpClientFactory());
    const helpers = new Wot.Core.Helpers(servient);
    servient.start().then((thingFactory) => {
        helpers
            .fetch('http://plugfest.thingweb.io:8083/testthing')
            .then((td) => {
                thingFactory.consume(td).then((thing) => {
                    thing.writeProperty('bool', false);
                });
            })
            .catch((error) => {
                window.alert("Could not fetch TD.\n" + error);
            });
    });
</script>
```

Expected result:
No errors, successful write operation.

Actual result:
Error in console, write operation not performed.

<img width="1424" alt="image" src="https://user-images.githubusercontent.com/14371560/173258886-4fce085e-b2f6-43c7-b7e4-734d6033d681.png">

sebastiankb:
just for record: the [webui app](http://plugfest.thingweb.io/webui/) has the same problem when I call an action:

![image](https://user-images.githubusercontent.com/13832739/175528766-37e9b0a3-7232-4286-a2f7-54dc6a54e5bc.png)

danielpeintner:
@relu91 I am not sure if the issue is *really* solved.
I tried the WebUI and I still run into issues

> Error: Readable.from is not available in the browser

Note:  It is different to the one @sebastiankb and others were seeing before...

To reproduce the problem

1. Build node-wot
2. Open `<node-wot-folder>/examples/browser/index.html` locally in a browser and consume http://plugfest.thingweb.io:8083/testthing
3. Invoke "int-int" action

danielpeintner:
FYI: I updated the online version also.
see http://plugfest.thingweb.io/webui/

PR #789, opened 2022-06-29
Fix writing property on browsers 

relu91:
This PR focuses on the resolution of issue #785. Due to https://github.com/whatwg/fetch/issues/1438 we can't use streams when writing property or invoking an action. Therefore, these changes introduce a specific HttpClient for browsers that override the generateFetchRequest method to read the stream in a buffer. 

Moreover, the PR fixes smaller issues:
- The HTTP server correctly answers CORS requests for PUT methods in the properties path
- When a `Readable` is forcibly closed the underlying Readable is correctly closed and the error is propagated. 
- `Readable.from` is not available in readable-stream v3.x.x so now we are forcing browserify to use the new v4.x.x. That's why in the PR you can see a new build.js script. Its purpose is to properly set up browserify. I couldn't find a way to achieve the same effect using just CLI arguments.


relu91:
Being just a bug fix I'd go ahead and merge it.

PR #792, opened 2022-06-30
Strictly check for undefined in getClientFor

relu91:
Notice that I've tried to add a test for this change but the current test suite is at a very high abstraction level. At that level, the bug is negligible since if you provide 0 as formIndex the code will end up choosing the first form anyway. Unless the first form contains a form that has a client but in this case, the operation will fail anyway.


danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/788

Issue #771, opened 2022-05-25
Crash if time passes between "readProperty" and "value" function invocation.


FerrariAndrea:
Hi,
Considering the two following line of code, if we do something that takes time between these the service crash with no error.
First line of code:
`const reader = await thing.readProperty("prop_name");`
Second line of code: 
`const ris = await reader.value();`
Otherwise, if we call these two functions one after the other all goes well.

Here you can find the full snippet of code that I'm using to see the bug:
```ts
 const td_url = "http://plugfest.thingweb.io:8083/counter";
   try {
        const servient = new Servient();
        servient.addClientFactory(new HttpClientFactory(undefined));
        const wot = await servient.start();
        const td = (await axios.get(td_url)).data;
        const thing = await wot.consume(td as ThingDescription);
        console.log("BEFORE readProperty");
        const reader = await thing.readProperty("count");
        console.log("AFTER readProperty");
        //await delay(5000); //<--DECCOMENT FOR SEE THE BUG
        console.log("Here we have the console.log");
        console.log("BEFORE value");
        const ris = await reader.value(); //here we have the bug
        console.log("AFTER value");
        console.log("Here we haven't the console.log and for some reason, the script exit with no error (skipping the catch too)."); 
        console.log("WotTest result: ", ris);
    } catch (err) {
        console.log("WotTest: ResolveTD error 1:", err);
    }
```

Following the output with the 11° line commented (without a delay between the two functions):
```
BEFORE readProperty
[core/helpers] Helpers found scheme 'http'
[core/helpers] Helpers found scheme 'http'
[core/helpers] Helpers found scheme 'coap'
[core/consumed-thing] ConsumedThing 'counter' has no client in cache (-1)
[core/servient] Servient checking for 'http' scheme in 1 ClientFactories
[core/servient] Servient creating client for scheme 'http'
[binding-http] HttpClientFactory creating client for 'http'
[core/consumed-thing] ConsumedThing 'counter' got new client for 'http'
[core/consumed-thing] ConsumedThing 'counter' setting credentials for [HttpClient] based on thing security
[core/servient] Servient looking up credentials for 'urn:uuid:1a32da56-a9ba-434a-a049-aceb883586fa' (@deprecated)
[binding-http] HttpClient using security scheme 'nosec'
[core/consumed-thing] ConsumedThing 'counter' reading http://plugfest.thingweb.io:8083/counter/properties/count
[...]

PR #793, opened 2022-06-30
Create WoTStream lazily

relu91:
Now InteractionOutput creates a WoTStream only when the data property is accessed. This avoids pulling data from the underlying stream before it is been used. Previously this caused #771. I also added two tests to verify the correct behavior. 

One note. Now if `InteractionOutput.data` is called the `InteractionOutput.value` function will refuse to read the stream even if the underlying stream was not read by the caller. This is due to the fact that `ReadableStream` has a mechanism to automatically pull data even if nobody is reading creating an unpredictable behavior.

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -32,20 +32,23 @@ export class InteractionOutput implements WoT.InteractionOutput {
     private content: Content;
     private parsedValue: unknown;
     private buffer?: ArrayBuffer;
-    data?: ReadableStream;
+    private _stream?: ReadableStream;
     dataUsed: boolean;
     form?: WoT.Form;
     schema?: WoT.DataSchema;
+
+    public get data(): ReadableStream {
+        if (this._stream) {
+            return this._stream;
+        }
+        return (this._stream = ProtocolHelpers.toWoTStream(this.content.body) as ReadableStream);
+    }
 
     constructor(content: Content, form?: WoT.Form, schema?: WoT.DataSchema) {
         this.content = content;
         this.form = form;
         this.schema = schema;
         this.dataUsed = false;
-
-        if (content && content.body) {
-            this.data = ProtocolHelpers.toWoTStream(content.body) as ReadableStream;
-        }
     }
 
     async arrayBuffer(): Promise<ArrayBuffer> {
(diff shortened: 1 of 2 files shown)


packages/core/src/interaction-output.ts
```diff
@@ -32,20 +32,23 @@ export class InteractionOutput implements WoT.InteractionOutput {
     private content: Content;
     private parsedValue: unknown;
     private buffer?: ArrayBuffer;
-    data?: ReadableStream;
+    private _stream?: ReadableStream;
```
danielpeintner, line 35: Is this variable used somewhere?

danielpeintner:
> One note. Now if `InteractionOutput.data` is called the `InteractionOutput.value` function will refuse to read the stream even if the underlying stream was not read by the caller. This is due to the fact that `ReadableStream` has a mechanism to automatically pull data even if nobody is reading creating an unpredictable behavior.

Some comments:
* I am still not sure about the consequences. 
* Shouldn't the *new* data() function set `dataUsed` to `true` right away?

packages/core/src/interaction-output.ts
relu91: yes it is used inside the data getter

danielpeintner: OMG, sorry 🙈

packages/core/src/interaction-output.ts
```diff
@@ -32,20 +32,23 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
     dataUsed: boolean;
     form?: WoT.Form;
     schema?: WoT.DataSchema;
 
+    public get data(): ReadableStream {
+        if (this._stream) {
```
danielpeintner, line 41: `this.dataUsed = true; ` // don't we need to set that ?

packages/core/src/interaction-output.ts
```diff
@@ -70,6 +73,10 @@ export class InteractionOutput implements WoT.InteractionOutput {
             throw new NotSupportedError(message);
         }
 
+        if (this._stream) {
+            throw new Error("Can't call value function after retrieving the stream");
+        }
```
danielpeintner, line 78: I think this error needs to be added to both functions

`arrayBuffer() ` AND `value<T>()`

packages/core/src/interaction-output.ts
relu91: Technically, it should be set only after somebody reads the stream... but given the behavior above, we could do that!

danielpeintner (approved):

PR #794, opened 2022-06-30
fix CLI compilation errors due to #758

relu91:
Another step.


danielpeintner:
Note: I checked "cli" in https://github.com/eclipse/thingweb.node-wot/issues/758

PR #796, opened 2022-07-06
chore: use cache in CI

relu91:
This morning I read a Twitter thread from @mcollina, about caching node:_modules to speed up CI runs. Since our runs last already an eternity I think we can try this minor tweak. Let's see if it improves the situation. :)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,7 +29,7 @@ jobs:
               run: npm run build
 
             - name: Test with coverage report
-              run: npm run coverage
+              run: npm run coverage:only
 
             - name: Upload to codecov.io
               uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


danielpeintner:
CI was failing. Restarted them...

danielpeintner:
@relu91 how yo you want to continue with this PR?

relu91:
if of all I would like to understand if it is properly working. I was expecting a considerable boost given the fact that in our repository `npm ci` takes a while. However, I am not seeing such a difference between CI runs:
image
🔝Using cache
image
🔝Without cache

relu91:
Digging deeper, the `cache` option does not cache node_modules but the good news is that in our setup `npm ci` does not take THAT much time. On average the install step takes around 1m. The problem is the test step that takes the most part of the time  ~4min. I think we can improve a little bit by skipping the build in that step since we have already built in the previous one.

relu91:
It seems that the last change didn't improve the execution time either. I guess we just have very slow tests.... 😢

danielpeintner:
@relu91 how should we go on with this PR?

relu91:
> @relu91 how should we go on with this PR?

It does not solve that much but we can merge it. We have a problem with our slow tests. I'll open an issue to keep track of this possible enhancement.

PR #798, opened 2022-07-07
update wot-thing-description-types and wot-thing-model-types

danielpeintner:
Version 1.1.0-06-July-2022


relu91 (approved):
I'm glad that it was painless this time 🤣

PR #799, opened 2022-07-08
Reorganising CLI Readme information

egekorkan:
I am moving CLI readme to its own package's readme and thus tidying up the main one. 

I have started in the web editor but it is not easy to check for prettier etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,48 +203,7 @@ You can then start the applications with node by running `node server.js` and `n
 
 #### CLI Tool
 
-You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool's Readme]([url](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/cli)) to find out more.
-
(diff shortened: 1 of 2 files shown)


egekorkan:
Prettier check fails even though I am executing `npm run lint`

relu91:
I think that the CI is not picking our prettier configuration: https://github.com/eclipse/thingweb.node-wot/blob/master/.github/workflows/ci.yaml#L70

egekorkan:
> In my case running `npm run format` locally created some changes, see [bf03fcb](https://github.com/eclipse/thingweb.node-wot/commit/bf03fcba9860c190d1c3a4248e43f2dd762b1366).
> 
> @egekorkan please check whether the result is as expected.

All fine with me:) Also, it seems that the action is not complaining

danielpeintner (approved):

relu91 (approved):
LGTM, I'm not sure why the text about log level has been removed, but given the fact that we are going to migrate to "debug" library it should not be a problem. Anyhow we can fix it later. Let's go :)

PR #804, opened 2022-07-14
Update npm wot thing description types 2022 07 13

danielpeintner:
Note: I am sometimes **VERY** surprised how _unrelated_ changes (see https://github.com/w3c/wot-scripting-api/commit/8c73b707462b494787d2d261e8f09b7d70dadcfc) break the types (`@context/@language`) used in node-wot.

Does anyone know why? Kind of frustrating.

### DIFF
--- a/packages/binding-http/package.json
+++ b/packages/binding-http/package.json
@@ -63,7 +63,7 @@
         "query-string": "^7.1.1",
         "rxjs": "5.5.11",
         "slugify": "^1.4.5",
-        "wot-thing-description-types": "^1.1.0-06-July-2022"
+        "wot-thing-description-types": "^1.1.0-13-July-2022"
     },
     "scripts": {
         "build": "tsc -b",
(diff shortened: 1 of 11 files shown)


danielpeintner:
Can someone take a look? @relu91 
It essentially just updates the dependencies and refactors the method of setting the language in *one* helper function to avoid touching several places in the future

relu91 (approved):
Sorry @danielpeintner for the delay I wanted to investigate a little bit why the new context was causing problems with our previous code. It seems that the new data schema for `@context` is pretty complex and it seems that the translator library is doing what it can but the type is a little bit messed up:
```ts
const arrayContext: [] | ["https://www.w3.org/2022/wot/td/v1.1", ...(string | {
    [k: string]: unknown;
})[]] | [unknown, unknown, ...unknown[]] | [unknown, ...unknown[]]
``` 
The last two types are basically `unknown[]` probably we should experiment and understand why it is producing that type.  But, for the time being, I am ok to merge this PR!

relu91:
Also, the type includes `[]` which should not be valid cause the content of `@context` must not be empty.

danielpeintner:
> Sorry @danielpeintner for the delay I wanted to investigate a little bit why the new context was causing problems with our previous code. It seems that the new data schema for `@context` is pretty complex and it seems that the translator library is doing what it can but the type is a little bit messed up:

I agree, the type looks rather strange.
I am sure there is a way to overrule a given type but once we start doing that we are getting back to where we started..

PR #808, opened 2022-07-17
Support observeProperty in binding-firestore

hidetak:
I was able to deal with it only by modifying the test, example code and README.

### DIFF
--- a/packages/binding-firestore/src/firestore-handler.ts
+++ b/packages/binding-firestore/src/firestore-handler.ts
@@ -74,7 +74,7 @@ export const writeDataToFirestore = async (
     content: Content,
     reqId: string = null
 ): Promise<void> => {
-    console.debug("[debug] writeDataToFirestore topic:", topic, " value:", content, reqId);
+    console.debug("[debug] writeDataToFirestore topic:", topic, reqId);
     const ref = firestore.collection("things").doc(encodeURIComponent(topic));
     const data = { updatedTime: Date.now(), reqId, content: "" };
     if (content && content.body) {
--- a/packages/binding-firestore/src/firestore-handler.ts
+++ b/packages/binding-firestore/src/firestore-handler.ts
@@ -20,19 +20,21 @@ import "firebase/compat/firestore";
 import { Readable } from "stream";
 import { BindingFirestoreConfig } from "./firestore";
 
-let firebase: any;
+let firebase: typeof Firebase;
 if (Firebase.apps) {
     // for NodeJS
     firebase = Firebase;
 } else {
     // for Web browser (We'll deal with it later.)
-    firebase = window.firebase;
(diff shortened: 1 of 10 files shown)


packages/binding-firestore/src/firestore-handler.ts
```diff
@@ -74,7 +74,7 @@ export const writeDataToFirestore = async (
     content: Content,
     reqId: string = null
 ): Promise<void> => {
-    console.debug("[debug] writeDataToFirestore topic:", topic, " value:", content, reqId);
+    console.debug("[debug] writeDataToFirestore topic:", topic, reqId);
```
danielpeintner, line 77: We seemed to have missed it before but we usually add the "protocol" so that we know where the log comes from and we can filter it..

e.g., `console.debug("[binding-http]", `HttpServer ...`);`

Do you think we can do this for firestore too?

danielpeintner (commented):
Thanks for your update!

Looks good but I added a logging comment and wanted to ask you about resolving warnings like
"Unexpected any. Specify a different type".

We try to keep the code warning free. If really needed you can also an ignore warning. 
Thanks!

packages/binding-firestore/src/firestore-handler.ts
hidetak: @danielpeintner 

Thank you for giving comment !

> Looks good but I added a logging comment and wanted to ask you about resolving warnings like
"Unexpected any. Specify a different type".

How can I check the above message?

> Do you think we can do this for firestore too?

OK. I will modify the log to output `[binding-firestore]`.

danielpeintner: > > Looks good but I added a logging comment and wanted to ask you about resolving warnings like
> > "Unexpected any. Specify a different type".
> 
> How can I check the above message?

Visual Studio code should show you these warnings.
Besides that, running `npm run lint` in the root folder or for example in `packages/binding-firestore` will show you these warnings as well.
Moreover, the "Files changed" tab of Github highlights the same issues

> > Do you think we can do this for firestore too?
> 
> OK. I will modify the log to output `[binding-firestore]`.

👍

danielpeintner: W.r.t. debugging.

We are about to change debug module for logging.
see https://github.com/eclipse/thingweb.node-wot/pull/783/

danielpeintner (approved):
LGTM.

We can update the logging in a follow-up PR.
Note: we are bout to change logging to use the debug library.

@relu91 any concerns before merging?

danielpeintner:
checked with @relu91. Good to go.

hidetak:
@danielpeintner 
Thank you for checking and merging !

danielpeintner:
@hidetak thank you for your contribution.

PR #809, opened 2022-07-27
Correctly configure built-ins in browserify

relu91:
Sorry, it seems that when I cleaned up the `built.js` script I wrongly assigned the list of builtins to the root configuration object of browserify. Silly mistake. Now everything should work correctly on the browsers. 

fixes https://github.com/eclipse/thingweb.node-wot/issues/785


danielpeintner (approved):
I checked  the update with the browser UI sample provided in https://github.com/eclipse/thingweb.node-wot/issues/785#issuecomment-1173949837 and it works fine now.

👍

Issue #795, opened 2022-07-04
Configure/define the URL structure of an ExposedThing


danielpeintner:
At the moment the ExposedThing (or the actual binding) defines the structure of a thing.
e.g. for HTTP, 
* `<thing-name>/properties/<property-name>` for properties
* `<thing-name>/actions/<action-name>` for actions
* `<thing-name>/events/<event-name>` for events

There exist cases where the structure is predefined (by a standard or another document etc) and it should be possible in node-wot to build the URL structure as required.

I was wondering "what" would be easiest way to fulfill this requirement. Are there any proposals/ideas?

_I think offering a way to configure every endpoint might be cumbersome and leads to issues._

_Simpler_ alternatives I see are:
* provide an "url-rewrite" configuration/mechanims in node-wot for each binding (or at least to HTTP and CoAP) that defines for example some simple mappings like
"/building-abc/room-33/click-counter/v1/property-count" --> "/counter/properties/count"
Note1: with this approach we might run into name collisions. 

* do it *OUTSIDE* of node-wot --> one can (and MUST) use another HTTP server with a different port and can refer/forward to any URL.

Note2: offering just an alternative URL for another URL endpoint does not update a TD (this may or may not be an issue)

egekorkan:
Option 2 can be done anytime. I think option 1 would be a nice optional configuration. I think it can be done via the TD given into the produce method or via the binding configuration. We have similar ways for some other configuration so it can be up to taste.

sebastiankb:
@egekorkan 
>I think option 1 would be a nice optional configuration. I think it can be done via the TD given into the produce method

 I was also thinking about this. The developer needs to provide the `forms` and `href` entries when no default URL pattern is desired.

egekorkan:
Option 1 is not bad but we have to make sure that only relative URLs are used since the host part of the URL comes from the servient and not from the binding (which manages the forms href). There was previously a discussion about this at #632 . I propose to keep only one issue and move the comments from to the other.

PR #810, opened 2022-07-27
URL rewrite

danielpeintner:
This is a proof of concept to introduce URL rewrite capabilities in node-wot (i.e., using a desired URL for a given resource);

urlRewrite takes precedence and rewrites any possible "input" URL.
I tried to keep the changes as minimal as possible.

Moreover, I modified an example to show how this might look like in practice. You can run and play with it of course.

Questions/Downsides:
* One needs to know the *internals* of node-wot to know to which URL to point to.
  (e.g,. `"/myroot/foo/cnt"`  -->  `"/mycounter/properties/count"`)
* The ThingDescription itself contains the actual URL. Do we need to update the TD as well? What if multiple URLs point to the same resource?
* We do not differ between GET/PUT/POST request. Do we need to do so?

TODOs
* update [wot-servient-schema.conf.json](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/cli/src/wot-servient-schema.conf.json)
* update CLI ([1](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/cli/src/cli.ts#L61), ...)

Feedback is welcome!

fixes https://github.com/eclipse/thingweb.node-wot/issues/795

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -285,6 +285,23 @@ export default class HttpServer implements ProtocolServer {
         });
     }
 
+    private addUrlRewriteEndpoints(form: TD.Form, forms: Array<TD.Form>): void {
+        if (this.urlRewrite) {
+            for (const inUri in this.urlRewrite) {
+                const toUri = this.urlRewrite[inUri];
+                if (form.href.endsWith(toUri)) {
+                    const form2: TD.Form = JSON.parse(JSON.stringify(form)); // deep copy
+                    form2.href = form2.href.substring(0, form.href.lastIndexOf(toUri)) + inUri;
+                    forms.push(form2);
+                    console.debug(
+                        "[binding-http]",
+                        `HttpServer on port ${this.getPort()} assigns urlRewrite '${form2.href}' for '${form.href}'`
+                    );
+                }
+            }
+        }
+    }
+
     public addEndpoint(thing: ExposedThing, tdTemplate: WoT.ExposedThingInit, base: string): void {
         for (const type of ContentSerdes.get().getOfferedMediaTypes()) {
             let allReadOnly = true;
@@ -317,6 +334,7 @@ export default class HttpServer implements ProtocolServer {
                     thing.forms = [];
                 }
                 thing.forms.push(form);
+                this.addUrlRewriteEndpoints(form, thing.forms);
             }
 
             for (const propertyName in thing.properties) {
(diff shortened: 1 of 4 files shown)


danielpeintner:
@sebastiankb please let me know whether the following configuration allows you to do what you asked for in https://github.com/eclipse/thingweb.node-wot/issues/795

```
servient.addServer(
    new HttpServer({
        port: 8081, // (default 8080)
        urlRewrite: { // create additional *fake* entry-points by forwarding it to the actual endpoint
            "/myroot/foo/cnt": "/mycounter/properties/count",
            "/mysecondURL/for/the/same/resource": "/mycounter/properties/count",
        },
    })
);
```

relu91 (approved):
I like the minimal impact of this change 👍🏻. I would go ahead but I'd keep #795 open, IMHO we should define something that directly works with `forms` from `ExposeThingInit`

danielpeintner:
Updated the PR:
* *new* endpoints get added to ThingDescription
* added simple test for urlRewrite feature

Question: shall we add urlRewrite to CLI as well?
I noticed there are more HTTP configurations missing there?
* address
* baseUri and
* urlRewrite

We could do this in a follow-up PR...

see https://github.com/eclipse/thingweb.node-wot/blob/07494392e1def691f06adeae4224afac7cd4cd65/packages/cli/src/wot-servient-schema.conf.json#L32-L62

danielpeintner:
@relu91 @sebastiankb 
PR should be ready now.

It allows URL Rewriting and it adds these new URLs endpoints to the TD as well...

relu91:
I noticed this comment from @egekorkan : 
> Option 1 is not bad but we have to make sure that only relative URLs are used since the host part of the URL comes from the servient and not from the binding (which manages the forms href). There was previously a discussion about this at https://github.com/eclipse/thingweb.node-wot/issues/632 . I propose to keep only one issue and move the comments from to the other.

I think that the PR missing the validation side. In my understanding, it is possible to write something like `{ "/properties/test" : "http://hello.com/test" }`, which is something not desirable. Should we validate both In URI and out URI to be relative references? Maybe we can do this in a future PR.

egekorkan:
It can definitely be in a future PR but such a validation is needed otherwise people will be confused with the behavior

danielpeintner:
> It can definitely be in a future PR but such a validation is needed otherwise people will be confused with the behavior

I created https://github.com/eclipse/thingweb.node-wot/issues/828.
I added my personal opinion there..

PR #813, opened 2022-08-02
Use new logging functions in binding packages

JKRhb:
This PR applies the new logging approach to the binding packages. I was a bit unsure about the examples, which is why I left them out for now. Moreover, the logging in the `td-tools` package is a bit provisional for now, as I could not import the logging functions from the `core` package, as this would cause a circular dependency.

In the other packages, however, logging should be much improved and cleaner already. The fact that it needs to turned on explicitly also makes the CI logs much more readable.

### DIFF
--- a/packages/binding-coap/src/coap-client-factory.ts
+++ b/packages/binding-coap/src/coap-client-factory.ts
@@ -17,9 +17,11 @@
  * CoAP client Factory
  */
 
-import { ProtocolClientFactory, ProtocolClient } from "@node-wot/core";
+import { ProtocolClientFactory, ProtocolClient, createLoggers } from "@node-wot/core";
 import CoapClient from "./coap-client";
 import CoapServer from "./coap-server";
+
+const { debug } = createLoggers("binding-coap", "coap-client-factory");
 
 export default class CoapClientFactory implements ProtocolClientFactory {
     public readonly scheme: string = "coap";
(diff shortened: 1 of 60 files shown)


JKRhb:
(Sorry for the many CI runs – every run except the last one could actually be canceled.)

relu91 (approved):
🤩🤩thank you @JKRhb !!! I guess it was a tedious job but it was very well needed 🥳 . I would go ahead and merge your contribution when CI is happy. However, it seems that you didn't update the `setLogLevel` function on the `cli` package. We should probably talk about it in an issue.

JKRhb:
Thank you, @relu91 :)

> However, it seems that you didn't update the setLogLevel function on the cli package. We should probably talk about it in an issue.

Yeah, I was a bit unsure about these functions, as it seemed to me as if the log level should cause messages to be printed out for the user independently of their knowledge about the debug module/environment variables. If we switched to using `debug` here as well, the `setLogLevel` function/action could become obsolete. As this seems to require a bit more discussion, opening an issue is definitely a good idea :)

danielpeintner:
I created https://github.com/eclipse/thingweb.node-wot/issues/814 to keep track of the discussion.

Note: I am fine with merging but I would prefer to merge https://github.com/eclipse/thingweb.node-wot/pull/808 first since it changes firestore binding. I assume this will cause merge conflicts since both touch firestore code...

JKRhb:
> Note: I am fine with merging but I would prefer to merge #808 first since it changes firestore binding. I assume this will cause merge conflicts since both touch firestore code...

No problem, I can rebase this PR against the master branch once #808 is merged :)

danielpeintner:
> No problem, I can rebase this PR against the master branch once #808 is merged :)

@JKRhb as mentioned, merging https://github.com/eclipse/thingweb.node-wot/pull/808 caused conflicts...

JKRhb:
> > No problem, I can rebase this PR against the master branch once #808 is merged :)
> 
> @JKRhb as mentioned, merging #808 caused conflicts...

I rebased onto the current master branch, the conflicts should be resolved now :)

danielpeintner (approved):

PR #817, opened 2022-08-10
resolves CLI warnings

danielpeintner:


packages/cli/src/cli.ts
```diff
@@ -205,6 +205,7 @@ const args = program.args;
 
 // .env parsing
 const env: dotenv.DotenvConfigOutput = dotenv.config();
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 if (env.error && (env.error as any).code && (env.error as any).code !== "ENOENT") {
```
relu91, line 209: Can't we cast to `Error`?

relu91 (changes requested):

packages/cli/src/cli.ts
danielpeintner: Unfortunately not...

`error TS2339: Property 'code' does not exist on type 'Error'`

I also tried more dedicated errors extending `Error` interface... but wasn't successful to find one.. not sure which kind does have `code`.
Do you?

relu91: https://stackoverflow.com/a/49562477/925955 what about this answer? is it pretty old and I'm quite puzzled that it wasn't fixed... but still worth trying?

danielpeintner: Good finding. That works!
I updated the PR accordingly...

relu91 (approved):

PR #818, opened 2022-08-10
disable eslint warnings

danielpeintner:


relu91 (approved):
Not ideal, but given the complexity required to fix it properly, I can live with it.

PR #821, opened 2022-08-11
docs: update copyright year based on git log

danielpeintner:


relu91 (approved):
Easy one, merging right away

PR #833, opened 2022-09-16
update node-opcua to v2.77.0

erossignon:


danielpeintner:
fails, not sure why... created https://github.com/eclipse/thingweb.node-wot/pull/832 with a fresh package-lock file..

PR #840, opened 2022-09-26
fix: avoid wrapping Content into Content

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/838


danielpeintner:
@relu91 @egekorkan please have a look

egekorkan:
> @relu91 @egekorkan please have a look

Looks a bit weird to me but together with the issue text it makes sort of sense. I have lost track of this part of the code since some time ago :disappointed:

relu91 (approved):
Probably it was a left over of the process that converted everything into streams. Good to go!

PR #844, opened 2022-10-14
feat(binding-coap): add support for TD Content-Format negotiation and CoRE Link-Format

JKRhb:
This PR adds support for two features from the Discovery specification for the CoAP server implementation.

On the one hand, CoAP clients can perform a simple Content-Format negotiation, resulting in either `application/json` (Content-Format 50) or `application/td+json` (Content-Format 432), depending on a request's `Accept` option. If no `Accept` option should be set or the value is not supported, the server falls back to the default value `application/td+json`.

On the other hand, I added support for the CoRE Link-Format and the well-known URI `/.well-known/core` ([RFC 6690](https://datatracker.ietf.org/doc/html/rfc6690) where a list of links of all hosted TDs is exposed. In the process, I defined a couple of simple interfaces which could potentially be used to add more parameters or links (e.g., regarding individual affordances) in the future.


packages/binding-coap/src/coap-server.ts
```diff
@@ -197,6 +215,52 @@ export default class CoapServer implements ProtocolServer {
...
+    }
+
+    private negotiateContentFormat(
+        req: IncomingMessage,
+        res: OutgoingMessage,
+        availableContentFormats: string[],
+        defaultContentFormat: string
+    ) {
+        const accept = req.headers.Accept;
+        console.log(accept);
```
relu91, line 254:
```suggestion
```

relu91 (changes requested):
I like the minimal impact of this feature, well done @JKRhb. I spotted a minor problem below, we can commit and merge if @danielpeintner is ok! This would imporve a little bit the discovery 
implementation report. Can I ask you to write a couple of tests to verify content negotionation and proper answers for the core-link format?

packages/binding-coap/src/coap-server.ts
danielpeintner: Agree, remove or add proper logging

danielpeintner:
> I like the minimal impact of this feature, well done @JKRhb. I spotted a minor problem below, we can commit and merge if @danielpeintner is ok!

Looks good to me... fixing now or later... both seems fine to me.

packages/binding-coap/src/coap-server.ts
JKRhb: Oh, thank you, that was only for debugging purposes 😅

JKRhb:
Thank you for the feedback! I now added both a couple of debug statements and also test cases for both new features :)

JKRhb:
Hmm, the firestore tests failed for some reason. Could you restart the last workflow run?

packages/binding-coap/test/coap-server-test.ts
```diff
@@ -251,4 +253,88 @@ class CoapServerTest {
...
+                    type: "string",
+                    forms: [],
+                },
+            },
+        });
+        testThing.setPropertyReadHandler("test", (options) => {
+            return new Promise<InteractionInput>((resolve, reject) => {
+                resolve(null);
+            });
+        });
```
relu91, line 276: I think you can remove this read handler together with the TD test property for simplicity. 
```suggestion
      
```

relu91 (changes requested):
Thank you! another minor point below and let's hope for a CI greenlight

packages/binding-coap/test/coap-server-test.ts
JKRhb: Oh, you are right! I removed the handlers and the unneeded properties :)

relu91 (approved):
Good to go. 🟩

danielpeintner (approved):

PR #845, opened 2022-10-17
chore: update TS versions for TD and TM

danielpeintner:


danielpeintner:
We have issues in browser-bundle... grr
I had the locally too but was hoping it is a caching issue... need to look into

relu91:
What about this error: https://github.com/eclipse/thingweb.node-wot/actions/runs/3263939568/jobs/5363790618 ?

danielpeintner:
The issue with browserify remains :-(
https://github.com/eclipse/thingweb.node-wot/actions/runs/3273332391/jobs/5385413069

danielpeintner:
The error message is as follows (at least locally)

```
> build
> tsc -b && npm run build -w packages/browser-bundle

error TS2688: Cannot find type definition file for 'bl'.
  The file is in the program because:
    Entry point for implicit type library 'bl'

error TS2688: Cannot find type definition file for 'bl'.
  The file is in the program because:
    Entry point for implicit type library 'bl'

Found 2 errors.
```

https://www.npmjs.com/package/@types/bl seems to be used by coap.

@JKRhb maybe you know whether those types are used somewhere not properly... I might be wrong but maybe you have a hint..

JKRhb:
Hmm, I tried to reproduce the error, @danielpeintner, but I haven't seen any issues regarding `bl` locally :/ Instead I am getting the same errors as in the CI job.

However, one of the issues linked in https://github.com/browserify/browserify/issues/1986 seems to have a similar problem (see https://github.com/rdfjs/N3.js/issues/310), related to `readable-stream`. That module got recently updated to the latest version in `node-coap`, so my guess would be that this could be the root cause here :/

JKRhb:
Thanks for the hint regarding the `bl` types! I will remove them from `node-coap` shortly -- maybe that already resolves the problem :)

danielpeintner:
node-coap v1.2.2 seems to have resolved the coap issue. Thanks @JKRhb!

danielpeintner:
firestore fails. Other than that I think this PR is **finally**  fine now.

@relu91 ?

PR #846, opened 2022-10-17
feat(binding-coap): add support for Size2 estimate

JKRhb:
In order to also support the assertion [`exploration-server-coap-size2`](https://w3c.github.io/wot-discovery#exploration-server-coap-size2), this PR adjusts the CoAP binding to be able to include the option in responses. However, there are a couple of bugs which prevent this PR from being complete at the moment, therefore I opened it only as a draft for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2645,7 +2645,6 @@
             "version": "3.0.0",
             "resolved": "https://registry.npmjs.org/abort-controller/-/abort-controller-3.0.0.tgz",
             "integrity": "sha512-h8lQ8tacZYnR3vNQTgibj+tODHI5/+l06Au2Pcriv/Gmet0eaj4TwWH41sO9wnHDiQsEj19q0drzdWdeAHtweg==",
-            "dev": true,
             "dependencies": {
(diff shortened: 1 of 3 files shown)


JKRhb:
Hmm, in the meantime, I arrived at the conclusion that this feature should probably be added to node-coap itself. I will adjust the PR accordingly once node-coap has better Size2 support.

JKRhb:
With https://github.com/mcollina/node-coap/pull/351 merged, the CoAP server should now support the Size2 option by default :) However, for some reason the test is currently failing when trying to use block-wise transfer :/ I will investigate this a bit further.

JKRhb:
Found the bug and opened a PR with a fix in node-coap. Once that one is merged and version 1.2.1 of node-coap is released, we can go forward with this PR :)

JKRhb:
Updated node-coap to a version containing the fix and adjusted the test case :)

JKRhb:
Do you think we can move forward with this? :)

relu91 (approved):

PR #849, opened 2022-10-19
fix(binding-coap): send error response if content negotiation fails

JKRhb:
Dealing a bit more with CoAP content negotiation, I noticed in the [RFC](https://www.rfc-editor.org/rfc/rfc7252#section-5.10.4) that instead of using a default Content-Format in the case of failed negotiation, an error response ~should~ must be sent. This PR takes that into account, adjusting the internal `negotiateContentFormat` method and the test. Since this caused the logic to become a bit more complicated, I added some documentation to hopefully clarify how the new, now RFC-compliant process works.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -244,23 +244,49 @@ export default class CoapServer implements ProtocolServer {
         }
     }
 
-    private negotiateContentFormat(
-        req: IncomingMessage,
-        res: OutgoingMessage,
-        availableContentFormats: string[],
-        defaultContentFormat: string
-    ) {
-        const accept = req.headers.Accept;
-
-        if (typeof accept === "string" && availableContentFormats.includes(accept)) {
-            debug(`Received available Content-Format ${accept} in Accept option.`);
-            res.setHeader("Content-Format", accept);
+    /**
+     * Handles a CoAP request for an ExposedThing, negotiates the TD Content-Format and sends
+     * a response.
+     *
+     * If a specific Content-Format for the TD is requested by a client, as indicated by
+     * an Accept option, it will be set for the outgoing response if it is supported.
+     * If no Accept option is set, the default Content-Format will be used as a fallback.
+     *
+     * If an Accept option is present but the Content-Format is not supported, the response
+     * will be sent with a status code `4.06` (Not Acceptable) and an error
+     * message as a diagnostic payload in accordance with RFC 7252, sections 5.10.4 and
+     * 5.5.2.
+     *
+     * @param req The incoming request.
+     * @param res The outgoing response.
+     * @param thing The ExposedThing whose TD is requested.
+     */
+    private async handleTdRequest(req: IncomingMessage, res: OutgoingMessage, thing: ExposedThing) {
+        if (req.method !== "GET") {
+            res.code = "4.05";
+            res.end("Method Not Allowed");
             return;
         }
 
-        debug("Request did not contain an accept option or Content-Format is not supported.");
+        const accept = req.headers.Accept;
 
-        res.setHeader("Content-Format", defaultContentFormat);
+        const contentSerdes = ContentSerdes.get();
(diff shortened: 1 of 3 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -341,14 +371,17 @@ export default class CoapServer implements ProtocolServer {
...
                             res,
                             [ContentSerdes.DEFAULT, ContentSerdes.TD],
                             ContentSerdes.TD
                         );
-                        res.code = "2.05";
-                        res.end(JSON.stringify(thing.getThingDescription()));
+
+                        if (negotiationSuccessful) {
+                            res.code = "2.05";
+                            res.end(JSON.stringify(thing.getThingDescription()));
```
ektrah, line 383: Could the whole function could be simplified like this?

```JS
switch (accept) {
    case ContentSerdes.DEFAULT:
    case ContentSerdes.TD:
    case null:
        res.code = "2.05";
        res.setHeader("Content-Format", accept || ContentSerdes.TD);
        res.end(JSON.stringify(thing.getThingDescription()));
        break;
    default:
        res.code = "4.06";
        res.end(`Content-Format ${accept} is not supported by this resource.`);
        break;
}
```

JKRhb: Hmm, yeah, I incorporated your suggestion in d1f3ae1965a9098051a345ac98682ff1d1ad6a35. My initial thoughts were that this method could be a generic one for content negotiation, but maybe that is a bit too much. Having a specific method could make more sense, at least for now.

ektrah: I think the generic solution would probably look like this:

```typescript
let serializer = SERIALIZERS[accept];
if (serializer) {
    let content = serializer.serialize(thing.getThingDescription());
    res.code = "2.05";
    res.setHeader("Content-Format", content.Type);
    res.end(content.Data);
} else {
    res.code = "4.06";
    res.end(`Content-Format ${accept} is not supported by this resource.`);
}
```

JKRhb: The latest commits now implement a generic version using the `ContentSerdes` class for serialization. In the future, this would also allow for different TD representations (e.g., as CBOR or XML), but it seems as if these two codecs are currently not supported/implemented.

JKRhb:
Hmm, does somebody know why the Windows tests are suddenly failing? :/

danielpeintner:
> it's actually the firestore tests Edit2: Or both actually?

we do have issues with firestore since a while (see https://github.com/eclipse/thingweb.node-wot/issues/827)..

danielpeintner (approved):

JKRhb:
@danielpeintner Thank you for approving :) I took the freedom of squashing the commits.

packages/binding-coap/test/coap-server-test.ts
```diff
@@ -297,19 +297,32 @@ class CoapServerTest {
         let responseCounter = 0;
 
         const defaultContentFormat = "application/td+json";
-        const unsupportedContentFormat = "application/cbor";
-        const contentFormats = [defaultContentFormat, "application/json", unsupportedContentFormat];
+        const unsupportedContentFormat = "application/cbor"; // TODO: Update once CBOR is supported
```
relu91, line 300: Wouldn't it be safer to just use a random string to check for unsupportedContent format? we could remove the TODO in this case.

relu91 (changes requested):
Minor point below

packages/binding-coap/test/coap-server-test.ts
JKRhb: You're right, very good point :)

JKRhb: Should be addressed now :)

danielpeintner:
Good to go from my side. I let @relu91 push the merge button.

relu91 (approved):

PR #856, opened 2022-10-24
refactor: turn Content into Class

### LINKED ISSUE #853: Introduce `toBuffer()` method for `Content`?
At the moment, in order to turn the stream inside a `Content` object into a buffer, you need to invoke the helper function `readStreamFully` with the `Content`'s `body` as its argument:

```ts
    const buffer = ProtocolHelpers.readStreamFully(content.body);
````

With this issue, I wanted to a propose a slight API change for `Content` (informed by its [`dart_wot` counterpart](https://github.com/namib-project/dart_wot/blob/main/lib/src/core/content.dart#L13)), making working with it a bit more handy.

In particular, I would propose turning it from an interface into a class and introducing a dedicated `toBuffer()` method. That would simplify the code above as follows, making it a bit more readable:

```ts
    const buffer = content.toBuffer();
````

I already have a branch ready which I could turn into a PR if this proposal seems reasonable to you.

JKRhb:
Resolves #853.

### DIFF
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -25,7 +25,7 @@ import { Subscription } from "rxjs/Subscription";
 // for Security definition
 import * as TD from "@node-wot/td-tools";
 
-import { ProtocolClient, Content, ContentSerdes, ProtocolHelpers, createLoggers } from "@node-wot/core";
+import { ProtocolClient, Content, ContentSerdes, createLoggers } from "@node-wot/core";
 import { CoapForm, CoapOption, CoapMethodName, isSupportedCoapMethod, isValidCoapMethod } from "./coap";
 import CoapServer from "./coap-server";
 import { Readable } from "stream";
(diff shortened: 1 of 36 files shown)


danielpeintner (approved):

relu91 (approved):
Looks good! I've just spotted one trivial thing. I'm seeing that some binding is using the `ContentSerders.DEFAULT` constant, now that we have a class we could create a subclass called DefaultContent to make everything less verbose. 

What do you think?

JKRhb:
> Looks good! I've just spotted one trivial thing. I'm seeing that some binding is using the `ContentSerders.DEFAULT` constant, now that we have a class we could create a subclass called DefaultContent to make everything less verbose.
> 
> What do you think?

Sure, sounds good, added this to the latest version of this PR :)

danielpeintner:
@relu91 I give it to you to merge since you raised some final comments

relu91 (approved):
Thanks!! 🥳

PR #857, opened 2022-10-24
fix: add uriVariables to TD

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/855

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -208,8 +208,8 @@ export default class HttpServer implements ProtocolServer {
 
     private updateInteractionNameWithUriVariablePattern(
         interactionName: string,
-        uriVariables: PropertyElement["uriVariables"],
-        thingVariables: PropertyElement["uriVariables"]
+        uriVariables: PropertyElement["uriVariables"] = {},
+        thingVariables: PropertyElement["uriVariables"]= {}
     ): string {
         if (
             (uriVariables && Object.keys(uriVariables).length > 0) ||
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -208,17 +208,32 @@ export default class HttpServer implements ProtocolServer {
 
     private updateInteractionNameWithUriVariablePattern(
         interactionName: string,
-        uriVariables: PropertyElement["uriVariables"]
+        uriVariables: PropertyElement["uriVariables"],
+        thingVariables: PropertyElement["uriVariables"]
     ): string {
-        if (uriVariables && Object.keys(uriVariables).length > 0) {
+        if (
+            (uriVariables && Object.keys(uriVariables).length > 0) ||
+            (thingVariables && Object.keys(thingVariables).length > 0)
+        ) {
             let pattern = "{?";
             let index = 0;
-            for (const key in uriVariables) {
-                if (index !== 0) {
-                    pattern += ",";
+            if (uriVariables) {
+                for (const key in uriVariables) {
+                    if (index !== 0) {
+                        pattern += ",";
+                    }
+                    pattern += encodeURIComponent(key);
+                    index++;
                 }
-                pattern += encodeURIComponent(key);
-                index++;
+            }
+            if (thingVariables) {
+                for (const key in thingVariables) {
+                    if (index !== 0) {
+                        pattern += ",";
+                    }
+                    pattern += encodeURIComponent(key);
+                    index++;
+                }
             }
             pattern += "}";
             return encodeURIComponent(interactionName) + pattern;
@@ -331,7 +346,8 @@ export default class HttpServer implements ProtocolServer {
             for (const propertyName in thing.properties) {
                 const propertyNamePattern = this.updateInteractionNameWithUriVariablePattern(
                     propertyName,
-                    thing.properties[propertyName].uriVariables
+                    thing.properties[propertyName].uriVariables,
+                    thing.uriVariables
                 );
                 const href = base + "/" + this.PROPERTY_DIR + "/" + propertyNamePattern;
                 const form = new TD.Form(href, type);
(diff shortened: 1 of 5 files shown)


danielpeintner:
[Helpers.parseInteractionOptions()](https://github.com/eclipse/thingweb.node-wot/blob/1fb5caad0d228ca04ff57b910352583ea0ccf432/packages/core/src/helpers.ts#L279-L315) is strict in a sense that if the `ThingInteraction` does not contain information about `uriVariables` it does not add them even though they are provided by the client
`thing.invokeAction("increment", undefined, { uriVariables: { step: 3 } });`

@relu91 others
I wonder whether this is correct or we should add the uriVariables EVEN if the TD does not mention them.. kind of handing them on..

egekorkan:
I was wondering about the source of the issue and now this explanation makes sense.

However:
- I think that we should not use uriVariables as an input in the examples. It is bad practice to do so by the TD spec:
> uriVariables are mainly for properties and events. When retrofitting an existing system, it may be necessary to use uriVariables for actions. In general, it is recommended to avoid uriVariables as much as possible when a new WoT-based system is designed.

Thus, I think that we should not do this PR.

- If such interaction is provided in the consumer script but the TD does not have it, at least a warning should be shown in the cli. I would say that a harder error can be also shown since this behavior is not allowed by the TD spec at https://w3c.github.io/wot-thing-description/#client-data-schema
- A bigger question is whether this input check should be done in general for consumer scripts, i.e. can a Consumer send a string when the Thing expects a number? I think that this should also not be allowed, following the same logic as the one above with the corresponding assertion.

danielpeintner:
> * I think that we should not use uriVariables as an input in the examples. It is bad practice to do so by the TD spec:

We can change it but we used the counter sample to show "what" you can do... we are not saying this is "how" you should do it. Anyhow, if people think we should do it differently it is fine by me also.

danielpeintner:
@egekorkan what if we extend the counter example to provide both... `step` in `uriVariables` with  and `body` and we properly state that the preferred way is to use the `body` version. Does that make sense?

egekorkan:
I think that if we simply want to showcase the usage of uriVariables, we can do so in a property. It can be that the urivariables pass the desired unit of a temperature or in the case of an object, uriVariables can filter to take a single key value pair of the object.

danielpeintner:
Okay, let's do it as follows:
* use uriVariables for properties (we do that already, see http://plugfest.thingweb.io:8083/counter/properties/countAsImage?fill=#FF9E2C).
 Note1: It need to check why the image gets reported as JSON string and not as `image/svg+xml`
 Note2: we need a sample for global uriVariables
*  remove uriVariables for actions
* fix the issue with [Helpers.parseInteractionOptions()](https://github.com/eclipse/thingweb.node-wot/blob/1fb5caad0d228ca04ff57b910352583ea0ccf432/packages/core/src/helpers.ts#L279-L315) being to strict and *NOT* taking into account _global_ uriVariables.
 I now see that the original TD was "OKAY" since we used global uriVariables.

@egekorkan  @relu91 does that make sense?

egekorkan:
I think that the proposals make sense. Also I think one of the main problems is that the counter example is given to the newcomers to node-wot while also being used for experimentation. I think that these should be split and the "newcomer example" should be the simplest thing possible that does really basic functionality. If you also think so, we can create a separate issue thread to decide on what is most important to show to a newcomer.

danielpeintner:
I suggest to fix in **this** PR the initial issue https://github.com/eclipse/thingweb.node-wot/issues/855 that global uriVariables are not passed properly.

I created several follow-up issues:
* create an actual newcomer example (counter gets more and more complicated), see https://github.com/eclipse/thingweb.node-wot/issues/860
* use local *and* global uriVariables for properties in counter (remove uriVariables for actions), see https://github.com/eclipse/thingweb.node-wot/issues/861
* `image/svg+xml` not properly reported for http://plugfest.thingweb.io:8083/counter/properties/countAsImage?fill=#FF9E2C, see https://github.com/eclipse/thingweb.node-wot/issues/862

packages/binding-http/src/http-server.ts
```diff
@@ -208,17 +208,32 @@ export default class HttpServer implements ProtocolServer {
 
     private updateInteractionNameWithUriVariablePattern(
         interactionName: string,
-        uriVariables: PropertyElement["uriVariables"]
+        uriVariables: PropertyElement["uriVariables"],
+        thingVariables: PropertyElement["uriVariables"]
```
relu91, line 212:
```suggestion
        uriVariables: PropertyElement["uriVariables"] = {},
        thingVariables: PropertyElement["uriVariables"]= {}
```

packages/binding-http/src/http-server.ts
```diff
@@ -208,17 +208,32 @@ export default class HttpServer implements ProtocolServer {
...
     ): string {
-        if (uriVariables && Object.keys(uriVariables).length > 0) {
+        if (
+            (uriVariables && Object.keys(uriVariables).length > 0) ||
+            (thingVariables && Object.keys(thingVariables).length > 0)
+        ) {
```
relu91, line 217:
```suggestion
      const variables = Object.assign({},uriVariables, thingVariables)
        if ( Object.keys(variables).length > 0 ) {
```

relu91 (changes requested):
I think you can simplify a little bit the code (to be tested)

packages/binding-http/src/http-server.ts
danielpeintner: I am fine with the proposed changes even though I have to say that it seems a matter of taste to me what "simplify" means in this case 🙈.

Moreover, the proposed change create a new variable `variables` which after the `if` is no longer used.
@relu91 If you want me to make these changes I am fine. If you agree this is more a preference change I would go ahead as is...

relu91: Ah right, I see your point. My main goal was to suggest a shorter and more readable version of the same algorithm (it can be valued objectively by counting lines, code branches, and replicated code). Although complaining about variable creation overhead right now feels like an early optimization problem to me, I leave you to decide 😃 you can merge the PR without addressing my review comments.

danielpeintner: 👍

relu91 (approved):
thank you for applying the changes!

PR #864, opened 2022-11-07
Function-based discovery API alternative

JKRhb:
After today's discussion in the WoT Scripting API call, I decided to provide an alternative PR to #773 that uses a more "functional" approach to discovery (as proposed by @relu91 in https://github.com/w3c/wot-scripting-api/issues/222), modelling each discovery method as its own method of an internal `ThingDiscovery` object.

Since there was some discussion regarding the return type of the `direct` method, I provided two alternatives, one of which returns a `ThingDescription` object while the other one returns an iterable `AsyncGenerator` (which is a slightly different concept to an `AsyncIterator`, more similar to Dart Streams with a more intuitive API in my opinion).

As I mentioned in the call, a generator/iterator-based return type would also make it possible to support discovery from a multicast URL with the `direct` method, simple yielding multiple TDs if the underlying platform should receive multiple responses. However, this might only be relevant for CoAP over UDP, so this might not be the strongest argument for a generator-based return type. Still, I think it might be worth considering.

In general, I think this looks promising, and I am looking forward to your feedback :)

Edit: Thinking about it, the `AsyncGenerator` should rather still be an `AsyncIterator` in order to be able to cancel the discovery process.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es6",
+        "target": "es2018",
         "lib": ["dom"],
         "skipLibCheck": false,
         "module": "commonjs",
--- a/packages/binding-coap/src/coap-client.ts
+++ b/packages/binding-coap/src/coap-client.ts
@@ -62,11 +62,8 @@ export default class CoapClient implements ProtocolClient {
         req.setOption("Accept", "application/td+json");
         return new Promise<Content>((resolve, reject) => {
             req.on("response", (res: IncomingMessage) => {
-                let contentType = res.headers["Content-Format"];
-                if (typeof contentType !== "string") {
-                    contentType = "application/td+json";
-                }
-                resolve({ type: contentType, body: Readable.from(res.payload) });
+                const contentType = (res.headers["Content-Format"] as string) ?? "application/td+json";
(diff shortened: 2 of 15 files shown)


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
     "compilerOptions": {
-        "target": "es6",
+        "target": "es2018",
```
danielpeintner, line 3: _No real review yet.. just since I stumbled over this change_

I am not sure but I think we used to have issues with the browser-bundle when increasing the target version.. I might be wrong though...

JKRhb:
I'll convert this PR to a draft for now as it is still WIP (or might even be closed in the end).

JKRhb:
I tried to align this PR a bit more to the current state of the spec, but it seems to me as if we would need to update the typescript-definitions for that to really work 🤔

JKRhb:
In general, I think it should be possible to replace the `WoTHelpers.fetch()` method with the `requestThingDescription` function as a first step towards a closer aligned node-wot when it comes to discovery.

JKRhb:
Hmm, I think in order to make things a bit cleaner, I will close this PR and replace it with another one that only implements the `requestThingDescription` method. There we we can then have some focused discussion on the implementation details.

PR #865, opened 2022-11-08
chore: disable testing for binding-firestore and binding-mbus

danielpeintner:
see issues described in https://github.com/eclipse/thingweb.node-wot/issues/827

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,15 @@ jobs:
               with:
                   node-version: ${{ matrix.node-version }}
                   cache: "npm"
+
+            - name: Update node-gyp
+              run: |
+                  $WhereNode = Get-Command node | Select-Object -ExpandProperty Definition
+                  $NodeDirPath = Split-Path $WhereNode -Parent
+                  $NodeModulesPath = $NodeDirPath + "\node_modules\npm\node_modules\npm-lifecycle"
+                  cd $NodeModulesPath
+                  npm install node-gyp@8.x
+
             - name: npm 7
               # npm workspaces requires npm v7 or higher
               run: npm i -g npm@7 --registry=https://registry.npmjs.org
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ jobs:
 
         strategy:
             matrix:
-                os: [macos-latest, ubuntu-latest, windows-2019]
+                os: [macos-12, ubuntu-latest, windows-2019]
                 node-version: [14.x, 16.x, 18.x]
 
         steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,8 @@ jobs:
                   cache: "npm"
 
             - name: Update node-gyp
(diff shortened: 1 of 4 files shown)


relu91:
I think it is not the test, but firestore dependencies. You have to exclude the package in the workspaces: https://github.com/eclipse/thingweb.node-wot/blob/master/package.json#L37

danielpeintner:
Now mbus is causing issues ?

`npm ERR! path /Users/runner/work/thingweb.node-wot/thingweb.node-wot/node_modules/node-mbus`
see https://github.com/eclipse/thingweb.node-wot/actions/runs/3420221062/jobs/5694753037#step:5:7

relu91:
seems that both packages are using old native libraries that are causing problems?

danielpeintner:
The _funny_ part is that re-running the tests often helps.. this means to me at least it is somewhat temporal !?

relu91:
I excluded firebase also from the building process

danielpeintner:
Mhh, seems not enough. I disabled/excluded mbus also to see whether that helps...

relu91:
> Mhh, seems not enough. I disabled/excluded mbus also to see whether that helps...

There are some people to complain about node-gyp and windows-latest. Maybe [this](https://github.com/actions/setup-node/issues/280#issuecomment-1139752165) can help... but I would defer it since I don't have the bandwidth to try that fix.

relu91:
it is failing again.... maybe it is the Modbus binding. It depends on serialport too. 😢

danielpeintner:
Shall we use `windows-2019` instead of `windows-latest`

relu91:
> Shall we use windows-2019 instead of windows-latest

It would be better to use windows latest but I'm okay with it for focusing on other features.

relu91:
I guess we have to reopen this one: https://github.com/eclipse/thingweb.node-wot/issues/678

relu91:
this chain of failures starts to be creepy 👻 😆 .

danielpeintner:
Yes, very strange... since now mac-os-latest fails with gyp !?

danielpeintner:
let's try `macos-12` instead of `macos-latest` (maybe it helps to use a different [runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)).
Anyhow, VERY VERY strange !?

relu91:
It seems that all the latest runners have some sort of problem with node-gyp... `macos-10.15` is depecrated accordingly to the documentation. What is weird is that I got the same problem with a node alpine docker image today. Maybe it is something broken in the latest npm 8.x.y version? 🤔 .

danielpeintner:
Mhh, I also hope the issues get solved on their own and hence I suggest to use `non-latest` runners do catch actual issues and open a new issue to track this problem? What do people think?

Note: I will try to bring mbus (and firestore) back to see whether the issues are gone also..

relu91:
> Shall we exclude mbus & firestore for now and use the older runners to get at least clean runs for all the PRs?

Yes I think this should be the priority now... let's fix what we can and have an issue to keep track of this problem

danielpeintner:
Closed in favor of https://github.com/eclipse/thingweb.node-wot/pull/878

Issue #870, opened 2022-11-18
Docker build fails


ctron:
For me, the docker build fails:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'commander@9.4.1',
npm WARN EBADENGINE   required: { node: '^12.20.0 || >=14' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'readable-stream@4.2.0',
npm WARN EBADENGINE   required: { node: '^12.22.0 || ^14.17.0 || >=16.0.0' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'readable-stream@4.2.0',
npm WARN EBADENGINE   required: { node: '^12.22.0 || ^14.17.0 || >=16.0.0' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@opentelemetry/semantic-conventions@1.8.0',
npm WARN EBADENGINE   required: { node: '>=14' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@serialport/bindings-cpp@10.8.0',
npm WARN EBADENGINE   required: { node: '>=12.17.0 <13.0 || >=14.0.0' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@serialport/bindings-interface@1.2.2',
npm WARN EBADENGINE   required: { node: '^12.22 || ^14.13 || >=16' },
npm WARN EBADENGINE   current: { node: 'v12.16.1', npm: '7.24.2' }
npm WARN EBADENGINE }
npm WARN deprecated @npmcli/move-file@2.0.1: This functionality has been moved to @npmcli/fs
[...]

PR #871, opened 2022-11-18
Fix the build by locking versions during the container build

ctron:
This adds the lock file to the build container, to use the expected versions and not let the build process select newer ones.

I believe this is ok, as the container build already selects a fixed build and runtime container tag. So the content of the container should be much more predictable.

This also adds full container names instead of shortnames.

Closes: #870

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,4 +8,3 @@ node_modules/
 coverage/
 .*/
 tmp/
-*-lock*.*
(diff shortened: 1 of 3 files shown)


relu91 (approved):
Seems it was an oversight to put the lock file in `.dockerignore`, LGTM, ready to go 👍🏻

PR #875, opened 2022-11-21
Fix receiving of MQTT commands

ctron:
Receiving of MQTT commands is currently broken. The reason for this is, that the callback is registered in a way that the "this" context is dropped.

Also see:  https://stackoverflow.com/questions/20279484/how-to-access-the-correct-this-inside-a-callback


packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -115,7 +115,9 @@ export default class MqttBrokerServer implements ProtocolServer {
         }
 
         // connect incoming messages to Thing
-        this.broker.on("message", this.handleMessage);
+        this.broker.on("message", (receivedTopic, rawPayload, packet) =>
+            this.handleMessage(receivedTopic, rawPayload, packet)
+        );
```
relu91, line 120:
```suggestion
        this.broker.on("message", this.handleMessage.bind(this));
```

To keep things short I usually use the `bind` function. @danielpeintner what do you prefer?

danielpeintner: Either way is fine by me 👍

danielpeintner:
@ctron may I ask you to merge in master so that we get rid of the CI test issues we had since some days. Thanks!

danielpeintner:
FYI: I merged master in to get a clean run.

@relu91 I let you decide w.r.t. https://github.com/eclipse/thingweb.node-wot/pull/875#pullrequestreview-1188386149

relu91:
I rebased and fixed the commit, merging when CI is green.

PR #878, opened 2022-11-23
chore: have CI for tests across all Os's (manual) while default CI ru…

danielpeintner:
…ns on linux only

alternative for https://github.com/eclipse/thingweb.node-wot/pull/865

### DIFF
--- a/.github/workflows/ci-full.yaml
+++ b/.github/workflows/ci-full.yaml
@@ -1,13 +1,13 @@
-name: Default CI Pipeline
+name: Full CI Pipeline
 
-on: [push, pull_request]
+on: workflow_dispatch
 jobs:
     setup-and-test:
         runs-on: ${{ matrix.os }}
 
         strategy:
             matrix:
-                os: [ubuntu-latest]
+                os: [macos-latest, ubuntu-latest, windows-latest]
                 node-version: [14.x, 16.x, 18.x]
 
         steps:
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "workspaces": [
         "./packages/td-tools",
         "./packages/core",
-        "./packages/binding-!(firestore-browser-bundle)",
+        "./packages/binding-!(firestore-browser-bundle|firestore)",
         "./packages/cli",
         "./packages/examples",
         "./packages/browser-bundle"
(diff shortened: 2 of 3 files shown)


danielpeintner:
@JKRhb Sorry for pinging you again.

```
   1) CoAP server implementation
       should support TD Content-Format negotiation:
     Error: Unknown Content-Format: application/foobar
```

Have you seen such issues on your side?
https://github.com/eclipse/thingweb.node-wot/actions/runs/3530516899/jobs/5922593923#step:7:195

I am seeing it in *this* PR and also in https://github.com/eclipse/thingweb.node-wot/pull/865
https://github.com/eclipse/thingweb.node-wot/actions/runs/3522562908/jobs/5928441681#step:7:195

JKRhb:
Oh, sorry, using that fictitious content type must have broken something :/ I'll look into it!

JKRhb:
Should be fixed in #879 now :)

.github/workflows/ci-full.yaml
```diff
@@ -0,0 +1,78 @@
+name: Full CI Pipeline
+
+on: workflow_dispatch
```
JKRhb, line 3: Maybe we could also let the full CI pipeline only run for changes on the main branch instead? This way it does not slow down progress in PRs but still uncovers OS-specific errors.

danielpeintner: At the moment the *full* test is not run automatically. Hence one needs to trigger it manually. When doing so once can usually select the branch it should run this action.. this is at least how I have been seeing this working.

Do you mean something else ?
Do you mean we should use `on: [push]` only and remove `pull_request` ?

JKRhb: > Do you mean something else ?
> Do you mean we should use on: [push] only and remove pull_request ?

Exactly, that is what I meant :)

danielpeintner: Makes sense!
@relu91 any thoughts?

danielpeintner:
Note: firestore tests fail on ubuntu_latest also.
Hence, I will remove firestore from the test set.

relu91 (approved):
LGTM... just remember to have a tracking issue so that in the future we can restore the full checklist.

PR #879, opened 2022-11-24
fix(binding-coap): fix content negotiation test

JKRhb:
This PR fixes an issue with the current CoAP tests raised by @danielpeintner in https://github.com/eclipse/thingweb.node-wot/pull/878#issuecomment-1326023114. The content negotiation test currently fails due to the fact that `application/foobar` (which was added as an example for an unsupported content type) is not registered in node-coap and therefore throws an error.

In order to resolve this problem, this PR adds a call to the `registerFormat` function and registers `application/foobar` with the reserved numeric value 65000 within the test in question.


danielpeintner (approved):
LGTM

Tests are failing due to firestore (needs to be fixed else)

PR #881, opened 2022-11-25
Fixed package.json dependency versions of exposed-thing example

ArPhil:
Hi!

I am new to node-wot and wanted to quickly setup a demo with two things. I found the exposed-thing template in the examples folder.

While one can use the example when executing `npm install` and `npm build` from the root directory, copying the example folder and executing the same command within the exposed-thing folder fails for two reasons:

1. `"@node-wot/binding-http": "0.8.3"` and `"@node-wot/core": "0.8.3"` are not published on the npm registry yet. Without installing modules and building from the root of the project, npm fails to load these modules from the registry.
2. `"ajv": "^7.0.4"` breaks the code in `index.js` (importing and using ajv) since the way of importing the library changed from version `^6.12.4` (which is locked in the `package-lock.json` in the root).

I suggest chaging the dependency versions as commited in this PR because:

- It does not break installing and building from the root folder
- but also allows copying the exposed-thing demo folder as standalone project and installing + building the project from there.

For me, such a basic structure of a WoT project is very helpfull, because I clearly see the required dependencies, which are not so obvious while having a mono repo with many npm workspaces.

Happy to get some feedback on this! :)

### DIFF
--- a/examples/templates/exposed-thing/package.json
+++ b/examples/templates/exposed-thing/package.json
@@ -16,10 +16,10 @@
         "tslint": "5.12.1"
     },
     "dependencies": {
-        "@node-wot/binding-http": "0.8.3",
-        "@node-wot/core": "0.8.3",
+        "@node-wot/binding-http": "^0.8.2",
+        "@node-wot/core": "^0.8.2",
         "request": "2.88.0",
-        "ajv": "^7.0.4"
+        "ajv": "^6.12.6"
     },
     "optionalDependencies": {},
     "scripts": {
--- a/examples/templates/exposed-thing/package.json
+++ b/examples/templates/exposed-thing/package.json
@@ -16,8 +16,8 @@
         "tslint": "5.12.1"
     },
     "dependencies": {
-        "@node-wot/binding-http": "^0.8.2",
-        "@node-wot/core": "^0.8.2",
+        "@node-wot/binding-http": "0.8.x",
+        "@node-wot/core": "0.8.x",
         "request": "2.88.0",
         "ajv": "^6.12.6"
     },


examples/templates/exposed-thing/package.json
```diff
@@ -16,10 +16,10 @@
         "tslint": "5.12.1"
     },
     "dependencies": {
-        "@node-wot/binding-http": "0.8.3",
-        "@node-wot/core": "0.8.3",
+        "@node-wot/binding-http": "^0.8.2",
+        "@node-wot/core": "^0.8.2",
```
danielpeintner, line 20: This is correct. It might have been done (automatically) when upgrading the dependencies for the next release..

examples/templates/exposed-thing/package.json
```diff
@@ -16,10 +16,10 @@
...
         "request": "2.88.0",
-        "ajv": "^7.0.4"
+        "ajv": "^6.12.6"
```
danielpeintner, line 22: @egekorkan May I ask your input since you were maintaining the templates

examples/templates/exposed-thing/package.json
danielpeintner: @egekorkan I wonder whether we might stick to a *fix* version just to avoid any issues. Any thoughts?

danielpeintner:
@ArPhil thanks for your input/feedback.
I need to point you to [CONTRIBUTING ](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md) which requires signing Eclipse ECA et cetera..

examples/templates/exposed-thing/package.json
egekorkan: So we should definitely fix node-wot versions since node-wot does not follow the semantic versioning pattern. Only upgrades in the latest number should be allowed automatically, i.e. update from 0.8.2 to 0.8.3 should be always fine

examples/templates/exposed-thing/package.json
egekorkan: I think that this is fine. We should however upgrade the actual code that uses ajv at some point.

relu91 (approved):
LGTM, thank you! @ArPhil you might be also interested in https://github.com/UniBO-PRISMLab/wam (it is a little bit outdated but it can generate a nice project stub for you in seconds :) )

danielpeintner:
@ArPhil once signed the Eclipse ECA please let me know so that I can re-run the check. We are not allowed to merge unless this is done.
Note: This needs to be done just once

ArPhil:
@egekorkan , I modified the node module versions to 8.2.x, since this allows the update to minor (semver) versions as you suggested. I added a second signed commit.

@danielpeintner , Sorry for my late reply. I signed the Eclipse ECA and ammended the signature to my first commit.

Hope this is OK the way now. :-)

examples/templates/exposed-thing/package.json
egekorkan: @ArPhil I also like the .x notation, good idea!

ArPhil:
@relu91 , should be fixed now via rebasing (at least the git log looks good to me now). Sorry for the inconvenience...

PR #883, opened 2022-11-30
Upgrade dockerfile

### LINKED ISSUE #869: NodeJS 12 is EOL
NodeJS 12 is EOL, the current LTS version is 18.

I think the container images and build instructions should be updated accordingly.

relu91:
As discussed in #869, Node 12 is in its EOL state. This PR upgrades the Dockerfile with the latest Node LTS version. As a consequence, I also removed `npm@7` install step because it is now bundled with Node. 

fix #869

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/node:12.16.1-alpine3.11 as BUILD
+FROM docker.io/library/node:18-alpine as BUILD
 RUN apk add --no-cache \
 	gcc \
     g++ \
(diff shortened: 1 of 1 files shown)


danielpeintner:
See comments https://github.com/eclipse/thingweb.node-wot/issues/869#issuecomment-1327289377

@relu91 Shall/can we update the readme  in this PR or in yet another PR?

relu91:
> @relu91 Shall/can we update the readme in this PR or in yet another PR?

Let's merge this PR but don't close the issue. We can fix the README problem later.

danielpeintner (approved):

Issue #886, opened 2022-12-01
node-fetch needs Content-Length explicitly for Streams


msrasheed:
## Scenario
I am trying to submit a POST request with an action with the following code snippet:
```js
dataToSend = {passive: true};
thing.invokeAction("login", dataToSend).catch((err) => console.log(err));
```
but the request generated looks like this
```http
POST /api/login HTTP/1.1
X-Api-Key: <some_api_key>
Content-Type: application/json
Accept: */*
User-Agent: node-fetch/1.0 (+https://github.com/bitinn/node-fetch)
Accept-Encoding: gzip,deflate
Connection: close
Host: 10.0.1.64:5000
Transfer-Encoding: chunked

10
{"passive":true}
0


```
The thing description I'm using is attached (as a txt file cause GitHub).
[octoprint_td.txt](https://github.com/eclipse/thingweb.node-wot/files/10128491/octoprint_td.txt)

## Cause
Did some digging and was able to reproduce the issue using `node-fetch`
```js
const fetch = require('node-fetch');
const stream = require('stream');

content = Buffer.from(JSON.stringify({passive: true}));
console.log(content.length);

const requestInit = {
  method: "POST",
  body: stream.Readable.from(content)
};

request = new fetch.Request('http://10.0.1.64:5000/api/login', requestInit);
request.headers.append('X-Api-Key', "<some_api_key>");
request.headers.append('Content-Type', "application/json");

console.log(request.headers);

fetch(request).then(resp => {
  console.log(resp.status);
  console.log(resp.body.read().toString());
});
```
The issue is related to [this line](https://github.com/node-fetch/node-fetch/blob/55a4870ae5f805d8ff9a890ea2c652c9977e048e/README.md?plain=1#L549) in the `node-fetch` README that if the body type is a Stream, then Content-Type won't be automatically determined (as far as I know, all parameters are converted to streams and then passed to bindings).

Adding the following remedies the issue
```js
request.headers.append('Content-Length', 16);
```

danielpeintner:
Thanks for your finding and your PR https://github.com/eclipse/thingweb.node-wot/pull/887.
I added some comments in the PR.

relu91:
Thank you for using node-wot! About the issue, I don't think it is a bug. When using [`chunked` transfer encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding) and the body must look like what you are showing:
```
chunk size
<content>
0
```

[...]

PR #887, opened 2022-12-01
fix(binding-http): node-fetch content-length and streams fix

msrasheed:
In node-fetch, if body is of request is a Stream, [Content-Length is not set automatically](https://github.com/node-fetch/node-fetch/blob/55a4870ae5f805d8ff9a890ea2c652c9977e048e/README.md?plain=1#L549) and results in weird output.

Submitted related issue #886 

Signed-off-by: moiz <mrasheed@purdue.edu>

### DIFF
--- a/packages/binding-http/src/http-client-impl.ts
+++ b/packages/binding-http/src/http-client-impl.ts
@@ -175,7 +175,8 @@ export default class HttpClient implements ProtocolClient {
     }
 
     public async invokeResource(form: HttpForm, content?: Content): Promise<Content> {
-        const headers = content ? [["content-type", content.type]] : [];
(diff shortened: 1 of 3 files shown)


danielpeintner:
First of all thank you for your finding and your contribution!

Some comments:
* We do have failing HTTP Client tests, https://github.com/eclipse/thingweb.node-wot/actions/runs/3588659002/jobs/6044341218#step:7:388
* Do you think it is possible to come up with a test that fails before and passes afterwards w.r.t. this issue?
* Minor: I see some [style check issues](https://github.com/eclipse/thingweb.node-wot/actions/runs/3588659002/jobs/6044341424) as well which need to be resolved. Using the husky workflow we have installed this should not happen.

PR #892, opened 2022-12-06
feat: allow to process submodel only

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/888

### DIFF
--- a/packages/td-tools/test/AssetInterfaceDescriptionTest.ts
+++ b/packages/td-tools/test/AssetInterfaceDescriptionTest.ts
@@ -23,15 +23,43 @@ import { promises as fs } from "fs";
 class AssetInterfaceDescriptionUtilTest {
     private assetInterfaceDescriptionUtil = new AssetInterfaceDescriptionUtil();
 
+    @test async "should correctly transform sample JSON AID_submodel HTTP v03 into a TD"() {
+        const modelFullString = await (await fs.readFile("test/util/AID_v03.json")).toString();
+        const modelFull = JSON.parse(modelFullString);
+        const modelSub = modelFull.submodels[1];
+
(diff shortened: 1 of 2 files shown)


danielpeintner:
@sebastiankb you should be able to test this locally with

`npm install https://github.com/eclipse/thingweb.node-wot.git#pull/892/head`

packages/td-tools/test/AssetInterfaceDescriptionTest.ts
```diff
@@ -125,7 +153,7 @@ class AssetInterfaceDescriptionUtilTest {
             .to.have.lengthOf(1);
 
         // filter Modbus and HTTP and submodel only
-        const td3 = this.assetInterfaceDescriptionUtil.transformToTD(modelAID, `{"title": "myTitle"}`, "Modbus|HTTP");
+        const td3 = this.assetInterfaceDescriptionUtil.transformAAS2TD(modelAID, `{"title": "myTitle"}`, "Modbus|HTTP");
```
sebastiankb, line 156: is it also possible to filter by interface name?

danielpeintner: I think that's what it is doing...
We currently have for example the following term in JSON

`"idShort": "Interface1 (Modbus; s. slides from 21.6.2022)"`

The regex "Modbus|HTTP" look into _this_ `idShort` and tries to find either Modbus or HTTP...

One could also use "Interface1" for the regex to select Modbus.

sebastiankb: ok, perfect. From my view we can merge this PR

PR #894, opened 2022-12-10
feat: respect tm:optional

fatadel:
This PR modifies `getPartialTDs` method, so that it respects `tm:optional` and returns an array where the first element has all affordances, and the rest are constructed by removing one by one the optional ones.

### DIFF
--- a/packages/td-tools/src/thing-model-helpers.ts
+++ b/packages/td-tools/src/thing-model-helpers.ts
@@ -458,6 +458,42 @@ export class ThingModelHelpers {
             data = this.fillPlaceholder(data, options.map);
         }
         tmpThingModels.unshift(data); // put itself as first element
+
+        if (data["tm:optional"]) {
+            // Collect optional affordances
+
+            const optionalAffordances = {
+                properties: [],
+                actions: [],
+                events: [],
+            } as Record<string, string[]>;
+
+            for (const optional of data["tm:optional"]) {
+                const [_, type, name] = optional.split("/");
+                optionalAffordances[type].push(name);
+            }
+
+            delete data["tm:optional"];
+
+            // Add a version of partial td without an optional affordance
+            for (const type in optionalAffordances) {
+                for (const affordance of optionalAffordances[type]) {
+                    const newAffordance = Object.keys(data[type] as DataSchema)
+                        .filter((key) => key !== affordance)
+                        .reduce((obj, key) => {
+                            obj[key] = (data[type] as DataSchema)[key];
+                            return obj;
+                        }, {} as DataSchema);
+
+                    const newData = {
+                        ...data,
+                        [type]: newAffordance,
+                    };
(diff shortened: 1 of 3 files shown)


fatadel:
> No review yet. I just noticed that you introduced some eslint warnings see https://github.com/eclipse/thingweb.node-wot/pull/894/files
> 
> BTW, I wonder why the eslint pipeline check does not mark it as an error

Will fix now, but there is one warning I am not sure about. So, I've assigned an unused variable the name `_`, which is usually the case. However, the warning wants me to change the name of that variable to `Test` 🤷🏻‍♂️ I am not sure if this is a great decision, so please let me know.

fatadel:
There is another thing with this PR. When constructing partial TDs removing the optional affordances one by one, it doesn't consider a combination of the optional affordances. I wanted to add it in the next iteration but thinking now maybe it's better to implement that initially.

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -458,6 +458,42 @@ export class ThingModelHelpers {
...
+            // Collect optional affordances
+
+            const optionalAffordances = {
+                properties: [],
+                actions: [],
+                events: [],
+            } as Record<string, string[]>;
+
+            for (const optional of data["tm:optional"]) {
+                const [_, type, name] = optional.split("/");
```
danielpeintner, line 472: I would either rename the variable as requested or add an ignore statement

danielpeintner:
> This PR modifies `getPartialTDs` method, so that it respects `tm:optional` and returns an array where the first element has all affordances, and the rest are constructed by removing one by one the optional ones.

@relu91 are you okay with the change since the code it is mainly coming from you...

fatadel:
> I'm trying to understand the use case for this feature. The initial idea for `getPartialTDs` was that the method would act as the `new` operator in any well-known OOP language. In other words, its duty was to instantiate a concrete Thing Description that would "implement" the ThingModel passed as a parameter. That's why we intentionally ignored the `optional` parameter because a TD without them would still be a valid "implementation" of that thing model. Furthermore, the semantic of the collection of TDs returned is that they resolve the composition of the thing model passed as argument.
> 
> I think this update violates the original design, nothing wrong with that I'm okay to revise it if needed. But perhaps it would be better to simply add a parameter in `CompositionOptions` that is able to state if the user wants the optional interaction affordances and which one. In practice:
> 
> ```ts
> export type CompositionOptions = {
>     baseUrl?: string;
>     selfComposition?: boolean;
>     map?: Record<string, unknown>;
>     keepOptional: boolean | Record<string,boolean>
> };
> ```
> 
> With that, you are still able to respect the semantic of the returned collection (i.e. only return when there is a composed Thing model), but you can also specify that the returned TD or TDs should contain all or some of the optional properties. What do you think?

Very good point, I totally agree, I just didn't know the initial idea/design of this method.

PR #896, opened 2022-12-14
feat(http-server): implement TD content negotiation

JKRhb:
This PR implements a simple content negotiation mechanism for the HTTP binding and adds two ~an~ additional tests. In the process, I applied some refactoring to the language negotiation code.

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -542,6 +542,35 @@ export default class HttpServer implements ProtocolServer {
         }
     }
 
+    private async handleTdRequest(thing: ExposedThing, req: http.IncomingMessage, res: http.ServerResponse) {
+        const td = thing.getThingDescription();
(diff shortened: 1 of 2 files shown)


danielpeintner (approved):
Thank, LGTM 👍 

The only thing I would have done is to split the test into 2 tests.
1 for success and 1 for fail but I can also live with it as is ;-)

JKRhb:
> Thanks, LGTM 👍
> 
> The only thing I would have done is to split the test into 2 tests. 1 for success and 1 for fail but I can also live with it as is ;-)

Thank you for the feedback, should be adjusted now :)

danielpeintner:
Great, thanks!
Let's give @relu91 some time to look at the PR as well

relu91 (approved):
Awesome, let's merge this (thank you also for refactoring!! 🥳 )

PR #901, opened 2023-01-10
fix: add missing unsubscribe to http subscription

### LINKED ISSUE #900: Unable to `stop()` https Subscription
After subscribing to an event using the https bindings via for example `const sub = consumedThing.subscribeEvent('evt', handler)`, `sub.stop()` has no effect.

I believe this is because in https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-http/src/http-client-impl.ts#L171 the argument for the Subscription constructor should not be `null`.

I'll add a PR fixing this

derwehr:
fixes #900

### DIFF
--- a/packages/binding-http/test/http-client-test.ts
+++ b/packages/binding-http/test/http-client-test.ts
@@ -56,6 +56,7 @@ interface TestVector {
 const port1 = 30001;
 const port2 = 30002;
 const port3 = 30001;
+const port4 = 30003;
 
 class TestHttpServer implements ProtocolServer {
     public readonly scheme: string = "test";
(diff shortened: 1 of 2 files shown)


danielpeintner:
@derwehr thanks for your PR. May I ask you to sign Eclipse ECA, see https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md#legal-requirements

With regards to the PR, do you think you can come up with a test that fails before the change and passes after it?
Thanks for your contribution 👍

derwehr:
Added such a test and signed the ECA now :heavy_check_mark:

packages/binding-http/test/http-client-test.ts
```diff
@@ -456,4 +457,50 @@ class HttpClientTest2 {
...
+                    eventSpy();
+                    if (counter === 2) {
+                        sub.unsubscribe();
+                        // wait 100 ms, so that tests fail if unsubscribe didn't work
+                        await new Promise((resolve) => setTimeout(resolve, 100));
+                        // eslint-disable-next-line no-unused-expressions
+                        eventSpy.should.have.been.called.twice;
+                        server.close();
+                        done();
+                    }
```
danielpeintner, line 495: I think we should add a branch to stop this test so that it doesn't hang indefinitely if unsubscribing doesn't work..  maybe something like this

```

} else if (counter > 2) {
   server.close();
   done(new Error("unsubscribe didn't work as expected"));
}
```

danielpeintner (changes requested):
Thanks! The test looks good. I may suggest to stop the test in case of failure (see below).
Anyhow, I can confirm that the test passes with the fix and fails without it.

Can you also fix copyright years to 2023 in both files.
This would also be done by the script we run from time to time but I think it is better to do this as part of this PR.

packages/binding-http/test/http-client-test.ts
derwehr: Done

danielpeintner: Great, thanks!

danielpeintner (approved):
Looks good to me!

danielpeintner:
@derwehr can you fix the coding style issue
see https://github.com/eclipse/thingweb.node-wot/actions/runs/3902813393/jobs/6666397234

danielpeintner (approved):
LGTM

danielpeintner:
@relu91 others?
Is there anything more before we merge this PR?

relu91 (approved):
Sorry, I took some time to review the code paths and check if it works with our latest scripting API specs. Quite complex to be sure at 100% but it fixes user issues and he provides a good test. We can merge it!

PR #906, opened 2023-01-16
fix(binding-http): fix content negotiation

JKRhb:
This PR should resolve https://github.com/eclipse/thingweb.node-wot/issues/905. I already tested the examples in the browser and they are working again :) I will also another test case so that this is also asserted by the CI.

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -544,16 +544,27 @@ export default class HttpServer implements ProtocolServer {
 
     private async handleTdRequest(thing: ExposedThing, req: http.IncomingMessage, res: http.ServerResponse) {
         const td = thing.getThingDescription();
-        let accept = req.headers.accept;
         const contentSerdes = ContentSerdes.get();
 
-        // TODO: Better handling of wildcard values
-        if (accept === "*/*") {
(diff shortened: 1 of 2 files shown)


relu91 (approved):

packages/binding-http/src/http-server.ts
```diff
@@ -544,16 +544,27 @@ export default class HttpServer implements ProtocolServer {
 
     private async handleTdRequest(thing: ExposedThing, req: http.IncomingMessage, res: http.ServerResponse) {
         const td = thing.getThingDescription();
-        let accept = req.headers.accept;
         const contentSerdes = ContentSerdes.get();
 
+        // TODO: Parameters need to considered here as well
```
danielpeintner, line 549: Nit: need to **be** considered

danielpeintner (approved):
LGTM

One minor issue in a comment, which we could fix before merging...

packages/binding-http/src/http-server.ts
JKRhb: Thank you :) Should be fixed now.

danielpeintner: 👍

PR #907, opened 2023-01-16
fix: take into account content type of ExposeThingInit

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/902
fixes https://github.com/eclipse/thingweb.node-wot/issues/862


relu91 (approved):
LGTM

PR #908, opened 2023-01-16
Move endianness from binding-modbus to core

### LINKED ISSUE #903: Inverted properties for  `readAllOperation` in modbus client
## TL;DR;
Modbus client has a bug when you try to read multiple properties in parallel. Some of these values are swamped and the end result is wrong. Below you can find a longer explanation, I'm assigning this to myself therefore I tried to be precise so that I can spare another additional 2 hours of debugging 🤣 . 

## Issue
The Modbus client has a complex algorithm to compact multiple operations in one single transaction cleverly.  Check the documentation to [read more](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-modbus#implementation-notes). 

### My setup
I have a Modbus device with 3 contiguous input registers encode in a Float32 Little Endian format. In practice:
| Address  | 16-value  |
|---|---|
|  0 |  Most significative 16 of the first value |
|  1 |  Least significative 16 of the first value|
|  2 |  Most significative 16 of the second value |
|  3 |  Least significative 16 of the second value|
|  4 |  .. |
|  .. | ...   |

### The issue
When using the `readAll`, thanks to our non-blocking strategy the Modbus client understands that we are trying to get 3 continuous values and optimize generating one single request. The client correctly returns the buffer but then our binding implementations manipulates the result to obtain a regular array with **big-endian** values in the [execute()](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-modbus/src/modbus-connection.ts#L91) function. However, the logic implemented there does not work for this corner case because it just reverses the all returned buffer:
```ts
 if (
                    this.endianness === ModbusEndianness.LITTLE_ENDIAN_BYTE_SWAP ||
                    this.endianness === ModbusEndianness.LITTLE_ENDIAN
                )
                    result.buffer.reverse();
```

The correct way to translate this little endian 32 Float values to big endian format would be using `result.buffer.swap32()` but I would rather delegate the whole transformation to the codec. 

### Planned fix
I think that payload manipulation should be delegated to codecs as much as possible and protocol-bindings should just hand over the requests/responses. Therefore, I am keen to remove the logic above a bring it to the `octect-codec` and fix it. This implies also fixing the `contentToValue` function that it seems ignores the `byteSeq` parameter

relu91:
This PR mainly fixes #903, but it adds a bunch of other changes aimed to improve raw encoding and decoding. Since there are some refactor commits I suggest to start reviewing from start to end using this agenda:
- ecd0107defbc2e2fc8337cbaf78b3f645ad4abb5: is the main change. Here I remove the logic of handling endianess in the modbus client and moved it to the core codecs. 
- 5f67eb4304c4a66b06b8f1df8244f666a8137231: while debugging I found out that Buffer.toString() does not print the nice output you can see from `console.log(buff) // prints <Buf 0x2e 0xc5>`. Now the buffer is printed correctly plus some other information about the parameters used while decoding the value. 
- fc72c1ed2b3b0cec79a2d921e18c2af2a0bfd27e while I was at it I refactored the two main parsing functions of the codec
- eb4275d5b561d26cd60420b78c5377b3e73b1f9a: I found out that some of the problems were caused by the fact the Modbus client was not reporting back the right contentType, I fixed it here.

### DIFF
--- a/packages/core/src/codecs/octetstream-codec.ts
+++ b/packages/core/src/codecs/octetstream-codec.ts
@@ -17,6 +17,7 @@ import { ContentCodec } from "../content-serdes";
 import { DataSchema, DataSchemaValue } from "wot-typescript-definitions";
 import { getFloat16, setFloat16 } from "@petamoriken/float16";
 import { createLoggers } from "../logger";
+import { Endianness } from "../protocol-interfaces";
 
 const { debug, warn } = createLoggers("core", "octetstream-codec");
 
@@ -57,7 +58,7 @@ export default class OctetstreamCodec implements ContentCodec {
     ): DataSchemaValue {
         debug(`OctetstreamCodec parsing '${bytes.toString()}'`);
 
-        const bigendian = parameters.byteorder !== "littleendian"; // default to big endian
+        const bigendian = !parameters.byteSeq?.includes(Endianness.LITTLE_ENDIAN); // default to big endian
         let signed = parameters.signed !== "false"; // default to signed
 
         // check length if specified
@@ -78,6 +79,11 @@ export default class OctetstreamCodec implements ContentCodec {
                 dataType = typeSem[2];
                 dataLength = +typeSem[3] / 8 ?? bytes.length;
             }
+        }
+
+        // Handle byte swapping
+        if (parameters.byteSeq?.includes("BYTE_SWAP") && dataLength > 1) {
+            bytes.swap16();
         }
 
         // determine return type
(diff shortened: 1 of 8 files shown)


danielpeintner:
I scanned the changes and I did not notice an obvious issue.

Having said that, I might not be the right person to review and test this PR? Who could be a good reviewer? @sebastiankb , @mkovatsc ?

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -203,116 +216,137 @@ export default class OctetstreamCodec implements ContentCodec {
...
+        const length = options.dataLength ?? 4;
+        const { bigendian, signed, byteSeq } = options;
+
+        if (typeof value !== "number") {
+            throw new Error("Value is not a number");
+        }
+
+        // warn about numbers being too big to be represented as safe integers
+        if (!Number.isSafeInteger(value)) {
+            warn("Value is not a safe integer");
```
danielpeintner, line 257: I think it makes sense to add the actual `value` in the log statement

danielpeintner (approved):
I looked through the PR and did not find any obvious issue (I added one nit comment).

I still hope @sebastiankb and/or @mkovatsc can find the time to review as well.
Otherwise we can ahead with this PR as is and update if we encounter issues.

danielpeintner:
@relu91 besides the minor debug statement I think we should go on with this PR. What do you think?

relu91:
@danielpeintner should be ready to go!

Issue #904, opened 2023-01-13
Move client cache to `Servient` level (?)


relu91:
For some serial protocols (like Modbus), having multiple clients connected to the same endpoint (address+port) creates faulty responses. For other protocols, does it make sense to have such a global cache? Does having them cached per Thing makes sense? 

We could keep the current behavior but we definitely need a configuration variable or something else to indicate that sometimes we want clients to be cached globally. Let's discuss it in this issue

egekorkan:
It may make sense for MQTT as well (there it would apply to the Thing too). It might be interesting in some deployments where multiple Consumer scripts run on a single servient. These scripts would actually get a single published message and be distributed via node-wot, reducing the bandwidth requirements and thus costs (a managed cloud case) in general.

Only problem I can see in protocols (Modbus with baud rate in RTU) that allow different kind of connections. Maybe some clients want to connect with different parameters. In mqtt.js there are some options: https://github.com/mqttjs/MQTT.js#client that would show this problem. However, I am not sure if this actually relevant for us. Plus, even if there is, the developer is free to create multiple servients in this case.

danielpeintner:
It seems that at least on binding level we might need such a configuration.

Maybe, it makes even sense that not only the binding makes a decision but also the user can overrule this decision, right?

relu91:
> Maybe, it makes even sense that not only the binding makes a decision but also the user can overrule this decision, right?

Yeah, maybe some custom property on the form level like `node-wot:caching : ["global", "thing"]` .

>Only problem I can see in protocols (Modbus with baud rate in RTU) that allow different kind of connections. Maybe some clients want to connect with different parameters. In mqtt.js there are some options: https://github.com/mqttjs/MQTT.js#client that would show this problem. However, I am not sure if this actually relevant for us. Plus, even if there is, the developer is free to create multiple servients in this case.

I don't think this is going to be problematic. Remember that we are not caching the connection between the two things but rather the client. Connection pooling is still something that the protocol binding is handling (at least in the current architecture), one example is the Modbus client or the HTTP client for SSE.

JKRhb:
> For some serial protocols (like Modbus), having multiple clients connected to the same endpoint (address+port) creates faulty responses. For other protocols, does it make sense to have such a global cache? Does having them cached per Thing makes sense?
> 

For CoAP, it would definitely make sense as well, e.g., to be able to cache DTLS connections.

relu91:
[...]

PR #909, opened 2023-01-16
Use singleton pattern in the Modbus client factory

relu91:
As discussed in #904, with the Modbus client caching should happen per `Servient` as multiple connections to the same ModbusTCP network will cause collisions. For this reason in the issue, I discussed the possibility to cache clients in a global manner, but when I was starting to implement I figured out that this feature is already there! we have a factory. Therefore, for Modbus, I simply implement a singleton and should work for most of the use cases. With this change, one Servient will have only one instance of the `ModbusClient`, that will manage the dispatch of requests to multiple hosts. It also manages connections and, consequently, can serialize multiple requests to one single host.

### DIFF
--- a/packages/binding-modbus/src/modbus-client-factory.ts
+++ b/packages/binding-modbus/src/modbus-client-factory.ts
@@ -19,12 +19,26 @@ const debug = createDebugLogger("binding-modbus", "modbus-client-factory");
 
 export default class ModbusClientFactory implements ProtocolClientFactory {
     public readonly scheme: string = "modbus+tcp";
+    private singleton: ModbusClient;
 
     public getClient(): ProtocolClient {
-        debug(`Creating client for '${this.scheme}'`);
-        return new ModbusClient();
+        debug(`Get client for '${this.scheme}'`);
+        if (!this.singleton) {
+            this.init();
+        }
+        return this.singleton;
     }
 
-    public init = (): boolean => true;
-    public destroy = (): boolean => true;
+    public init(): boolean {
+        debug(`Initializing client for '${this.scheme}'`);
+        this.singleton = new ModbusClient();
+        return true;
+    }
+
+    public destroy(): boolean {
+        debug(`Destroying client for '${this.scheme}'`);
+        this.singleton.stop();
+        this.singleton = undefined;
+        return true;
+    }
 }
--- a/packages/binding-modbus/src/modbus-client-factory.ts
+++ b/packages/binding-modbus/src/modbus-client-factory.ts
@@ -12,10 +12,11 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
-import { ProtocolClientFactory, ProtocolClient, createDebugLogger } from "@node-wot/core";
+import { ProtocolClientFactory, ProtocolClient, createDebugLogger, createWarnLogger } from "@node-wot/core";
 import ModbusClient from "./modbus-client";
 
 const debug = createDebugLogger("binding-modbus", "modbus-client-factory");
+const warn = createWarnLogger("binding-modbus", "modbus-client-factory");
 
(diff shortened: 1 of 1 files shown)


packages/binding-modbus/src/modbus-client-factory.ts
```diff
@@ -19,12 +19,26 @@ const debug = createDebugLogger("binding-modbus", "modbus-client-factory");
...
+        debug(`Get client for '${this.scheme}'`);
+        if (!this.singleton) {
+            this.init();
+        }
+        return this.singleton;
     }
 
-    public init = (): boolean => true;
-    public destroy = (): boolean => true;
+    public init(): boolean {
```
danielpeintner, line 32: Does this method need to be `public`?

I assume it is `public` to allow recreating the singleton instance. IF so would choose a better name...

danielpeintner (commented):
I wonder about `init()`.

see comment below.

packages/binding-modbus/src/modbus-client-factory.ts
danielpeintner: SHOULD it be possible to recreate the singleton from the outside?

relu91: I think is coming from [ClientFactoryInterface](https://github.com/eclipse/thingweb.node-wot/blob/3acb1aab2d0c1a277896c8a871a2a92910ee2554/packages/core/src/protocol-interfaces.ts#L79) and how its lifecycle is treated in the core. I don't think I can make it private/protected.

danielpeintner: I see and I agree.

But the current implementation, whenever `init()` is called, it re-creates the singleton.
I wonder whether it makes sense to add a check like the following since it can be triggered/called from the outside

```
public init(): boolean {
	if (!this.singleton) {
		debug(`Initializing client for '${this.scheme}'`);
        	this.singleton = new ModbusClient();
	}
        return true; // not sure if it should report false in case the singleton was already created.. I don't think so
}
```

Having said that, I don't know how likely (or if all) it is possible that init() is called multiple times.

Anyhow, "singleton" to me means that the class itself needs to take care that there is just one instance, which cannot change over time....
Do you see my point?

packages/binding-modbus/src/modbus-client-factory.ts
```diff
@@ -19,12 +19,26 @@ const debug = createDebugLogger("binding-modbus", "modbus-client-factory");
 
 export default class ModbusClientFactory implements ProtocolClientFactory {
     public readonly scheme: string = "modbus+tcp";
+    private singleton: ModbusClient;
 
     public getClient(): ProtocolClient {
-        debug(`Creating client for '${this.scheme}'`);
-        return new ModbusClient();
+        debug(`Get client for '${this.scheme}'`);
+        if (!this.singleton) {
```
danielpeintner, line 26: With my proposal this if check should move into `init()`;

packages/binding-modbus/src/modbus-client-factory.ts
```diff
@@ -19,12 +19,26 @@ const debug = createDebugLogger("binding-modbus", "modbus-client-factory");
...
-    public destroy = (): boolean => true;
+    public init(): boolean {
+        debug(`Initializing client for '${this.scheme}'`);
+        this.singleton = new ModbusClient();
+        return true;
+    }
+
+    public destroy(): boolean {
+        debug(`Destroying client for '${this.scheme}'`);
+        this.singleton.stop();
```
danielpeintner, line 40: I think it also makes sense to add a check whether this.singleton is actually defined.. otherwise calling stop() will lead to issues...

packages/binding-modbus/src/modbus-client-factory.ts
relu91: I see, good point! pushing fix soon

danielpeintner (approved):

PR #922, opened 2023-01-31
fix http client and server property "subscription"

lifaon74:
Hi, currently the `thing.observeProperty` was not working on the browser and server for the http binding.

- the HEAD header was not supported so I add it
- calling `.stop()` on the returned Subscription was not working. It came from the fact that the Client was not recycled and was generated each time by the Servient. => I'm not totally sure of this fix as it touches pretty abstract code.

Reviews are welcome

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -756,6 +756,9 @@ export default class HttpServer implements ProtocolServer {
                                     res.writeHead(500);
                                     res.end(err.message);
                                 }
+                            } else if (req.method === "HEAD") {
+                                res.writeHead(202);
+                                res.end();
                             } else {
                                 respondUnallowedMethod(res, "GET");
                             }
@@ -854,6 +857,11 @@ export default class HttpServer implements ProtocolServer {
                                         respondUnallowedMethod(res, "GET, PUT");
                                     }
                                     // resource found and response sent
+                                    return;
+                                } else if (req.method === "HEAD") {
+                                    // HEAD support for long polling subscription
+                                    res.writeHead(202);
+                                    res.end();
                                     return;
                                 } else {
                                     respondUnallowedMethod(res, "GET, PUT");
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -110,7 +110,11 @@ class ConsumedThingEvent extends TD.ThingEvent {
 abstract class InternalSubscription implements Subscription {
     active: boolean;
 
-    constructor(protected readonly thing: ConsumedThing, protected readonly name: string) {
+    constructor(
+        protected readonly thing: ConsumedThing,
+        protected readonly name: string,
+        protected readonly client: ProtocolClient
+    ) {
         this.active = true;
     }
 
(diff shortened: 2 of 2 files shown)


relu91:
Hi! first of all, thank you really much for taking the time to give your contribution. Just some minor things before we can move on:
1. Since this is an eclipse project new contributors have to sign the Eclipse Contributor Agreement ( see [how to contribute](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md#legal-requirements) and let us know if there are any problems). 
2. Even after your last commit there are still some linting errors, can you fix them too? 

Regarding the first issue, I think the changes look good but I would need some time to verify the second point.

lifaon74:
Done.

danielpeintner (commented):
Looking at the code I do not see any obvious issue. 👍 
However I would like to wait for @relu91 to take a closer look.

Two questions:
* You mentioned the issue was in browser only but do you think a test could be created in [binding-http](https://github.com/eclipse/thingweb.node-wot/tree/master/packages/binding-http/test) that fails before this change?
* Is the change for `package-lock.json` actually needed. I do not really understand why it is there. I think there is no need.

Thanks!

lifaon74:
1) before my change, the `thing.observeProperty` was returning a promise which never resolved. Indeed, the on the LongPollingSubscription the 'resolve' was only called when the 'HEAD' request success (which never appends).
Moreover, the calling `stop` on the returned Subscription was not working neither, because a new HttpClient was created each time. Thus, no "internal" Subscriptions were existing, and the cancel was never done.

2) Probably not. However, I suggest you to add a `.nvmrc` file with a specific version of node (ex: `v16`), this way, running `nvm use` followed by `npm i` for new incomers will generate the same package-lock instead of an updated one. As, I was not aware of which node version to use for the project, I used `16.19.0` LTS

danielpeintner:
> 2\. Probably not. However, I suggest you to add a `.nvmrc` file with a specific version of node (ex: `v16`), this way, running `nvm use` followed by `npm i` for new incomers will generate the same package-lock instead of an updated one. As, I was not aware of which node version to use for the project, I used `16.19.0` LTS

I see your points. Anyhow, we state in our [README](https://github.com/eclipse/thingweb.node-wot#as-a-standalone-application) that people should call `npm ci` (instead of `npm i`)

Note: I believe even with a version set in `.nvmrc` you may end up creating slight changes in `package-lock.json` since packages might have been updated etc.

Can you please remove the changes in `package-lock.json`. Thanks!

lifaon74:
Updated

relu91 (changes requested):
First of all, thank you for reporting and providing a solution for the issue. However, your solution for point number 2, has the side effect of caching clients for every operation. We lately discussed it in https://github.com/eclipse/thingweb.node-wot/issues/904 and we decided that is something that the binding level should decide. 

Even though the current core-protocol interface is far to be perfect a more conservative approach to solve the issue is to store the client in `InternalSubscription` class inside `ConsumedThing`. May I ask you to propose such a change?

lifaon74:
Hi, sadly I'm not confident enough to make such an update. The code is pretty opaque to me, and probably result of design choices that I'm not aware.

relu91:
hey! no problem at all. We'll update the PR accordingly and merge it asap. thank u again.

relu91:
Did a quick fix, @lifaon74 can you test it?

relu91 (approved):

danielpeintner:
It looks good to me. I hope @lifaon74 can provide feedback whether it solves the issues he encountered.

danielpeintner (approved):

Issue #831, opened 2022-09-16
Slow tests


relu91:
After tinkering with #796 I found out that the most time-consuming activity in the CI is our tests. Usually, they take minutes even if sometimes in ubuntu they might be faster. We should pinpoint if there are any sleep or timeouts that we can remove and in general seek to improve the performance of our solution. 

As this issue is a little vague, I would add a concrete todo action:
- Verify the presence of timeouts and remove those wherever possible.

relu91:
To ease the improvement we could also use: https://github.com/sinonjs/fake-timers

danielpeintner:
From my side we can close the issue.
@relu91 ?

relu91:
Checked one of the latest runs. Tests now complete in around 1 min and the delays are caused by long install time (see https://github.com/eclipse-thingweb/node-wot/actions/runs/8880185289/job/24379667144 - 4m to install). I think we reached a reasonable speed-up of our tests and we can close this issue now.

PR #923, opened 2023-02-01
Improve test timing

relu91:
I did some improvements in our test suites hoping to reduce wasted time while developing with node-wot. With the reference to #831, the situation is improved but it is not perfect. Some caveats:
- fake timers do not work with sockets ( I went deep down inside node.js internals and I figured out that they are not using the usual `setTimeout` or equivalent method but rather an internal version. Mocking the `Socket.setTimeout` method does not work either -> I gave up). This still causes some slowness in the Modbus protocol binding.
- mbus still times out at 2sec even if you set another timeout
- OPC-UA is still quite slow. 

In general I think we should save about 5 seconds or more each run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12709,6 +12709,7 @@
                 "wot-thing-description-types": "^1.1.0-13-October-2022"
             },
             "devDependencies": {
+                "@sinonjs/fake-timers": "^10.0.2",
                 "@testdeck/mocha": "^0.1.2",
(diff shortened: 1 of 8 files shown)


danielpeintner (commented):
Even small improvements are good improvements 👍 

I noticed that there are 2 check warnings about
* sleep no longer used
* counter assigned but value never used

I think we should fix them before merging.

danielpeintner (approved):

PR #927, opened 2023-02-13
Pull Request for asset-interface-description.ts file

Kaz040:
Hello, to start with, I am completely  new to JS and TS and this is my first time make a PR.  

When I was trying to test Modubs implementation of AID in node-wot, I bumped into a bug, the AID <-> TD code did not consider **qualifier** of each datapoints in AID. we actually include the valueType there which corresponds to the TD dataschema **type**.

So, I added a little line of codes in asset-interface-description.ts file to include the datatype modelled in AID for TD properties.  i did for only properties because the final for how to model the data type has not been made but it is important for modbus testing due to its payload type(octet-stream). 

Thank you and Regards.

### DIFF
--- a/packages/td-tools/src/util/asset-interface-description.ts
+++ b/packages/td-tools/src/util/asset-interface-description.ts
@@ -365,6 +365,17 @@ export class AssetInterfaceDescriptionUtil {
                 thing.properties[key].forms = [];
 
                 for (const vi of value) {
+                    //The first if condition is expected to be temporary. will be adjusted or removed when a decision on how the datapoint's datatype would be modelled is made for AID
+                    if(vi.interaction.constraints.length > 0 && vi.interaction.constraints.length > 0){
+                        if(vi.interaction.constraints[0].value === "float"){
+                            thing.properties[key].type = "number";
+                        }
+                        else{
+                            thing.properties[key].type = vi.interaction.constraints[0].value;
+                        }
+
+                    }
+
                     if (vi.endpointMetadata) {
                         vi.secNamesForEndpoint = secNamesForEndpointMetadata.get(vi.endpointMetadata);
                     }
--- a/packages/td-tools/src/util/asset-interface-description.ts
+++ b/packages/td-tools/src/util/asset-interface-description.ts
@@ -359,21 +359,23 @@ export class AssetInterfaceDescriptionUtil {
             for (const entry of smInformation.properties.entries()) {
                 const key = entry[0];
                 const value: AASInteraction[] = entry[1];
-                logInfo("Property" + key + " = " + value);
+                logInfo("vProperty" + key + " = " + value);
 
                 thing.properties[key] = {};
                 thing.properties[key].forms = [];
 
                 for (const vi of value) {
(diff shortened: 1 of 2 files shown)


danielpeintner:
Hi @Kaz040. Thanks for your contribution 👍.

Some points.
* You need to sign Eclipse ECA as described in [CONTRIBUTING.md](https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md) before we can merge/accept any PR.
* Can you provide a test-case in [AssetInterfaceDescriptionTest.ts](https://github.com/eclipse/thingweb.node-wot/blob/master/packages/td-tools/test/AssetInterfaceDescriptionTest.ts) so that we can better understand what currently does not work as expected. Thanks!

packages/td-tools/test/AssetInterfaceDescriptionTest.ts
```diff
@@ -184,6 +184,12 @@ class AssetInterfaceDescriptionUtilTest {
         expect(tdObj.securityDefinitions[tdObj.security[0]]).to.have.property("scheme").that.equals("nosec");
 
         expect(tdObj).to.have.property("properties").to.have.property("count");
+        // just inserter by Kaz
```
danielpeintner, line 187: I guess a description what it checks is better than a name...

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -359,12 +359,25 @@ export class AssetInterfaceDescriptionUtil {
             for (const entry of smInformation.properties.entries()) {
                 const key = entry[0];
                 const value: AASInteraction[] = entry[1];
-                logInfo("Property" + key + " = " + value);
+                logInfo("vProperty" + key + " = " + value);
```
danielpeintner, line 362: I believe this change is unintended

danielpeintner (changes requested):
I think we are getting close 👍.

Unfortunately, the Eclipse ECA check still fails... maybe because of the commit history. *All* commits need to be done with the email address registered at Eclipse. If you are failing to resolve, a complete *new* PR with the right commit history might be another solution.

BTW, I added 2 more minor comments.

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -359,12 +359,25 @@ export class AssetInterfaceDescriptionUtil {
...
 
                 thing.properties[key] = {};
                 thing.properties[key].forms = [];
 
                 for (const vi of value) {
+                    // The first if condition is expected to be temporary. will be adjusted or removed when a decision on how the datapoint's datatype would be modelled is made for AID
+
+                    if (vi.interaction.constraints instanceof Array && vi.interaction.constraints) {
```
danielpeintner, line 370: I believe the order of the checks should change, first null/undefined check and then whether it is array.

Kaz040:
Hello Daniel.

Thank you for your input and comments. I will make necessary adjustments and get back.

danielpeintner:
Note: replaced by https://github.com/eclipse/thingweb.node-wot/pull/929

PR #928, opened 2023-02-16
fix: catch errors when linked tm is unavailable

fatadel:


fatadel:
> Can you include the failed URL in the error? might be useful to understands which link in broken, when you compose multiple TMs

Done ✅

relu91 (approved):

danielpeintner:
Note: I am not sure why the CI runs take that long...

PR #939, opened 2023-03-18
feat(binding-coap)!: use blockSize instead of blockSZX

JKRhb:
This PR incorporates the changes made in https://github.com/w3c/wot-binding-templates/pull/259 into the CoAP binding implementation.


JKRhb:
Hmmm. Seems like the build failure is caused by `@serialport/bindings`.

relu91 (approved):

PR #943, opened 2023-03-27
Fix subscribe unsubscribe and then subscribe again MQTT issue

### LINKED ISSUE #942: Unable to subscribe-unsubscribe and then resubscribing in binding-mqtt
Hi,

it's not possible to subscribe to a event and then unsubscribe from this event followed by resubscribing later in the binding-mqtt package.

Example for this looks like this:
```javascript
const subscription = await thing.subscribeEvent("test-event", async (interactionOutput) => { ... });
await subscription.stop();
const subscription = await thing.subscribeEvent("test-event", async (interactionOutput) => { ... });
// no events will be received
```

After some debugging in the binding-mqtt package the problem is:
https://github.com/eclipse/thingweb.node-wot/blob/85a1bb7a7003695d1d4234adc81e9cfce2e99c40/packages/binding-mqtt/src/mqtt-client.ts#L62-L63

The subscription is only done if the client emits the `connect` event. Thus, if the client is already connected and we want to subscribe the `this.client.subscribe(topic);` method is not executed.

A possible solution could be: Check if the client is connected (no `connect` event will be fired) and subscribe to the topic.
I don't know if it would be necessary to check if there is already an active subscription
```javascript
if (this.client.connected) {
    this.client.subscribe(topic);
    resolve(new Subscription(() => {
        this.client.unsubscribe(topic);
    }));
}
```

Best regards,
Matthias

relu91:
Fixes #942 (more explanations about the adopted solution on the commit messages).

### DIFF
--- a/packages/binding-mqtt/test/mqtt-client-subscribe-test.unit.ts
+++ b/packages/binding-mqtt/test/mqtt-client-subscribe-test.unit.ts
@@ -75,8 +75,12 @@ describe("MQTT client implementation", () => {
                         done(err);
                     }
                 })
-                .then(() => mqttClient.invokeResource(form, new Content("", Readable.from(Buffer.from("test")))))
-                .then(() => mqttClient.stop())
+                .then(async (sub) => {
+                    await mqttClient.invokeResource(form, new Content("", Readable.from(Buffer.from("test"))));
+                    // Need to manually unsubscribe because stopping the client will unsubscribe all subscriptions
+                    sub.unsubscribe();
+                    await mqttClient.stop();
+                })
                 .catch((err) => done(err));
         }).timeout(10000);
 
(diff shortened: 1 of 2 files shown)


relu91:
@ilovemilk , I added a small test but it would be nice to get also your feedback. Thank you!

packages/binding-mqtt/test/mqtt-client-subscribe-test.unit.ts
```diff
@@ -75,8 +75,43 @@ describe("MQTT client implementation", () => {
                         done(err);
                     }
                 })
-                .then(() => mqttClient.invokeResource(form, new Content("", Readable.from(Buffer.from("test")))))
-                .then(() => mqttClient.stop())
+                .then(async (sub) => {
+                    await mqttClient.invokeResource(form, new Content("", Readable.from(Buffer.from("test"))));
+                    // Need to manually unsubscribe because stopping the client will unsubscribe all subscriptions
```
relu91, line 80:
```suggestion
                    // Need to manually unsubscribe because stopping the client will not unsubscribe all subscriptions
```

Forgot to put a **not**

danielpeintner (approved):
LGTM.
The updated comment should be commited...

PR #952, opened 2023-04-13
update TS definitions

danielpeintner:
"wot-typescript-definitions": "0.8.0-SNAPSHOT.25"
"wot-thing-description-types": "1.1.0-23-March-2023"

### DIFF
--- a/examples/templates/exposed-thing/package.json
+++ b/examples/templates/exposed-thing/package.json
@@ -11,7 +11,7 @@
     "license": "MIT",
     "devDependencies": {
         "typescript": "4.4.3",
-        "wot-typescript-definitions": "0.8.0-SNAPSHOT.23",
+        "wot-typescript-definitions": "0.8.0-SNAPSHOT.25",
         "@types/node": "16.4.13",
         "tslint": "5.12.1"
     },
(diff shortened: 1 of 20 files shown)


danielpeintner:
I had to remove event handler code (which the API does not provide anymore)
@relu91 please check

BTW, since "wot-typescript-definitions" now depends on "wot-thing-description-types" shall we clean-up package.json dependencies by removing the latter where possible?

relu91 (approved):
LGTM.

relu91:
> BTW, since "wot-typescript-definitions" now depends on "wot-thing-description-types" shall we clean-up package.json dependencies by removing the latter where possible?

I don't have a clear answer. My experience is that sometimes VSCode does not pick indirect dependencies (a library that is a dependency of another) and it gets quite annoying. I am not 100% sure if it is (or was) a bug of VScode language server or is that is uncommon to not explicitly declare that dependency in the `package.json`. I would keep the status quo even if it opens up the possibility to miss-match the two versions. On the other hand, if somebody is willing to do some research on the topic an provide a good explanation of how it should work I am open to removing the dependency.

Issue #953, opened 2023-04-13
Adding CBOR support


egekorkan:
Currently, it is not possible to use CBOR in the contentType but it would be good to have it for showing multiple content type usage.

danielpeintner:
Is the section in https://github.com/eclipse/thingweb.node-wot#mediatype-support not enough?

There is a sample code how one can integrate its own codec ...

Note: we cannot (and I believe **should** not) add various codecs. People can add codecs themselves...

egekorkan:
I have created this more for a discussion and to see if there is a general interest in it. The issue came from WoT CG discussions where OCF over CBOR and CoAP ie being done via WoT (node-wot to be specific).

relu91:
> I agree with you in general, @danielpeintner, but I would say that CBOR is fundamental enough for IoT use cases to be supported by the core package (also considering that it is a formal Internet Standard, just as JSON). I think I would rather draw a line at vendor-specific media types or media types limited to a specific protocol like NETCONF.

I would agree with Jan here, I think we can manage to handle also CBOR.

danielpeintner:
I am not opposed of adding CBOR. My main concern is that we keep adding packages and codecs (and the associated libraries).

We once had the idea to maintain just the core part in node-wot and bindings should come on top. 
Adding codec goes in the opposite direction...

JKRhb:
Hmm, would that imply that new and maybe also already existing Codecs should be moved to their own binding packages? (In this case, for example, something like `binding-cbor`?)

My point was simply that if there are Codecs in the core package, then CBOR should probably be one of them. Then again, CBOR is mostly used with CoAP so I can understand your concern that this additional dependency is not really necessary for something like HTTP or MQTT.

danielpeintner:
At the moment codecs are in `core` which I think is not the right place.

I envision something like the following for packages
* binding-http
* binding-coap
* ...
 
and in the same spirit for codecs

* codec-json
* codec-plain-text
* ...

For the "bindings" we are there while for "codecs" we are not. 

In the end, an application can pick the necessary binding(s) and codec(s).

What do you think? Does that make sense or is it to much hurdles?
Any thought appreciated...

egekorkan:
> @egekorkan do you mean "what happens now"?

No no, it was just a question for the future direction but I think you have answered it like I was expecting :)

JKRhb:
[...]

PR #958, opened 2023-04-23
feat: add initial CBOR codec support

JKRhb:
Corresponding with #953, this PR adds initial support for CBOR encoding and decoding. For now, the Codec is added to the `core` package, but as discussed in the issue, it should probably be moved to its own `codec-cbor` package, either in this PR or a follow-up.

Future additions to the codec should include support for custom tag encoding/decoding[^1], but for now, at least all basic JavaScript data types should be supported (via the [cbor](https://github.com/hildjj/node-cbor) package).

[^1]: See [RFC 8949, section 3.4](https://www.rfc-editor.org/rfc/rfc8949.html#name-tagging-of-items) for more information on CBOR tags (sorry for confusing the purpose of tags a bit at the community meetup on Thursday 😅).


package.json
```diff
@@ -77,5 +77,8 @@
         "tslint": "5.12.1",
         "typescript": "4.4.3",
         "typescript-standard": "^0.3.36"
+    },
+    "dependencies": {
+        "cbor": "^8.1.0"
```
JKRhb, line 82: I accidentally added this dependency to the wrong package – I will update the PR shortly.

egekorkan:
@JKRhb could you also change the readme at https://github.com/eclipse/thingweb.node-wot/blob/master/README.md?plain=1#L79

JKRhb:
Thank you for testing! :)

> @JKRhb could you also change the readme at https://github.com/eclipse/thingweb.node-wot/blob/master/README.md?plain=1#L79

Done in 1e3e895c9341f045868a598adf6ddc460e7a6507 :)

danielpeintner (approved):
LGTM

relu91 (approved):
Awesome! 🥳

PR #968, opened 2023-05-02
TestThing - fire events on write

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/966

### DIFF
--- a/examples/testthing/testthing.js
+++ b/examples/testthing/testthing.js
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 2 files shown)


danielpeintner:
@egekorkan please take a look

egekorkan (approved):

egekorkan:
I think that it is good and can be merged. One improvement could be explaining in the description of the events that they are fired when a property is written. If someone does not know the source code, they do not know what to do to trigger them.

relu91 (approved):

relu91:
> I think that it is good and can be merged. One improvement could be explaining in the description of the events that they are fired when a property is written. If someone does not know the source code, they do not know what to do to trigger them.

I think we can add this in the future. Meanwhile, let's move on with this one.

PR #971, opened 2023-05-05
chore: update dev version to 0.8.7

danielpeintner:


relu91 (commented):
@danielpeintner there are mysterious additions and deletions in the `package-lock.json` do you know why? can you clean them up?

danielpeintner:
I think you mean that lines like the following lines got removed
"license": "Apache-2.0",
"license": "MIT",
...

To be honest I don't know what happened. I re-created the package-lock.json and they are still not there.
I don't think is an issue tough ... what do you think?

danielpeintner:
@relu91 do you think we can go on with this PR?

relu91:
I'm still a little bit worried to see other random changes in the future, but let's go 😄 .

relu91 (approved):

PR #974, opened 2023-05-05
chore: add Node.js 20 and remove Node.js 14 workflow

danielpeintner:
Note: we can also start "thinking" about removing Node.js 14

### DIFF
--- a/.github/workflows/ci-full.yaml
+++ b/.github/workflows/ci-full.yaml
@@ -8,7 +8,7 @@ jobs:
         strategy:
             matrix:
                 os: [macos-latest, ubuntu-latest, windows-latest]
-                node-version: [14.x, 16.x, 18.x, 20.x]
+                node-version: [16.x, 18.x, 20.x]
 
         steps:
             - name: Checkout
--- a/.github/workflows/ci-full.yaml
+++ b/.github/workflows/ci-full.yaml
@@ -8,7 +8,7 @@ jobs:
         strategy:
             matrix:
                 os: [macos-latest, ubuntu-latest, windows-latest]
-                node-version: [14.x, 16.x, 18.x]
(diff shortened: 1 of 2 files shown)


egekorkan:
> Note: we can also start "thinking" about removing Node.js 14

Yup we can remove it. We used to keep older versions due to Eclipse server but that is nicely updated to 16 and we can update further if needed.

danielpeintner:
Note: I removed Node.js 14 from GitHub action workflow.
Question: shall we exclude mbus from package.json like we do for firestore till we have a solution?
https://github.com/eclipse/thingweb.node-wot/blob/master/package.json#L41

relu91 (approved):
Tests are all green I would merge it.

danielpeintner:
Interesting. The test used to fail because of issues with mbus but they seem to be gone after i restarted the workflow.

@egekorkan you noticed the same issues with mbus and Node.JS 20.
Can you check whether they are gone on our side as well

egekorkan:
> Interesting. The test used to fail because of issues with mbus but they seem to be gone after i restarted the workflow.
> 
> @egekorkan you noticed the same issues with mbus and Node.JS 20. Can you check whether they are gone on our side as well

Yes it is gone as well

danielpeintner:
The workflow now refuses to run
see https://github.com/eclipse/thingweb.node-wot/actions/runs/4960136649

Let's see whether this issue is just the case for this specific case

JKRhb:
> The workflow now refuses to run see https://github.com/eclipse/thingweb.node-wot/actions/runs/4960136649
> 
> Let's see whether this issue is just the case for this specific case

Hmm, I think there is an opening square bracket missing in the workflow file.

relu91:
Some how we got a greenlight from Github but he didn't actually run CI. Something we have to look after next time :/

PR #976, opened 2023-05-07
fix(binding-mqtt): fix incorrect segment indices and change index and…

### LINKED ISSUE #874: Bug in MQTT topic parsing
Creating an action, and trying to invoke it with MQTT, I can see the following in the logs (device: `wot1`, action: `reset`):

```
  mqttjs:client sendPacket :: packet: {
  mqttjs:client   cmd: 'subscribe',
  mqttjs:client   subscriptions: [
  mqttjs:client     {
  mqttjs:client       topic: 'wot1/actions/reset',
  mqttjs:client       qos: 0,
  mqttjs:client       nl: false,
  mqttjs:client       rap: false,
  mqttjs:client       rh: 0,
  mqttjs:client       properties: undefined
  mqttjs:client     }
  mqttjs:client   ],
  mqttjs:client   qos: 1,
  mqttjs:client   retain: false,
  mqttjs:client   dup: false,
  mqttjs:client   messageId: 21279
  mqttjs:client } +0ms
[…]
node-wot:binding-mqtt:mqtt-broker-server:warn MqttBrokerServer at undefined received message for invalid topic 'wot1/actions/reset' 
```

So the device subscribes to `wot1/actions/reset` and then claims a command to `wot1/actions/reset` is invalid.

Taking a look at the code, the bug is simple to spot:

https://github.com/eclipse/thingweb.node-wot/blob/68084d2afc60e7e162462e061cdcbee2fab64978/packages/binding-mqtt/src/mqtt-broker-server.ts#L195

It checks for 4 segments. However, the topic of an action command will only have 3! The same error is made a few lines below for properties.

I assume the implementation expects a leading slash, in which cases the indices would match. However, during the subscription the a leading slash is not used.

So the implementation must decide, either use a leading slash for BOTH the topic filter during subscription and the topic parsing during processing. Or never use/expect it.

The way it is implemented right now, it seems just broken.

hasanheroglu:
… length values to enums

Closes #874

### DIFF
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -39,6 +39,18 @@ import { ActionElement, PropertyElement } from "wot-thing-description-types";
 
 const { info, debug, error, warn } = createLoggers("binding-mqtt", "mqtt-broker-server");
 
+enum SegmentLength {
+    "action" = 3,
+    "property" = 4,
+}
+
+enum SegmentIndex {
+    "thingName" = 0,
+    "interactionType" = 1,
+    "interactionName" = 2,
+    "interactionExtension" = 3,
+}
+
 export default class MqttBrokerServer implements ProtocolServer {
     readonly scheme: string = "mqtt";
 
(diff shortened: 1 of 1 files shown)


danielpeintner:
@egekorkan can you take a look

danielpeintner (commented):
Question: Can we come up with a test to avoid regressions in the future?
I am not sure, but maybe some very basic interactions with https://test.mosquitto.org/ ?

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -39,6 +39,18 @@ import { ActionElement, PropertyElement } from "wot-thing-description-types";
...
+    "property" = 4,
+}
+
+enum SegmentIndex {
+    "thingName" = 0,
+    "interactionType" = 1,
+    "interactionName" = 2,
+    "interactionExtension" = 3,
+}
+
```
relu91, line 53: Can you explain your trail of thoughts to use an enum to model topic parsing?

relu91 (commented):
Changes look good, but before approving I agree with Daniel that it is better to have test. You can add one at the https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts . There is no need to depend on https://test.mosquitto.org/. Finally, I added a question below.

packages/binding-mqtt/src/mqtt-broker-server.ts
hasanheroglu: For me it was really easy to mix up numbers while coding so I thought it would be easier to use something more explanatory. Since topic segments always have the similar order, I think with enums the code becomes more readable.

egekorkan:
I like the changes, thanks @hasanheroglu ! Tests would be indeed very beneficial

packages/binding-mqtt/src/mqtt-broker-server.ts
danielpeintner: Personally I would not have used enums but I will definitely not object if you prefer keeping it.
Note: In other bindings we don't use enums for the same type of hierarchy.

relu91: If it is not a problem I would follow the principles of "less surprise" and "consistency". You can use a prefix or suffix to group semantically `const`. Sorry, for being picky you did a great job already but I think it is worth doing this small refactor.

hasanheroglu: I did the change here: [3952483](https://github.com/eclipse/thingweb.node-wot/pull/976/commits/395248339d74b23bf1f41cbd22c70c01988859fc).

relu91 (approved):
Thank you really much! it's good to go.

PR #981, opened 2023-05-08
fix: report SVG values without quotes

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/965

### DIFF
--- a/packages/core/src/codecs/text-codec.ts
+++ b/packages/core/src/codecs/text-codec.ts
@@ -91,7 +91,14 @@ export default class TextCodec implements ContentCodec {
         if (be) {
             return Buffer.from(body, be);
         } else {
-            return Buffer.from(body);
+            if (this.subMediaType === "image/svg+xml" && typeof value === "string") {
+                // Note: necessary to get buffer without quotes around the string value
+                const buff = Buffer.alloc(value.length);
+                buff.write(value);
+                return buff;
+            } else {
+                return Buffer.from(body);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


danielpeintner:
Note: I think it is a very hack fix and I am not sure what we can/should to else.

`Buffer.from("XXX");`

returns a buffer of the `XXX` string around with quotes.

packages/core/src/codecs/text-codec.ts
```diff
@@ -91,7 +91,14 @@ export default class TextCodec implements ContentCodec {
         if (be) {
             return Buffer.from(body, be);
         } else {
-            return Buffer.from(body);
+            if (this.subMediaType === "image/svg+xml" && typeof value === "string") {
```
danielpeintner, line 94: Question: Maybe we can add an option in the TextCodec constructor whether wrapping is needed or not. That makes it more generic. What do you think?

danielpeintner: Created `bufferDirectWrite` options since it is needed for HTML, SVG etc as well..

relu91 (changes requested):
Checked https://datatracker.ietf.org/doc/html/rfc822#appendix-A and it seems that also text/plain does not require quotes around the body, so I think we can get rid of the boolean modifier and always return the unquoted value. 

Also as a solution to the issue, instead of checking if the value in input is a string and doing Buffer manipulation, why don't we do the following:
```ts
        let body = "";
        if (value !== undefined && typeof(value) === 'string') {
            body = value
        } else {
            body = JSON.stringify(value)
        }
```

danielpeintner:
@relu91 I integrated you proposed changes. Please have a look.

relu91 (approved):
LGTM

PR #983, opened 2023-05-08
feat(binding-mqtt): add write support

hasanheroglu:


danielpeintner:
Before merging we should add tests.

packages/binding-mqtt/src/mqtt-client.ts
```diff
@@ -100,7 +100,21 @@ export default class MqttClient implements ProtocolClient {
...
+        const topic = requestUri.pathname.slice(1);
+        const brokerUri = `${this.scheme}://${requestUri.host}`;
+
+        if (this.client === undefined) {
+            this.client = mqtt.connect(brokerUri, this.config);
+        }
+
+        // if not input was provided, set up an own body otherwise take input as body
+        if (content === undefined) {
+            this.client.publish(topic, JSON.stringify(Buffer.from("")));
```
relu91, line 113: Why the extra `JSON.stringify`?

hasanheroglu: I simply took the code from `invokeResource` below this function since they logically do the same thing, but It seems we don't need JSON.stringy in those places. Fixed here: [8cc4518](https://github.com/eclipse/thingweb.node-wot/pull/983/commits/8cc45187f96e92148b967c7dc57d9fc0e8f44ab9)

danielpeintner (approved):

PR #985, opened 2023-05-09
fix: report allproperties with proper type

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/984

### DIFF
--- a/packages/binding-http/test/http-server-test.ts
+++ b/packages/binding-http/test/http-server-test.ts
@@ -144,8 +144,8 @@ class HttpServerTest {
         resp = await (await fetch(uri + "properties/test")).text();
         expect(resp).to.equal('"off"');
 
-        resp = await (await fetch(uri + "properties")).text();
-        expect(resp).to.equal('{"test":"\\"off\\""}');
+        resp = await (await fetch(uri + "properties")).json();
+        expect(resp).to.deep.equal({ test: "off" });
 
         resp = await (await fetch(uri + "properties/test", { method: "PUT", body: "on" })).text();
         expect(resp).to.equal("");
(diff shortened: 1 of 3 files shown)


danielpeintner:
Note: readallproperties does not report any value other than JSON. In the test added SVG is not reported. I think this is the best we can do... 
@egekorkan maybe also a topic for TD.

egekorkan:
I agree that this is a spec issue (even more so than an implementation issue)

egekorkan:
See https://github.com/w3c/wot-thing-description/issues/1814 for more discussions

packages/binding-http/test/http-server-test.ts
```diff
@@ -549,6 +549,100 @@ class HttpServerTest {
         return httpServer.stop();
     }
 
+    @test async "should report allproperties"() {
```
relu91, line 552:
```suggestion
    @test async "should report allproperties excluding no json properties"() {
```
or create two tests with one with only json properties and one like the current version.

relu91 (changes requested):
I definitely agree this is the best we can do and if properly specified in the spec is not even a problem I think. 

What is a little bit weird with the current architecture is that the HTTP server needs to know how to serialize the payload. I would not block this PR for this, but I think in the future we should have a second look.

packages/binding-http/test/http-server-test.ts
danielpeintner: I had to re-read it multiple times to get it right.. but I think the addition should read
"excluding non-JSON properties"

danielpeintner, line 552:
```suggestion
    @test async "should report allproperties excluding non-JSON properties"() {
```

relu91 (approved):
Committed changes, ready to go

PR #987, opened 2023-05-09
fix(binding-mqtt): fix sending plain value instead of content object

### LINKED ISSUE #979: Mqtt handleAction and handleWriteProperty do not work as expected
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/binding-mqtt/src/mqtt-broker-server.ts#L269

https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/binding-mqtt/src/mqtt-broker-server.ts#L303

The lines above send just a JSON object or string (in case of a parsing error for handleAction) to `handleInvokeAction` and `handleWriteProperty` function, where functions need a `Content` object instead. And with these non-Content values `InteractionOutput`s are created, after we make a call for `InteractionOutput`s value error occurs because `InteractionOutput`s content is not actually a content but JSON object or string.

I have seen that for other protocol bindings using forms' contentType, `Content` object is created: 
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/binding-coap/src/coap-server.ts#L456-L460

Should we also apply same kind of logic here or should we follow some other method?

hasanheroglu:
Closes #979

### DIFF
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -304,7 +304,7 @@ export default class MqttBrokerServer implements ProtocolServer {
             try {
                 thing.handleWriteProperty(
                     segments[3],
-                    new Content(ContentSerdes.DEFAULT, Readable.from(JSON.parse(payload.toString()))),
+                    new Content(ContentSerdes.DEFAULT, Readable.from(payload.toString())),
                     options
                 );
             } catch (err) {
--- a/packages/binding-mqtt/src/mqtt-broker-server.ts
+++ b/packages/binding-mqtt/src/mqtt-broker-server.ts
@@ -267,8 +267,11 @@ export default class MqttBrokerServer implements ProtocolServer {
             ),
         };
 
+        const contentType = action.forms[options.formIndex].contentType ?? ContentSerdes.DEFAULT;
+        const inputContent = new Content(contentType, Readable.from(value));
+
         thing
-            .handleInvokeAction(segments[3], new Content(ContentSerdes.DEFAULT, Readable.from(value)), options)
+            .handleInvokeAction(segments[3], inputContent, options)
             .then((output: unknown) => {
                 if (output) {
                     warn(`MqttBrokerServer at ${this.brokerURI} cannot return output '${segments[3]}'`);
@@ -301,10 +304,13 @@ export default class MqttBrokerServer implements ProtocolServer {
                 ),
             };
 
+            const formContentType = property.forms[options.formIndex].contentType ?? ContentSerdes.DEFAULT;
+            const inputContent = new Content(formContentType, Readable.from(payload.toString()));
+
             try {
                 thing.handleWriteProperty(
                     segments[3],
-                    new Content(ContentSerdes.DEFAULT, Readable.from(payload.toString())),
+                    inputContent,
                     options
                 );
             } catch (err) {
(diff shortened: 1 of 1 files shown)


hasanheroglu:
Since segments indices are wrong, I could not add tests. If it is okay, I will add the tests with segment indices fix.

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -300,7 +306,11 @@ export default class MqttBrokerServer implements ProtocolServer {
             };
 
             try {
-                thing.handleWriteProperty(segments[3], JSON.parse(payload.toString()), options);
+                thing.handleWriteProperty(
+                    segments[3],
+                    new Content(ContentSerdes.DEFAULT, Readable.from(payload.toString())),
```
danielpeintner, line 311: Question: Should it be 
new Content(ContentSerdes.DEFAULT, Readable.from(JSON.parse(payload.toString()))
as it used to be in 
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/binding-mqtt/src/mqtt-broker-server.ts#L303

Please check.

danielpeintner:
> Since segment indices are wrong, I could not add tests. If it is okay, I will add the tests with segment indices fix.

I think the fix is easy enough that tests can (and should) follow later.
@egekorkan @relu91

packages/binding-mqtt/src/mqtt-broker-server.ts
danielpeintner: Note: this might need checks/tests especially with a variation of input: simple types (int, boolean, etc) vs composed types (like objects and arrays)

hasanheroglu: Checking now

hasanheroglu: Passing JSON object to `Readable.from` creates an issue except when JSON only a string, therefore it seems like we should not use `JSON.parse`.

danielpeintner: Thanks for checking 👍
We should add tests in follow-up PR for it.

egekorkan:
We will definitely follow up with the tests since so many PRs are interlinked at the moment.

danielpeintner (approved):

relu91 (changes requested):

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -266,7 +268,7 @@ export default class MqttBrokerServer implements ProtocolServer {
         };
 
         thing
-            .handleInvokeAction(segments[3], value, options)
+            .handleInvokeAction(segments[3], new Content(ContentSerdes.DEFAULT, Readable.from(value)), options)
```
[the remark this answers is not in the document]
hasanheroglu: Should we also apply the same for handlePropertyWrite? In there we try to get contentType from `packet.properties` but not from the forms.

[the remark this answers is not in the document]
hasanheroglu: Fixed here: [c32b712](https://github.com/eclipse/thingweb.node-wot/pull/987/commits/c32b712a0009913f077994bba17ede76bac7017e)

relu91 (approved):
Thank you LGTM

Issue #982, opened 2023-05-08
(binding-coap) options field send as null to handleObserveProperty and handleUnobserveProperty


hasanheroglu:
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/binding-coap/src/coap-server.ts#L427-L442

`options` field send as null and leads to an error in the core package:
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/core/src/exposed-thing.ts#L621-L627
https://github.com/eclipse/thingweb.node-wot/blob/09dfd6ab03805b7b39ab39433880496466d25e67/packages/core/src/exposed-thing.ts#L653-L658

Is there a specific reason we send options field null here?

egekorkan:
Just to be specific, the error pops up when trying to observe properties in coap

relu91:
@JKRhb after the different refactorings is this issue handled?

JKRhb:
Thank you for the ping, @relu91 :) During the refactoring, I actually noticed this bug as well and after double-checking, I think it should now be fixed! Therefore, I would close this issue – please leave a comment, @hasanheroglu, if you should notice that the issue persists!

PR #988, opened 2023-05-09
refactor(binding-coap): refactor server's destroy method

JKRhb:
In the context of #963. I noticed that the `CoapServer`'s `destroy` method could use some cleanup. This PR refactors the method a bit, using `async/await` instead of a `Promise`, making the code a bit more readable.

There are other places like these elsewhere in the code, I will try to do some further cleanup when I'll continue working on #963 and issues like #982.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -194,25 +194,21 @@ export default class CoapServer implements ProtocolServer {
         });
     }
 
-    public destroy(thingId: string): Promise<boolean> {
+    public async destroy(thingId: string): Promise<boolean> {
         debug(`CoapServer on port ${this.getPort()} destroying thingId '${thingId}'`);
-        return new Promise<boolean>((resolve, reject) => {
-            let removedThing: ExposedThing;
-            for (const name of Array.from(this.things.keys())) {
-                const expThing = this.things.get(name);
-                if (expThing?.id === thingId) {
-                    this.things.delete(name);
-                    this.coreResources.delete(name);
-                    removedThing = expThing;
-                }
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):

packages/binding-coap/src/coap-server.ts
```diff
@@ -194,25 +194,21 @@ export default class CoapServer implements ProtocolServer {
...
-                    this.things.delete(name);
-                    this.coreResources.delete(name);
-                    removedThing = expThing;
-                }
-            }
-            if (removedThing) {
-                info(`CoapServer succesfully destroyed '${removedThing.title}'`);
-            } else {
-                info(`CoapServer failed to destroy thing with thingId '${thingId}'`);
+        for (const name of Array.from(this.things.keys())) {
```
JKRhb, line 199: Having another look at the code, I think this could further be simplified like so:

```suggestion
        for (const name of this.things.keys()) {
```

(I can't test it out locally at the moment, therefore I left this comment as a note for later.)

relu91: Tried on node console, `Array.from` seems completely redundant.

relu91 (approved):

packages/binding-coap/src/coap-server.ts
JKRhb: Thank you for confirming! :) Then I will apply this suggestion and you could squash and merge this PR :)

JKRhb: The commits are now already squashed, I made a minor change to one variable name, the PR should now be ready :)

PR #990, opened 2023-05-11
chore: remove mbus and firestore bindings

danielpeintner:
link last supported version in README.md


relu91:
@danielpeintner can we close this PR?

danielpeintner:
Sure, let's close it.

Firestore has been removed in the meanwhile already, see https://github.com/eclipse-thingweb/node-wot/pull/1021.

One reason of removing both bindings was that it failed to compile on Node.js 20. 
We still need to look into that issue for mbus, see https://github.com/eclipse-thingweb/node-wot/issues/1004.

PR #992, opened 2023-05-11
Tests for mqtt interactions

hasanheroglu:


hasanheroglu:
Tests were failing for `handleAction` because of input handling. So I used same logic  as `handlePropertyWrite` for `handleAction`.

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
```diff
@@ -155,4 +155,432 @@ describe("MQTT client implementation", () => {
...
+                    WoT.consume(thing.getThingDescription()).then((client) => {
+                        const input = "writeProperty";
+
+                        client
+                            .writeProperty(propertyName, input)
+                            .then(() => {
+                                setTimeout(() => {
+                                    expect(input).to.equal(receivedInput);
+                                    thing.destroy().then(() => done());
+                                }, 100);
```
relu91, line 201: I'm not a fan of timeouts and I try to use them as less as possible. Using a `qos` up to 2 should make the publishing synchronous and therefore you should be able to verify the execution of the writing operation without timeouts.

relu91 (changes requested):
Some inputs:
- The testing file is called mqtt-client-**subscribe**-test.integration.ts : the new tests are about writing and invoking actions. We have two options: rename the file to mqtt-client-test.integration.ts and use a describe to group the different tests OR move the new tests in a new file.
- You tested for different input types which is great, but there is a lot of duplicated code. Can you try to refactor it to have less code duplicated? 
- I left some minor comments below about timeouts and failing.

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
hasanheroglu: I don't know if qos works in the implementation because I haven't seen any dependent logic to qos, could you explain it more how it works? Also If the mqtt broker server is slow would it matter we have qos 2 or not? We could get confirmation that we published our message but we might not know how long the process will take in the broker server.

danielpeintner:
Note: I noticed you merged in master which was not a good idea since the testing is broken at the moment but we are about to fix that, see https://github.com/eclipse-thingweb/node-wot/pull/995
You may need to merge master once again when the PR is merged.

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
egekorkan: @relu91 actually even with qos 2 and the function being synchronous, the Thing can take a while to process the input it received. I think that this timeout is necessary but can be smaller. However, running the action takes too long and 20ms was not enough at [this commit.](https://github.com/eclipse-thingweb/node-wot/pull/992/commits/ae91065e408cb2f0e4f430912dedc74fb11c611f#diff-c046a35f74347f972fe11252680105f35586ad7ef2cfbc6d2fd6376e105708c8R201).

Regarding making it qos 2, I think that this is a missing feature. At this [line](https://github.com/eclipse-thingweb/node-wot/blob/badc223b330d588f336435589270ffb0a41ad1a7/packages/binding-mqtt/src/mqtt-client.ts#L102), we are not using qos of the forms at all.

Ideally, this test should happen inside the write handler. Since it is a callback, we are not sure if it will work but @hasanheroglu will give it a try.

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -246,42 +246,22 @@ export default class MqttBrokerServer implements ProtocolServer {
          * https://github.com/mqttjs/MQTT.js/pull/1103
          * For further discussion see https://github.com/eclipse-thingweb/node-wot/pull/253
          */
-        let value;
-
+        let contentType = ContentSerdes.DEFAULT;
```
JKRhb, line 249: I think this and the following lines could be simplified like so:

```suggestion
        const contentType = packet?.properties?.contentType ?? ContentSerdes.DEFAULT;
```

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -312,7 +292,7 @@ export default class MqttBrokerServer implements ProtocolServer {
     ) {
         if (!property.readOnly) {
             let contentType = ContentSerdes.DEFAULT;
```
JKRhb, line 294: Same here:

```suggestion
            const contentType = packet?.properties?.contentType ?? ContentSerdes.DEFAULT;
```

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -246,42 +246,22 @@ export default class MqttBrokerServer implements ProtocolServer {
...
                 this.brokerURI,
-                ContentSerdes.DEFAULT
+                contentType
             ),
         };
 
-        const contentType = action.forms[options.formIndex].contentType ?? ContentSerdes.DEFAULT;
-        const inputContent = new Content(contentType, Readable.from(value));
+        const formContentType = action.forms[options.formIndex].contentType ?? ContentSerdes.DEFAULT;
+        const inputContent = new Content(formContentType, Readable.from(payload.toString()));
```
JKRhb, line 264: Since `payload` is a `Buffer`, is it actually necessary to call `toString()` here?

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
relu91: Using QOS 2 makes it equivalent to an HTTP POST/PUT, it is true that the thing could take a while but that's why there is a global timeout defined for each test. It should be something exceptional that would not happen in the circumstances of this test (cause usually, everything is local). 

About QoS 2 I was thinking that we had that, but the idea of testing it inside the writeHandler is not completely bad (even though is not a proper e2e test, meaning that we could mess up the writing of the payload somehow).

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
```diff
@@ -155,4 +155,432 @@ describe("MQTT client implementation", () => {
...
+                            })
+                            .catch((e) => {
+                                expect(true).to.equal(false);
```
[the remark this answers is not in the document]
hasanheroglu: I completely refactored the structure and to fail I used `done` with exception message.

packages/binding-mqtt/src/mqtt-broker-server.ts
hasanheroglu: Fixed at [48e615f](https://github.com/eclipse-thingweb/node-wot/pull/992/commits/48e615f4288766d88cc6e0e93cbc1de84956afc6)

packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
hasanheroglu: `expect` and `done` seems to work in a callback. New tests pass.

relu91 (approved):
Much better now! thank you! I think we can move on and fix the details in the future.

danielpeintner (approved):
It's good to see code coverage improve 👍

PR #993, opened 2023-05-11
refactor(binding-coap): refactor server's expose method

JKRhb:
This PR should further improve the code quality in the CoAP server implementation by refactoring the `expose` method. Some of the new code might be able to be more generalized and moved to the `core` or `td-tools` instead.

For some reason, the current diff is huge, caused by the `package-lock.json` file. Not only because of that, there is still a little more work required until this PR will be ready for review.

### DIFF
--- a/packages/binding-coap/package.json
+++ b/packages/binding-coap/package.json
@@ -43,6 +43,7 @@
         "node-coap-client": "1.0.8",
         "rxjs": "5.5.11",
         "slugify": "^1.4.5",
+        "wot-thing-description-types": "1.1.0-23-March-2023",
         "wot-typescript-definitions": "0.8.0-SNAPSHOT.25"
     },
     "scripts": {
--- a/packages/binding-coap/package.json
+++ b/packages/binding-coap/package.json
@@ -43,7 +43,6 @@
         "node-coap-client": "1.0.8",
         "rxjs": "5.5.11",
         "slugify": "^1.4.5",
-        "wot-thing-description-types": "1.1.0-23-March-2023",
         "wot-typescript-definitions": "0.8.0-SNAPSHOT.25"
     },
     "scripts": {
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -205,69 +205,73 @@ export default class CoapServer implements ProtocolServer {
         return [href, form];
     }
 
-    public async expose(thing: ExposedThing, tdTemplate?: WoT.ExposedThingInit): Promise<void> {
-        const urlPath = this.createThingUrlPath(thing);
-        const port = this.getPort();
+    private fillInPropertyBindingData(thing: ExposedThing, base: string, port: number, offeredMediaType: string) {
+        for (const [propertyName, property] of Object.entries(thing.properties)) {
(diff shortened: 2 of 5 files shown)


packages/binding-coap/package.json
```diff
@@ -43,6 +43,7 @@
         "node-coap-client": "1.0.8",
         "rxjs": "5.5.11",
         "slugify": "^1.4.5",
+        "wot-thing-description-types": "^1.1.0-23-March-2023",
```
danielpeintner, line 46:
```suggestion
        "wot-thing-description-types": "1.1.0-23-March-2023",
```

I other package.json we use a fix version since in the past we sometimes run into issues.

packages/binding-coap/src/coap-server.ts
```diff
@@ -136,81 +137,145 @@ export default class CoapServer implements ProtocolServer {
...
         }
+
+        return urlPath;
+    }
+
+    private createCoreResource(urlPath: string): void {
         this.coreResources.set(urlPath, { urlPath, parameters: thingDescriptionParameters });
+    }
+
+    // TODO: Could probably be defined as a general helper function
```
danielpeintner, line 154: That sounds like a good idea 👍

relu91:
Well done, I like the direction of simplify the expose method 👍🏻 .

JKRhb:
Okay, by using `npm install wot-thing-description-types@1.1.0-23-March-2023 --save --save-exact --workspace=./packages/binding-coap` I have now been able to cut down the number of lines added to the `package-lock.json` file from about 10.000 to 1. npm can be a bit strange sometimes :D

JKRhb:
I performed some additional refactoring in the latest commit. I think there is still some room for improvement, but that should probably be done in a follow-up PR to keep the PR's size manageable :)

relu91 (approved):
It's definitely more readable, I might have a slight preference for having the expose method before the "inner functions". I prefer having a top-down view starting from the public methods and then going down to the details. I'll let @danielpeintner decide.

JKRhb:
> It's definitely more readable, I might have a slight preference for having the expose method before the "inner functions". I prefer having a top-down view starting from the public methods and then going down to the details.

Oh, good point, I will reorder the methods and put the expose method at the top

packages/binding-coap/src/coap-server.ts
danielpeintner: @JKRhb do you still plan to create it as a general helper function? Maybe even in `core`?

packages/core/package.json
```diff
@@ -40,6 +40,7 @@
         "ts-node": "10.1.0",
         "typescript": "4.4.3",
         "typescript-standard": "^0.3.36",
+        "wot-thing-description-types": "1.1.0-23-March-2023",
```
danielpeintner: Not sure if this this dependency is actually needed since `wot-typescript-definitions` depends on `wot-thing-description-types`

see 
https://github.com/w3c/wot-scripting-api/blob/c92f945c24495cbf239cf847d8d96ef8930e8ec9/typescript/scripting-api/package.json#L26

danielpeintner (approved):

PR #994, opened 2023-05-11
refactor(coap-server): refactor request handling

JKRhb:
This PR applies some additional refactoring to the `CoapServer` class, removing some nesting from the `handleRequest` method. In follow-up PRs, the logic of this and other methods needs to be further refactored in order to reduce complexity and make the class more maintainable.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -32,10 +32,14 @@ import { Server, createServer, registerFormat, IncomingMessage, OutgoingMessage 
 import slugify from "slugify";
 import { Readable } from "stream";
 import { MdnsIntroducer } from "./mdns-introducer";
+import { PropertyElement, DataSchema } from "wot-thing-description-types";
 
 const { debug, warn, info, error } = createLoggers("binding-coap", "coap-server");
 
 type CoreLinkFormatParameters = Map<string, string[] | number[]>;
+
+// TODO: Move to core?
+type AugmentedInteractionOptions = WoT.InteractionOptions & { formIndex: number };
 
 const thingDescriptionParameters: CoreLinkFormatParameters = new Map(
     Object.entries({
@@ -56,8 +60,8 @@ export default class CoapServer implements ProtocolServer {
     private readonly ACTION_DIR = "actions";
     private readonly EVENT_DIR = "events";
 
-    private readonly port: number = 5683;
-    private readonly address?: string = undefined;
+    private readonly port: number;
+    private readonly address?: string;
 
     private mdnsIntroducer: MdnsIntroducer;
 
@@ -73,12 +77,8 @@ export default class CoapServer implements ProtocolServer {
     private readonly coreResources = new Map<string, CoreLinkFormatResource>();
 
     constructor(port?: number, address?: string) {
-        if (port !== undefined) {
-            this.port = port;
-        }
-        if (address !== undefined) {
-            this.address = address;
-        }
+        this.port = port ?? 5683;
+        this.address = address;
 
         // WoT-specific content formats
         registerFormat(ContentSerdes.JSON_LD, 2100);
@@ -307,15 +307,13 @@ export default class CoapServer implements ProtocolServer {
     }
 
     private handleWellKnownCore(req: IncomingMessage, res: OutgoingMessage) {
-        if (req.method === "GET") {
-            res.setOption("Content-Format", "application/link-format");
-            res.code = "2.05";
-            const payload = this.formatCoreLinkFormatResources();
-            res.end(payload);
-        } else {
-            res.code = "4.05";
-            res.end("Method Not Allowed");
+        if (req.method !== "GET") {
+            this.sendMethodNotAllowedResponse(res);
+            return;
         }
+
+        const payload = this.formatCoreLinkFormatResources();
+        this.sendContentResponse(res, payload, "application/link-format");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


JKRhb:
I think this PR should now finally be ready for review :) Unfortunately, the PR size turned out to be a lot larger than originally intended. However, I think the code should now be significantly more concise and hopefully also more readable, which should make it easier to extend the CoapServer in the future.

Some helper functions might be able to be generalized and moved to the core package in follow-up PRs. Also, there are still some workarounds in the code referring to bugs or missing features in node-coap, some of which might be fixed in the meantime. I will investigate that and will otherwise –  if the issues should still be present – try to resolve them.

packages/binding-coap/src/coap-server.ts
```diff
@@ -32,11 +32,15 @@ import { Server, createServer, registerFormat, IncomingMessage, OutgoingMessage
...
 import { Readable } from "stream";
 import { MdnsIntroducer } from "./mdns-introducer";
+import { PropertyElement, DataSchema } from "wot-thing-description-types";
 
 const { debug, warn, info, error } = createLoggers("binding-coap", "coap-server");
 
 type CoreLinkFormatParameters = Map<string, string[] | number[]>;
 
+// TODO: Move to core?
+type AugmentedInteractionOptions = WoT.InteractionOptions & { formIndex: number };
```
relu91, line 42: Why doesn't `WoT.InteractionOptions` have `formIndex` already? Current scripting API already mention it: https://w3c.github.io/wot-scripting-api/#dom-interactionoptions-formindex also in TS file they are indeed present:https://github.com/w3c/wot-scripting-api/blob/06a31b1efb06d8b1ec996cd18dab25c71a98b2b1/typescript/scripting-api/index.d.ts#L175.

packages/binding-coap/src/coap-server.ts
```diff
@@ -337,387 +335,471 @@ export default class CoapServer implements ProtocolServer {
...
-                ContentSerdes.get().getSupportedMediaTypes().indexOf(ContentSerdes.getMediaType(contentType)) < 0
-            ) {
-                res.code = "4.15";
-                res.end("Unsupported Media Type");
-                return;
+        thing
+            .handleObserveProperty(affordanceKey, listener, null)
+            .then(() => res.end())
+            .catch(() => res.end());
+
```
relu91, line 521: Sorry for the current formatting is hard to tell, but can't we use `trycatch` and `await` here?

relu91 (approved):
I have only questions but the content of the PR looks good!

packages/binding-coap/src/coap-server.ts
JKRhb: Good point, that has been a part of the original implementation which I found kind of strange therefore I didn't want to touch it initially 😅 I now tried to provide a proper solution as you proposed in 3b8604c4eb20954802c338b363e964f40956796e. Not sure if always invoking `res.end()` as in the original code is the correct approach here, though.

JKRhb: Slightly adjusted the code in 0bf0d5af70fbfbefcac2bc238084574a74b7cf6a to be closer to the original logic.

packages/binding-coap/src/coap-server.ts
danielpeintner: Agree, let look into it in a follow-up. Created #1039

packages/binding-coap/src/coap-server.ts
```diff
@@ -337,387 +335,478 @@ export default class CoapServer implements ProtocolServer {
...
-        } else {
+        if (isUnsupportedAcceptValue) {
             debug(`Request contained an accept option with value ${accept} which is not supported.`);
-            res.code = "4.06";
-            res.end(`Content-Format ${accept} is not supported by this resource.`);
+            this.sendErrorResponse(res, "4.06", `Content-Format ${accept} is not supported by this resource.`);
+            return;
         }
+
+        debug(`Received an available or no Content-Format (${accept}) in Accept option.`);
```
danielpeintner, line 354:
```suggestion
        debug(`Request contained accept option with value ${accept} which is supported (or default/empty).`);
```

I struggled over this debug statement and it confused me. I tried to explain it better but I have difficulties to convey the message 🤷‍♂️

packages/binding-coap/src/coap-server.ts
```diff
@@ -337,387 +335,478 @@ export default class CoapServer implements ProtocolServer {
...
+        req: IncomingMessage,
+        thing: ExposedThing,
+        res: OutgoingMessage,
+        affordanceKey: string
+    ) {
+        const listener = async (content: Content) => {
+            try {
+                this.streamContentResponse(res, content, { end: true });
+            } catch (err) {
+                error(`CoapServer on port ${this.getPort()} got internal error on read '${req.url}': ${err.message}`);
```
danielpeintner, line 512:
```suggestion
                error(`CoapServer on port ${this.getPort()} got internal error on observe '${req.url}': ${err.message}`);
```

Shouldn't it be observe?

packages/binding-coap/src/coap-server.ts
```diff
@@ -337,387 +335,478 @@ export default class CoapServer implements ProtocolServer {
...
-                        // resource found and response sent
-                        return;
-                    } // Event exists?
+        }
+
+        const observe = req.headers.Observe as number;
+
+        if (observe == null) {
+            debug(
+                `CoapServer on port ${this.getPort()} rejects '${affordanceKey}' read from ${Helpers.toUriLiteral(
```
danielpeintner, line 647: I think the debug statement might convey that it deals with an event

packages/binding-coap/src/coap-server.ts
```diff
@@ -337,387 +335,478 @@ export default class CoapServer implements ProtocolServer {
...
+            thingKey: segments[1],
+            affordanceType: segments[2],
+            affordanceKey: segments[3],
+        };
+    }
+
+    private sendContentResponse(res: OutgoingMessage, payload: Buffer | string, contentType: string) {
+        res.setOption("Content-Format", contentType);
+        res.code = "2.05";
+        res.end(payload);
```
danielpeintner, line 775: VERY MINOR
In the following code there are some parts that always read like

```
        res.code = ...;
        res.end(payload);
```

It could make sense to have an overarching method that deals with all of them.. so that all point to it..

danielpeintner (approved):
PR LGTM

Some comments below.. really just comments

packages/binding-coap/src/coap-server.ts
relu91: Thank you daniel, we can resolve this issue.

packages/binding-coap/src/coap-server.ts
JKRhb: Good point, I now turned the `sendErrorResponse` method into a generic `sendResponse` method :)

packages/binding-coap/src/coap-server.ts
JKRhb: Would this be a better alternative?

```suggestion
                `CoapServer on port ${this.getPort()} rejects '${affordanceKey}' event subscription from ${Helpers.toUriLiteral(
```

JKRhb: Took the freedom of already adding this suggestion to the PR as part of 5d0f086c474935df016f5c0b5031e1798ace0328, hope that was okay :)

relu91 (approved):
Reviewed once again, looks good!

JKRhb:
> Reviewed once again, looks good!

Awesome, thank you :) I just squashed the commits into one – the PR would be ready to merge when you are :) CC @danielpeintner

JKRhb:
> So in general, we had this non-written rule that once two committers agree on a PR we can merge it, but given the fact that daniel explicitly comment on this PR I would wait. It is not fixing any critical bug and there is no urgency. Is that ok @JKRhb?

Sure, from my side we can wait for Daniel's final approval :)

danielpeintner (approved):

danielpeintner:
PR looks good, thanks! I will merge it right away ...

> Awesome, thank you :) I just squashed the commits into one – the PR would be ready to merge when you are :) CC @danielpeintner

Side comment: I think we should avoid squashing commits unless there is a good reason.
Why do I say that? Squashing commits does no longer allow reviewers to check what has been updated recently. Hence, one needs to re-check everything again...
Having said that, I like it to squash the commits when merging the PR because doing so

1. Reviewer still see the history
2. The commit in master/main is still clean and just contains *one* commit even if the PR evolved a lot.

What do people think. Shall we setup some kind of rules for that?

danielpeintner:
What I would like to propose is the following.

* while we work on a PR and people are reviewing we should **not**  squash commits
* once we agree to merge the PR we can (and maybe should) squash all the commits into one. Hence we squash in the very last step.
 
image

relu91:
I'm ok with that, but only if the changes are closely related. I would not squash if the PR is fixing two issues or dealing with multiple aspects of the code base (e.g. refactoring + fixing).

JKRhb:
You are right @danielpeintner, sorry for squashing a bit prematurely :/ Next time, I will wait until all final approvals are given.

I guess as a general rule, maybe we can say that squashing should only be done when the reviewers gave their approval/ask a PR author to do so or during the merge process (as you've shown above).

PR #1001, opened 2023-05-17
Prefer application/td+json when is possibile instead of text/html 

### LINKED ISSUE #1000: TD for Counter is reported as HTML
I have just run the counter example from the master branch. Exposing the TD does not show JSON as the type but HTML. Not sure where the problem is coming from. The hosted version which is 2 weeks old is correct.

relu91:
Fix #1000. (we reached a nice milestone)

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -570,7 +570,16 @@ export default class HttpServer implements ProtocolServer {
 
                 return acceptValue;
             })
-            .filter((acceptValue) => contentSerdes.isSupported(acceptValue));
+            .filter((acceptValue) => contentSerdes.isSupported(acceptValue))
+            .sort((a, b) => {
+                // weight function last places weight more than first: application/td+json > application/json > text/html
+                const aWeight =
+                    ["text/html", "application/json", "application/td+json"].findIndex((value) => value === a) ?? 0;
+                const bWeight =
+                    ["text/html", "application/json", "application/td+json"].findIndex((value) => value === b) ?? 0;
+
+                return bWeight - aWeight;
+            });
 
         if (filteredAcceptValues.length > 0) {
             const contentType = filteredAcceptValues[0];
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -573,10 +573,12 @@ export default class HttpServer implements ProtocolServer {
             .filter((acceptValue) => contentSerdes.isSupported(acceptValue))
             .sort((a, b) => {
                 // weight function last places weight more than first: application/td+json > application/json > text/html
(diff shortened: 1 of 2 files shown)


relu91:
Pushed a small test. I noticed also that we are ignoring the [`quality` factor](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values). I think in the future we can order the Accept list accordingly but then choose our own preference list? It seems that the algorithm for choosing the content-type is not [well-defined](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#server-driven_content_negotiation)... I have to check if in discovery we state something about this.

packages/binding-http/src/http-server.ts
```diff
@@ -570,7 +570,16 @@ export default class HttpServer implements ProtocolServer {
...
                 return acceptValue;
             })
-            .filter((acceptValue) => contentSerdes.isSupported(acceptValue));
+            .filter((acceptValue) => contentSerdes.isSupported(acceptValue))
+            .sort((a, b) => {
+                // weight function last places weight more than first: application/td+json > application/json > text/html
+                const aWeight =
+                    ["text/html", "application/json", "application/td+json"].findIndex((value) => value === a) ?? 0;
+                const bWeight =
+                    ["text/html", "application/json", "application/td+json"].findIndex((value) => value === b) ?? 0;
```
danielpeintner, line 579: I wonder whether the fallback should **not** be `0` but maybe `-1`

Let's assume some fancy value comes in like `application/bla` besides `text/html`. Both would result in index 0 while I think we should favor `text/html` ...

danielpeintner (commented):
I just have a question w.r.t. sorting order

packages/binding-http/src/http-server.ts
relu91: I forgot that `findIndex` returns already -1 if the element is not found (not undefined), therefore `??` is pointless. I removed it and added a relevant test.

packages/binding-http/test/http-server-test.ts
```diff
@@ -741,6 +741,23 @@ class HttpServerTest {
...
+                inputHeaders: {
+                    Accept: "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
+                },
+                // We should favor application/td+json over text/html
+                expected: "application/td+json",
+                expectedResponseCode: 200,
+            },
+            {
+                inputHeaders: {
+                    Accept: "image/svg+xml,text/html,",
```
danielpeintner, line 755:
```suggestion
                    Accept: "image/svg+xml,text/html",
```

danielpeintner (approved):
LGTM.

I just added one minor fix... not really needed though

PR #1003, opened 2023-05-18
fix(examples/counter): correctly display images

relu91:
I noticed that the current web counter-example does not display the images "correctly". I'm not sure if this was something by design or not, but I think it would be better to not display images as simple text like the following:
image

With this PR the example looks like this:
image

### DIFF
--- a/examples/browser/counter.js
+++ b/examples/browser/counter.js
@@ -48,7 +48,18 @@ function showInteractions(thing) {
                 thing
                     .readProperty(property)
                     .then(async (res) => {
-                        ddItem.textContent = await res.value();
+                        switch (res.form?.contentType) {
+                            case "image/svg+xml":
+                                ddItem.innerHTML = await res.value();
+                                break;
+                            case "image/png;base64":
+                                const img = (await res.arrayBuffer()).toString("base64");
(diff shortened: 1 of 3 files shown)


danielpeintner:
REMINDER: Once it is merged we also need to fix the online version at http://plugfest.thingweb.io/examples/counter.html

danielpeintner:
> REMINDER: Once it is merged we also need to fix the online version at http://plugfest.thingweb.io/examples/counter.html

updated node-wot and the counter HTML/JS on http://plugfest.thingweb.io with teh tip of master

danielpeintner:
> **Note**: I noticed that right now the event subscription is broken because we are generating `hrefs` for event with URI variables like the following: `http://172.23.16.1:8080/counter/events/change{?step}`. I think the exposer is doing his job because we have declared a global UriVariables object, but I am not sure how to override it inside the event. It seems to be that the client is unable to subscribe to forms that contain URI variables templates, it might be a bug. Nonetheless, I don't think it makes sense to have a step parameter for the event `count`.

Shouldn't we create an issue? I did create https://github.com/eclipse-thingweb/node-wot/issues/1006

Issue #973, opened 2023-05-05
Memory Leak in Events


egekorkan:
The events need to be better handled since the hosted Things are throwing memory leak warnings such as:

> MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 end listeners added to [Readable]. Use emitter.setMaxListeners() to increase limit

relu91:
See the comment linked above for some initial investigation result.

danielpeintner:
The comment is not really linked, just the issue but I think you mean https://github.com/eclipse-thingweb/node-wot/pull/1008#issuecomment-1580334956

relu91:
Fixed in #1008 re-open if the problem arises again.

egekorkan:
The issue seems to still happen as shown in today's call.

Issue #230, opened 2020-05-26
in subscribeEvent calling resolv for first(every in fact) event prevent to catch later error sent by reject.


franckOL:
https://github.com/eclipse/thingweb.node-wot/blob/7bcc1e651757c4f1f8e6814d64b30a2711add6f0/packages/core/src/consumed-thing.ts#L395

This a correct implementatin regarding the specification https://www.w3.org/TR/wot-scripting-api/#the-subscribeevent-method.

But how to deal with error happens during polling that need to resubscribe.

I try to remove resolv, and it works, of course i have now a perpetual unresolved promise (but catching polling error that happens)

danielpeintner:
May I ask which binding do you use? CoAP?

franckOL:
I use http-binding

danielpeintner:
FYI: Related HTTP unsubscribe issue https://github.com/eclipse/thingweb.node-wot/pull/233.

I wonder whether a `else` branch is needed [here](https://github.com/eclipse/thingweb.node-wot/blob/7652a56c8d0dbb06b7c558bd52d2c05f47ebb534/packages/binding-http/src/http-client.ts#L186) calling 

`complete && complete()`

after the subscription is removed?
@egekorkan ?

relu91:
I agree that after a successful unsubscribe the complete function should be called. As I said in https://github.com/eclipse/thingweb.node-wot/pull/233#issuecomment-635228354, we should also abort the pending HTTP GET, which may cause a throw [here](https://github.com/eclipse/thingweb.node-wot/blob/7652a56c8d0dbb06b7c558bd52d2c05f47ebb534/packages/binding-http/src/http-client.ts#L173). Consequently, inside the catch clause, we should handle the exception that indicates the cancellation and call the complete method there too. 

Like so we avoid pending resources which may be freed only after one hour: `{ timeout: 60 * 60 * 1000 }`.

danielpeintner:
> we should also abort the pending HTTP GET, which may cause a throw [here](https://github.com/eclipse/thingweb.node-wot/blob/7652a56c8d0dbb06b7c558bd52d2c05f47ebb534/packages/binding-http/src/http-client.ts#L173). Consequently, inside the catch clause, we should handle the exception that indicates the cancellation and call the complete method there too.

If the line you cited throws we fall into the catch clause where we do already `error` and `complete`.
@relu91 not sure If I misunderstood your comment. Is there still anything missing there also?

[...]

PR #1008, opened 2023-05-19
Fix event subscription and improve Counter example

### LINKED ISSUE #1005: Events does not work with `URIVariables`
> **Note**: I noticed that right now the event subscription is broken because we are generating `hrefs` for event with URI variables like the following: `http://172.23.16.1:8080/counter/events/change{?step}`. I think the exposer is doing his job because we have declared a global UriVariables object, but I am not sure how to override it inside the event. It seems to be that the client is unable to subscribe to forms that contain URI variables templates, it might be a bug. Nonetheless, I don't think it makes sense to have a step parameter for the event `count`.

_Originally posted by @relu91 in https://github.com/eclipse-thingweb/node-wot/issues/1003#issuecomment-1553257383_

relu91:
This PR should eventually fix #1005 and might solve #973 as well. Moreover, I wanted to improve a little bit the UI in the counter-example to give users small feedback when events occur. Everything is still draft because I need to clean up the code, the commit log, and typescript typings. Other than that it should work and people can try it.


examples/browser/counter.html
```diff
@@ -27,7 +44,7 @@ <h1>Counter Client Example in the Browser</h1>
                     readonly="readonly"
                     id="td_addr"
                     type="url"
-                    value="http://plugfest.thingweb.io:8083/counter"
+                    value="http://localhost:8080/counter"
```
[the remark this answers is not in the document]
relu91: Well spotted, this is actually a change that I forgot to revert. I think we can keep `http://plugfest.thingweb.io:8083/counter` but in the future, we can give the user the ability to configure the URL by writing in the textbox.

relu91:
I reviewed the updated code and I noticed that I left a couple of `any`s around. This is shameful 😸 . I'll fix asap. Btw I was reading old issues and I think this is relevant to https://github.com/eclipse-thingweb/node-wot/issues/230. I think adding the abortController fixes the points that we were discussing. I just want to be sure if aborting the request in the `close` method is enough.

relu91:
About the `any` issue I find out that there were a bug in Typescript `4.4.3` until around `4.7.x`. I'll try to update the typescript version. Any reason to not do that?

examples/browser/counter.js
```diff
@@ -124,12 +126,16 @@ function showInteractions(thing) {
                     eventSubscriptions[evnt] = true;
                     thing
                         .subscribeEvent(evnt, async function (data) {
-                            console.log('Event "' + evnt + '"');
-                            updateProperties();
```
danielpeintner, line 128: since you remove the `updateProperties()` call the event triggers the event information but does not update the `count` values on the left in the UI. Is this desired?

danielpeintner:
> About the `any` issue I find out that there were a bug in Typescript `4.4.3` until around `4.7.x`. I'll try to update the typescript version. Any reason to not do that?

I don't see any reason not to update 👍

danielpeintner (commented):
I think it looks good but I still have the questions below (as raised before)

- for events: Once subscribed again it remembers the "old" payload while the event count is set back to zero: I think the payload should be also nulled. What do you think?
- How can we add a test to avoid regressions in the future?

relu91:
> for events: Once subscribed again it remembers the "old" payload while the event count is set back to zero: I think the payload should be also nulled. What do you think?

I should revert back to an empty string [here](https://github.com/eclipse-thingweb/node-wot/pull/1008/files#diff-44934db5222f0b75ec18572d443b189f1bc79a126bdb17f6058cc2e04e7350fbR154). Is that ok?

More info on #973. With the current changes the memory leak is fixed for the first subscription but the problem arise again if you follow these steps:
- subscribe
- generate a couple of events
- unsubscribe
- subscribe again
- generate another series of events (more than 11)
- **Memory leak detected**

I think this is due to the fact that [here](https://github.com/eclipse-thingweb/node-wot/blob/b0b689f8c57ea08b615129cd18fcf1748af65fdb/packages/binding-http/src/http-server.ts#L985) the server response is reused and the client keep it open. I have to investigate more.

relu91:
> How can we add a test to avoid regressions in the future?

It would be nice to have something similar to what `respec` has. They are using karma to run a couple of test suites directly on the browsers.

relu91:
https://github.com/eclipse-thingweb/node-wot/pull/1008/commits/26598bee954bd5932c64935477af4a334c5e7099 fix #973

egekorkan:
> > How can we add a test to avoid regressions in the future?
> 
> It would be nice to have something similar to what `respec` has. They are using [karma](https://karma-runner.github.io/6.4/index.html) to run a couple of test suites directly on the browsers.

In playground, we use [playwright](https://playwright.dev/) to test the interface: https://github.com/thingweb/thingweb-playground/blob/master/packages/web/test.js . It works pretty well and has a quite simplistic way to program the tests. Nothing against karma, just one alternative.

examples/browser/counter.js
danielpeintner: I understand that technically it is not needed but in the UI the counter value is not updated. Hence the value in the UI shows the correct value if the event is fired while on the left the "same" counter value still shows the old value...

danielpeintner:
> I should revert back to an empty string [here](https://github.com/eclipse-thingweb/node-wot/pull/1008/files#diff-44934db5222f0b75ec18572d443b189f1bc79a126bdb17f6058cc2e04e7350fbR154). Is that ok?

👍

examples/browser/counter.js
danielpeintner: I think the whole point of observing the `change` event is that I get notified if _someone else_ triggers a `count` change.
This can be that the UI is open in several browser windows or by Postman actions etc..

relu91: Right, let's do it like this. I'll fix it in this PR, but I'll also open an issue to propose this change in the browser example: observe properties and change the event from the current logic to an even fired every time that the count is divisible by 2. I guess it would make a better newbie example.

danielpeintner (approved):
LGTM

PR #1010, opened 2023-05-30
Use @node-oauth/express-oauth-server instead of express-oauth-server

relu91:
As we know the `express-oauth-server` gave us some headaches because it was using a vulnerable version of loadsh (or some other library). Sadly, the development of such a tool was discontinued and we were left with a broken dependency that gave us bad audit scores. Thanks to the work in [node-oauth](https://github.com/node-oauth), we now have a new version of the server with 0 vulnerabilities. This PR upgrade our dependency list with the new service. We are now down to 8 vulnerabilities but 0 critical (I also tried to solve netconf vulnerability but it seems that the library it is not very active for a long time). 

See also https://github.com/node-oauth/express-oauth-server/issues/5


danielpeintner (approved):
LGTM

Great to see the critical vulnerabilities be gone 👍

Issue #1012, opened 2023-06-09
OPC-UA example from README not working


danielpeintner:
I tried to run the OPC-UA example `demo-opcua1.ts` from [README](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-opcua/README.md) (see [gist](https://gist.github.com/danielpeintner/e6c0f3723dce910f2dfcb8128f10b964) for full code).

When trying to run it I get the following error

```
Cannot figure out field category for  { name: 'value', fieldType: 'Variant', defaultValue: null }
...\node_modules\node-opcua-factory\dist\builtin_types.js:262
        throw new Error("Cannot find schema for simple type " + name);
```

The issue seems to come from [node-opcua](https://www.npmjs.com/package/node-opcua) and not from node-wot code.
@erossignon do you have any advice?

danielpeintner:
Interesting: IF I run the sample `demo-opcua1` from https://github.com/eclipse-thingweb/node-wot/blob/master/packages/examples/src/bindings/opcua/demo-opcua1.ts I get a different error...

```
D:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\examples>node dist\bindings\opcua\demo-opcua1
D:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-opcua\dist\opcua-protocol-client.js:117
                    throw new Error("Cannot connected to endpoint " + endpoint + "\nmsg = " + err.message);
                          ^

Error: Cannot connected to endpoint opc.tcp://opcuademo.sterfive.com:26543
msg = The connection may have been rejected by server,
Err = (ERROR in waiting for data on socket ( timeout was = 60000 ms) undefined)
    at OPCUAProtocolClient.<anonymous> (D:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-opcua\dist\opcua-protocol-client.js:117:27)
    at Generator.throw (<anonymous>)
    at rejected (D:\Projects\WoT\thingweb\thingweb.node-wot-danielpeintner\packages\binding-opcua\dist\opcua-protocol-client.js:6:65)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v18.14.2
```

As if the server is not reachable even though the `endpointUrl` `"opc.tcp://opcuademo.sterfive.com:26543"` is the same?

erossignon:
opcuademo.sterfive.com was down and should be back online.

[...]

PR #1002, opened 2023-05-18
docs: reflect github repo renaming

danielpeintner:
eclipse/thingweb.node-wot to eclipse-thingweb/node-wot

PR #1019, opened 2023-06-12
fix: OPCUA peerDepencencies issue by adding missing package

erossignon:
* this commit explicitly add packages that appears in the peerDependencies of node-opcua-pubsub-client and node-opcua-json to the package.json of binding-opcua. This is to ensure that the expected versions of those packages will be loaded by npm and prevent conflict.
  * this commit also add specifically rxjs@5.5.11

fixes https://github.com/eclipse-thingweb/node-wot/issues/1012


relu91 (commented):
Are we seeing 0 changes in the root package-lock because those packages are already loaded? or is this a mistake?

danielpeintner:
@erossignon can you merge in updates from master.
https://github.com/eclipse-thingweb/node-wot/pull/1021 should resolve the issues w.r.t. consistency across submodels

relu91:
> Are we seeing 0 changes in the root package-lock because those packages are already loaded? or is this a mistake?

read #1002, see my other comment: https://github.com/eclipse-thingweb/node-wot/issues/1012#issuecomment-1588854262

erossignon:
> Are we seeing 0 changes in the root package-lock because those packages are already loaded? or is this a mistake?

Yes, correct, the top level package.json already includes all necessary packages that are in the peerDepencencies , that is why the problem is not surfacing when the example is run in the context of this module.

Adding the missing modules in binding-opcua makes it work for when binding-opcua is installed standalone.

erossignon:
@danielpeintner , this is now ready to merge.

danielpeintner:
> > Are we seeing 0 changes in the root package-lock because those packages are already loaded? or is this a mistake?
> 
> Yes, correct, the top level package.json already includes all necessary packages that are in the peerDepencencies , that is why the problem is not surfacing when the example is run in the context of this module.
> 
> Adding the missing modules in binding-opcua makes it work for when binding-opcua is installed standalone.

@relu91 are you ok with it?

relu91 (approved):
I don't want to block the PR so let's move on. But I want to say that (at least to me) it sounds weird to explicitly add the peerDeps in a library package just to simplify the installation for others. It kind of defeats the purpose of the peerDeps. This is also further supported by the fact that in `pnpm ` the problem does not show up (I heard that newer versions of `npm` are better at handling peerDeps). Anyhow, let's merge this.

Issue #1020, opened 2023-06-12
Remove firestore binding


danielpeintner:
The `binding-firestore` was commented out for quite some time now, see https://github.com/eclipse-thingweb/node-wot/blob/ba3f9326a92d21730827a00b42ff3bee16123c8d/tsconfig.json#L32

I think it is time to remove the binding as a whole. Anyhow, we should add a link to the "last" version so that people can pick up the work again.

Issue #1016, opened 2023-06-09
Fix version inconsistency introduced by PR 1008


danielpeintner:
see https://github.com/eclipse-thingweb/node-wot/pull/1008
or https://github.com/eclipse-thingweb/node-wot/actions/runs/5221226153/jobs/9425236133

PR #1021, opened 2023-06-12
chore: remove binding-firestore

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1020

fixes https://github.com/eclipse-thingweb/node-wot/issues/1016 also


relu91 (approved):
LGTM! thanks :)

PR #1024, opened 2023-06-15
tests/binding-opcua: fix guessing of endpoint to localhost

egekorkan:
Otherwise, the tests were failing in macos since it tries to connect to `opc.tcp://MACPYJ44WVW33:7890` which does not resolve. Ideally [this](https://github.com/node-opcua/node-opcua/blob/master/packages/node-opcua-server/source/base_server.ts#L433) should be changed at node-opcua.

@erossignon

### DIFF
--- a/packages/binding-opcua/test/client-test.ts
+++ b/packages/binding-opcua/test/client-test.ts
@@ -22,18 +22,15 @@ import { VariableIds, OPCUAServer } from "node-opcua";
 import { OPCUAProtocolClient, OPCUAForm, OPCUAFormInvoke } from "../src/opcua-protocol-client";
 import { OpcuaJSONCodec, schemaDataValue } from "../src/codec";
 import { startServer } from "./fixture/basic-opcua-server";
-
+const endpoint = "opc.tcp://localhost:7890";
 const { debug } = createLoggers("binding-opcua", "opcua-protocol-client");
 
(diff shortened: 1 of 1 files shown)


erossignon:
> Otherwise, the tests were failing in macos since it tries to connect to `opc.tcp://MACPYJ44WVW33:7890` which does not resolve. Ideally [this](https://github.com/node-opcua/node-opcua/blob/master/packages/node-opcua-server/source/base_server.ts#L433) should be changed at node-opcua.
> 
> @erossignon

Hi, I would not change this code.

It is always better to let the server provide the endpoint rather than crafting it manually.

MACPYJ44WVW33 is supposed to be the machine name on which you are running  and should always resolve, unless you have a hostname configuration mismatch on your computer.

Let's fix the root cause.

egekorkan:
Indeed it should resolve, I will have a look

relu91 (approved):
good to go from my side. I don't have a strong opinion to keep/remove the debug statement. I let @danielpeintner decide to merge as it is or add the debug directive.

danielpeintner:
@egekorkan have you been able to identify the issue (as mentioned in https://github.com/eclipse-thingweb/node-wot/pull/1024#issuecomment-1592859907)

I think it is a "Siemens" network/machine/setup issue..

PR #1025, opened 2023-06-16
docs(README): add a warning about node version

relu91:


relu91:
Oops, seems that editing the README online wasn't a great idea :) fixing soon.

README.md
```diff
@@ -99,6 +99,9 @@ cs.addCodec(new MyCodec("application/myType"));
 
 ### To use with Node.js
 
+> **Warning**: We are currently in the process to drop support for NodeJS versions 14 and 16. 
+> Node 20 is not yet supported (https://github.com/eclipse-thingweb/node-wot/issues/1004)
+
```
JKRhb, line 104: I think the problem is actually only the trailing whitespace here :)

```suggestion
> **Warning**: We are currently in the process to drop support for NodeJS versions 14 and 16.
> Node 20 is not yet supported (https://github.com/eclipse-thingweb/node-wot/issues/1004)
```

README.md
```diff
@@ -99,6 +99,8 @@ cs.addCodec(new MyCodec("application/myType"));
 
 ### To use with Node.js
 
+> **Warning**: We are currently in the process to drop support for NodeJS versions 14 and 16.
+> Node 20 is not yet supported (https://github.com/eclipse-thingweb/node-wot/issues/1004)
 All systems require:
```
JKRhb, line 104: Sorry, seems as if there is also a newline required here 😅

```suggestion

All systems require:
```

README.md
```diff
@@ -99,6 +99,9 @@ cs.addCodec(new MyCodec("application/myType"));
 
 ### To use with Node.js
 
+> **Warning**: We are currently in the process to drop support for NodeJS versions 14 and 16.
```
danielpeintner, line 102: I think one can still run on Node.js 14 but we do no longer run the CI tests for it
https://github.com/eclipse-thingweb/node-wot/blob/19cc4624a51ded3662cb8ac165c19d836b346ad1/.github/workflows/ci-full.yaml#L11

Hence, I think the line above needs to be rephrased or removed

danielpeintner (changes requested):

README.md
relu91: Something like: 
```suggestion
> **Warning**: We are currently in the process to drop support for NodeJS versions below 18. 
```

danielpeintner: Two comments:
* I think we should align the terminology. We use "NodeJS" and "Node.js" interchangeable ... I would got for "Node.js". This inconsistency was there before and maybe we can fix it with this PR
* w.r.t. to the *new* warning. Maybe we can say "We are currently no longer running tests with Node.js version 14". I think that's the status quo.  Sooner or later we might want to lift the minimum requirement. Currently we say "[NodeJS](https://nodejs.org/) version 14+"

README.md
```diff
@@ -99,6 +99,9 @@ cs.addCodec(new MyCodec("application/myType"));
...
+> Node 20 is not yet supported (https://github.com/eclipse-thingweb/node-wot/issues/1004)
```
danielpeintner, line 103:
```suggestion
> Node.js 20 is not yet supported (https://github.com/eclipse-thingweb/node-wot/issues/1004)
```

README.md
relu91: Maybe I have a too strict opinion, but if we are not running tests on a Node.js version it means that we are not supporting it. In other words, it might still work on older platforms but we can't be sure because we don't have green tests for it. I would like to clearly give the impression that we are not supporting older versions, but I am open to other opinions too. Do you think it is good to showcase possible support for older Node.js versions?

relu91:
```suggestion
> **Warning**: We no longer support Node.js version 14 and lower.
```

Ok, then the suggested change works for you?

JKRhb: Maybe we could rephrase it to "*actively* support" to imply that earlier versions may still be supported, we're just not testing it?

relu91: good point:

```suggestion
> **Warning**: We no longer actively support Node.js version 14 and lower.
```

danielpeintner: The proposed change by @JKRhb looks even better! This would resolve all my issues and we just need to resolve the merge conflict..

danielpeintner (approved):

PR #1026, opened 2023-06-17
fix(coap-server): correctly expose additional mediatypes

JKRhb:
As pointed out by @egekorkan on Discord, the testthing example currently does not expose forms with CBOR `contentType`s when using a CoAP server. Trying to debug the problem, I noticed that this is caused by the helper functions `updatePropertyFormWithTemplate`, `updateActionFormWithTemplate`, and `updateEventFormWithTemplate`.

After reading through their source code, I am not actually sure what these functions are supposed to do, therefore I simply removed their invocation from the CoAP server for the time being which fixed the problem. If they are necessary to integrate into a server implementation, maybe you could help me out with using them correctly instead, though. Thanks :)

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -192,8 +192,6 @@ export default class CoapServer implements ProtocolServer {
                 opValues
             );
 
-            ProtocolHelpers.updatePropertyFormWithTemplate(form, property);
-
             property.forms.push(form);
             this.logHrefAssignment(port, href, "Property", propertyName);
         }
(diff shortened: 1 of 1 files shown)


relu91:
If I recall correctly those set of methods are meant to update the form generated by the binding, trying to fulfill what the user set in the `PartialTD` passed in the `produce` method. My guess, is that using them in two different bindings causes some possible collisions 🤔 . However, the methods are poorly implemented and I think we should review them in the future. I propose merging this PR and keeping track of this issue.

relu91 (approved):

relu91:
I talked with @danielpeintner he is ok with merging PRs if we get two approvals while he is away. Since this PR is coming from @JKRhb a soon-to-be committer, I'm merging it for the reasons mentioned in the comment above.

PR #1027, opened 2023-06-18
feat(binding-http): add http server middleware

Luca8991:
Add a middleware for the http server to handle raw http requests.
An example on how to use the middleware can be found in the tests: https://github.com/Luca8991/node-wot/blob/master/packages/binding-http/test/http-server-test.ts#L53-L101

### DIFF
--- a/packages/binding-http/README.md
+++ b/packages/binding-http/README.md
@@ -174,15 +174,15 @@ The protocol binding can be configured using his constructor or trough servient 
 
 ```ts
 {
-    port?: number;                  // TCP Port to listen on
-    address?: string;               // IP address or hostname of local interface to bind to
-    proxy?: HttpProxyConfig;        // proxy configuration
-    allowSelfSigned?: boolean;      // Accept self signed certificates
-    serverKey?: string;             // HTTPs server secret key file
-    serverCert?: string;            // HTTPs server certificate file
-    security?: TD.SecurityScheme;   // Security scheme of the server
-    baseUri?: string                // A Base URI to be used in the TD in cases where the client will access a different URL than the actual machine serving the thing.  [See Using BaseUri below]
-    middleware?: HttpMiddleware;    // the instance of HttpMiddleware. See [Adding a middleware] section below.
+    port?: number;                           // TCP Port to listen on
+    address?: string;                        // IP address or hostname of local interface to bind to
+    proxy?: HttpProxyConfig;                 // proxy configuration
+    allowSelfSigned?: boolean;               // Accept self signed certificates
+    serverKey?: string;                      // HTTPs server secret key file
+    serverCert?: string;                     // HTTPs server certificate file
+    security?: TD.SecurityScheme;            // Security scheme of the server
+    baseUri?: string                         // A Base URI to be used in the TD in cases where the client will access a different URL than the actual machine serving the thing.  [See Using BaseUri below]
+    middleware?: MiddlewareRequestHandler;   // the MiddlewareRequestHandler function. See [Adding a middleware] section below.
 }
 ```
 
@@ -333,7 +333,6 @@ const middleware = async (req, res, next) => {
 };
 
 const httpServer = new HttpServer({
-    port,
     middleware,
 });
 
--- a/packages/binding-http/README.md
+++ b/packages/binding-http/README.md
@@ -182,6 +182,7 @@ The protocol binding can be configured using his constructor or trough servient 
     serverCert?: string;            // HTTPs server certificate file
     security?: TD.SecurityScheme;   // Security scheme of the server
     baseUri?: string                // A Base URI to be used in the TD in cases where the client will access a different URL than the actual machine serving the thing.  [See Using BaseUri below]
+    middleware?: HttpMiddleware;    // the instance of HttpMiddleware. See [Adding a middleware] section below.
 }
 ```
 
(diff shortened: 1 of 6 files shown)


packages/binding-http/src/http-server-middleware.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    constructor(handler: MiddlewareRequestHandler) {
+        this.handler = handler;
+    }
+
+    public handleRequest(req: http.IncomingMessage, res: http.ServerResponse, next: () => void): Promise<void> {
+        debug("Hit middleware");
+        return this.handler(req, res, next);
+    }
+}
```
relu91, line 83: Can't we just get rid of this class and use directly the `MiddlewareRequestHandler`. I don't see the benefit of having this container class other than having a debug message. 🤔

relu91 (changes requested):
Suggesting a small refactor.

packages/binding-http/src/http-server-middleware.ts
relu91: Maybe, it would be useful to have a class that takes as input a list of `MiddlewareRequestHandler`s and configure the next function accordingly? (just got this requirement from my local refactoring, but feel free to add it later)

Luca8991: > Can't we just get rid of this class and use directly the MiddlewareRequestHandler. I don't see the benefit of having this container class other than having a debug message.

The class was meant more as a base class that can a developer can extend to handle more complex logic, like keeping a "state" between on request and the other, or any other logic that comes to mind.

But I agree the `MiddlewareRequestHandler` function is what gets executed inside the HttpServer, so we can get rid of the class. Just let me know how to proceed.

> Maybe, it would be useful to have a class that takes as input a list of `MiddlewareRequestHandler`s and configure the next function accordingly? (just got this requirement from my local refactoring, but feel free to add it later)

How would you configure the `next` function based on the `MiddlewareRequestHandler`? I didn't really get it :thinking:

JKRhb: > Maybe, it would be useful to have a class that takes as input a list of `MiddlewareRequestHandler`s and configure the next function accordingly? (just got this requirement from my local refactoring, but feel free to add it later)

Maybe the type of the `middleware` parameter in the `HttpServer`'s constructor could also be changed to `HttpMiddleware | HttpMiddleware[]` to cover that? Although things might indeed become a bit complicated in the array case when it comes to the `next` function.

Maybe an alternative could be to let users pass in functions of type

```ts
export type AlternativeMiddlewareRequestHandler = (
    req: http.IncomingMessage,
    res: http.ServerResponse,
    abort: () => void,
) => Promise<void>;
```

and then the server does the chaining internally. In this case, I replaced `next()` with an `abort()` function which would allow users to stop the processing – the next middleware would be called by the server instead if no `abort()` happened before.

Do you think this could work? Or would having a `next()` function instead give users more control?

relu91: I was thinking of something like this:
```
let middlewares: MiddlewareRequestHandler[] // initialized somewhere
const first = middlewares.shift()
const entries = middleware.entries()

first(req,res, () => entries.next().value[1]()) // needs to check also if we are done but removed for giving the idea
```

this way the middleware can control whether to pass to the next function or not, without introducing the abort. Going back to the PR itself @Luca8991 I would remove HTTPMiddleware class and discuss further improvements in a future PR.

Luca8991: @relu91 I've pushed the commit that removes the `HTTPMiddleware` class

relu91 (approved):
Thank you really much for providing the requested changes! Ready to go.

packages/binding-http/src/http-server-middleware.ts
```diff
@@ -0,0 +1,63 @@
...
+ *         res.statusCode = 400;
+ *         res.end("Bad Request");
+ *         return;
+ *     }
+ *     // Pass all other requests to the WoT Servient
+ *     next();
+ * };
+
+ * const httpServer = new HttpServer({
+ *     port,
```
danielpeintner, line 45: The port is neither needed nor defined in this context.
Remove it?

packages/binding-http/README.md
```diff
@@ -301,10 +302,47 @@ The exposed thing on the internal server will product form URLs such as:
...
+        res.statusCode = 400;
+        res.end("Bad Request");
+        return;
+    }
+    // Pass all other requests to the WoT Servient
+    next();
+};
+
+const httpServer = new HttpServer({
+    port,
```
danielpeintner, line 336: The port is neither needed nor defined in this context.
Remove it?

packages/binding-http/README.md
```diff
@@ -182,6 +182,7 @@ The protocol binding can be configured using his constructor or trough servient
     serverCert?: string;            // HTTPs server certificate file
     security?: TD.SecurityScheme;   // Security scheme of the server
     baseUri?: string                // A Base URI to be used in the TD in cases where the client will access a different URL than the actual machine serving the thing.  [See Using BaseUri below]
+    middleware?: HttpMiddleware;    // the instance of HttpMiddleware. See [Adding a middleware] section below.
```
danielpeintner, line 185: I think HttpMiddleware should become MiddlewareRequestHandler

danielpeintner (changes requested):
Looks good in general, Thanks 👍.
I have just some minor requests.

danielpeintner (approved):
LGTM 👍

Issue #873, opened 2022-11-18
OAuth2 example does not work


ms499:
With @egekorkan I have discussed this issue. I tried to run the example oauth code provided. I have not changed any code and simply followed the instruction mentioned in the Readme.md file but nothing was getting printed when I ran the consumer.js file. Then I enabled the debug level so that the exposer.js can print the log and I saw the following error:
'node-wot:cli:cli:error WoT-Servient cannot start. Error: Servient does not support thing security schemes. Current scheme supported: OAuth secCandidate'

https://github.com/eclipse/thingweb.node-wot/tree/master/examples/security/oauth

danielpeintner:
Error message comes from `fillSecurityScheme()`

https://github.com/eclipse/thingweb.node-wot/blob/557adb69993d3f347a70202401b7b94edbe0fb96/packages/binding-http/src/http-server.ts#L481-L512

relu91:
As mentioned above the source of this issue is coming from the `ExposedThingInit`, but that's not the whole story. The problem is that `HttpsServer` currently supports only one scheme at a time. Therefore, the CLI cannot expose its base Servient thing because it wants it to use the no_sec security scheme.

relu91:
@danielpeintner @egekorkan @sebastiankb  should `HTTPServer` support multiple security schemes at once?

danielpeintner:
Mhh, good question. I can see use-cases where an HTTP server might be allowing multiple security schemes but I do not know the consequences w.r.t. implementation 🤔

egekorkan:
I do not have a very good answer but an answer at least:
- It should be able to it can support the features (combo sec) of the TD. We do not have to support all features though.
- Support on the client side is needed in order to consume TDs such as https://github.com/w3c/wot-testing/blob/main/data/input_2022/TD/philips-hue/TDs/tum-light1.td.jsonld
- A use case I can think of is related to the TD example above. That one exists because we have a simple HTTPS proxy that changes only the domain name of the local device with a public domain name (and adds basic auth). Since hue devices use API Key in the URL as a security scheme, those URIs are still visible in the public TD. If someone wants to do exactly this where the proxy is done with node-wot, that proxy's HTTP server needs to support two security schemes. This is an AND relationship (allOf in comboSec). This is bad practice all over but a possibility.
- A more likely use case if I want to have a TD where local IP address forms use nosec and the public IP address forms use basic auth. This is an OR relationship (oneOf in combo sec). 

[...]

PR #1030, opened 2023-06-27
refactor(binding-http/server): introduce routes for each TD endpoint

relu91:
Hi all, the goal of this PR is to declutter our HttpServer implementation. Over the years, it got a lot of new functionalities but it never had a second look at its software architecture. My intention was to maintain as much as possible intact the current code base, but also move the logic of handling requests to dedicated functions. Furthermore, as mentioned in the note below this PR introduce a routing library (taken from the well-known library Fastify). This simplified a lot the refactoring with the additional benefit to provide more flexibility in the future. The library is compact, small, and well-supported, so it shouldn't be a problem in the future. 

There are plenty of other changes that I would like to add, but I didn't want to overload the changes shown here. I also spotted some bugs/discrepancies in the handlers of the different TD affordances, which proves again the benefit of these refactors. I'll fix the problems in the following PRs. 

**Note**: that we are now use find-my-way for routing, which is a bit more complex than the previous approach. However, it is also more flexible and in the future, we can easily add more routes or let users configure them. Another caveat is the trailing slashes in the routes now map to a totally different path. For example, `/things` and `/things/` are now different routes. In the future, we might review this behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5087,6 +5087,11 @@
                 "node": ">=0.4.0"
             }
         },
+        "node_modules/fast-decode-uri-component": {
+            "version": "1.0.1",
+            "resolved": "https://registry.npmjs.org/fast-decode-uri-component/-/fast-decode-uri-component-1.0.1.tgz",
+            "integrity": "sha512-WKgKWg5eUxvRZGwW8FvfbaH7AXSh2cL+3j5fMGzUMCxWBJ3dV3a7Wz8y2f/uQ0e3B6WmodD3oS54jTQ9HVTIIg=="
+        },
         "node_modules/fast-deep-equal": {
             "version": "3.1.3",
             "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
@@ -5119,6 +5124,14 @@
             "resolved": "https://registry.npmjs.org/fast-levenshtein/-/fast-levenshtein-2.0.6.tgz",
             "integrity": "sha512-DCXu6Ifhqcks7TZKY3Hxp3y6qphY5SJZmrWMDrKcERSOXWQdMhU9Ig/PYrzyw/ul9jOIyh0N4M0tbC5hodg8dw==",
             "dev": true
+        },
+        "node_modules/fast-querystring": {
+            "version": "1.1.2",
+            "resolved": "https://registry.npmjs.org/fast-querystring/-/fast-querystring-1.1.2.tgz",
+            "integrity": "sha512-g6KuKWmFXc0fID8WWH0jit4g0AGBoJhCkJMb1RmbsSEUNvQ+ZC8D6CUZ+GtF8nMzSPXnhiePyyqqipzNNEnHjg==",
+            "dependencies": {
+                "fast-decode-uri-component": "^1.0.1"
+            }
         },
         "node_modules/fast-safe-stringify": {
             "version": "2.1.1",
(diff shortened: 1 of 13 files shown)


relu91:
Took the opportunity to also remove the deprecated `url.parse` call. :)

relu91:
Thank you @danielpeintner for reviewing and sorry for answering late. 

> I think restructering and decluttering the code is good 👍 but also dangerous 🤔

Totally agree. What I tried to do is to keep the code as much as intact as it was and verifying the counter example was working both in "headless" mode and in the browser. The tests are also passing, but as you know the coverage is not that great. Something to improve for sure in the future. I believe that this refactor would help introduce new tests and also new features. 

> You have comments about credentials checking saying
// TODO: refactor this part to move into a common place
Do you plan to move it into a common place in this PR? I think we should...

Yes, I plan to do so, but I didn't include the changes in this PR because, doing it right, would result in additional refactoring changes that I wanted to avoid cause it touches something delicate such as authentication. Plus this change would impact also #873. If you have a strong opinion I would defer it to another PR, as you said this is already a lot to digest. 


> What concerns me a bit is that /things and /things/ are now different routes and I wonder whether this might break existing apps when updating?

The router has a simple flag to change the current behavior. I was not sure that having /things and /things/ point to the same resource was correct (see also https://github.com/vaimee/zion/issues/7 and [Ben's comment](https://github.com/vaimee/zion/issues/7#issuecomment-1201093041). But I guess to make the number of breaking changes low I can allow the flag. Sending a commit soon.

JKRhb:
Thank you for this PR, @relu91! :) In my opinion, the changes look pretty good, although while going through the code I found a lot of potential for further refactoring – maybe I'll keep the comments I prepared so far for myself for now, though, in order not to make things overly complicated for this PR and add them later or to a separate PR. If everything (or the things we can currently test for) is working as before, then I think we could go forward with this PR.

Regarding the trailing slashes: From my understanding, `/things` and `/things/` should in fact be treated differently (I think the trailing slash introduces an empty path segment), but I am not entirely sure about it. Having an option to switch the behavior sounds very good, though!

relu91:
> Thank you for this PR, @relu91! :) In my opinion, the changes look pretty good, although while going through the code I found a lot of potential for further refactoring

I bet you found them! As I said in the intro: 
> There are plenty of other changes that I would like to add, but I didn't want to overload the changes shown here. I also spotted some bugs/discrepancies in the handlers of the different TD affordances, which proves again the benefit of these refactors. I'll fix the problems in the following PRs.

Keep your comments and maybe we can discuss your points in our Discord channel to further coordinate. 

> Regarding the trailing slashes: From my understanding, /things and /things/ should in fact be treated differently (I think the trailing slash introduces an empty path segment), but I am not entirely sure about it. Having an option to switch the behavior sounds very good, though!

So what are you suggesting here? to keep the current status (e.g. treating them as before) and expose the `Router` option to the user?

JKRhb:
> So what are you suggesting here? to keep the current status (e.g. treating them as before) and expose the `Router` option to the user?

I guess that could be a good solution to not introduce a breaking change for now. For a new major (or at this stage: minor) release, we could then simply switch the default.

> Keep your comments and maybe we can discuss your points in our Discord channel to further coordinate.

That sounds good! :)

danielpeintner:
I agree, once we have other breaking changes we can integrate this breaking change as well

relu91:
@danielpeintner @JKRhb given our discussion, we should be ready to merge this.

JKRhb:
Thank you, @relu91, LGTM :)

danielpeintner (approved):

danielpeintner:
LGTM as well. @relu91 feel free to merge and we can continue to work on the remaining TODOs left in this PR.

BTW, I created https://github.com/eclipse-thingweb/node-wot/issues/1037 to keep track of using the default option for not ignoring slashes

relu91:
> LGTM as well. @relu91 feel free to merge and we can continue to work on the remaining TODOs left in this PR.
> 
> BTW, I created #1037 to keep track of using the default option for not ignoring slashes

Thanks! 🥳

PR #989, opened 2023-05-10
docs: proposal for listing maintainers

danielpeintner:
fixes https://github.com/eclipse/thingweb.node-wot/issues/986

PR #1038, opened 2023-07-21
chore: create CODEOWNERS file

JKRhb:
I took the freedom of creating an initial version of a `CODEOWNERS` file that corresponds with the assignments from https://github.com/eclipse-thingweb/node-wot/pull/989.

With the file, the respective code owners are automatically requested for a review which can be especially useful when the project should further grow in size. However, only repository members with write-access can be added as code owners, which is why some of the packages are currently commented out instead of listing their maintainers as given in the README files (in this case, the default code owners will be requested for review instead).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,38 @@
+# Default owners (used as a fallback, last matching pattern takes the most precedence)
+* @danielpeintner @relu91
+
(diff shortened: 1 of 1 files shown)


relu91 (approved):
LGTM! thank you for adding :)

danielpeintner (approved):

PR #1040, opened 2023-07-26
fix (binding-modbus): Reading Modbus URI design

sebastiankb:
small fix: If the href is used to assign the register number, quantity and unitId, this was not processed correctly in the validateAndFillDefaultForm method. The parameters were not taken over into the new result set.


relu91 (approved):
I discovered that the tests were in place but wrongly checked the input form and not the result (which is the one really used by the client, see line 151). I fixed the tests and now the PR is ready to go. Given that Sebastian is a committer we can merge right away once CI is green.

PR #1042, opened 2023-07-28
Fix action handling for no contents

### LINKED ISSUE #1041: Invoking an action with no output over HTTP
I've noticed that the HTTP binding always returns `200 OK` after invoking an action, even if the response has no content. See e.g. the response to a POST on http://plugfest.thingweb.io:8083/testthing/actions/void-void.

Yet, the HTTP Semantics RFC states the following ([sec. 15.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#name-200-ok):
> a 200 response is expected to contain message content unless the message framing explicitly indicates that the content has zero length. If some aspect of the request indicates a preference for no content upon success, the origin server ought to send a 204 (No Content) response instead.

In some Consumer code, I use [Apache HttpComponents](https://hc.apache.org/index.html), which assumes a `200 OK` has some content. This is problematic if I use the async API of the library: because the response has no content, the `Future` object supposed to resolve with the content of the response never resolves. See e.g. the [basic async example](https://github.com/apache/httpcomponents-client/blob/5.1.x/httpclient5/src/test/java/org/apache/hc/client5/http/examples/AsyncClientHttpExchange.java) provided in the library's documentation.

Would it be possible to add a `Content-Length` header to the response? Or set status to `204 No Content` if the action has no output?

relu91:
This PR tackles both problems described in #1041. In particular, now the HTTP server responds using 204 when no content is provided and the payload contains the number zero for the testing ( see 1e7b3fc05d870734ae32249cb7b3df6c8d067dab). 

Fix #1041

### DIFF
--- a/packages/binding-http/test/http-server-test.ts
+++ b/packages/binding-http/test/http-server-test.ts
@@ -213,6 +213,40 @@ class HttpServerTest {
         return httpServer.stop();
     }
 
+    @test async "should return 204 when action has not output"() {
+        const httpServer = new HttpServer({ port: 0 });
+
+        await httpServer.start(null);
+
+        const testThing = new ExposedThing(null, {
+            title: "Test",
+            actions: {
+                noOutput: {
+                    output: { type: "string" },
+                    forms: [],
+                },
+            },
+        });
+
+        testThing.setActionHandler("noOutput", (input: WoT.InteractionOutput) => {
+            return new Promise<undefined>((resolve, reject) => {
+                resolve(undefined);
+            });
+        });
+
+        await httpServer.expose(testThing);
+
+        const uri = `http://localhost:${httpServer.getPort()}/test/`;
+
+        debug(`Testing ${uri}`);
+
+        const resp = await await fetch(uri + "actions/noOutput", { method: "POST" });
+
+        expect(resp.status).to.equal(204);
+
+        return httpServer.stop();
+    }
+
     @test async "should check uriVariables consistency"() {
         const httpServer = new HttpServer({ port: 0 });
 
--- a/packages/binding-http/test/http-server-test.ts
+++ b/packages/binding-http/test/http-server-test.ts
@@ -204,10 +204,16 @@ class HttpServerTest {
         resp = await (await fetch(uri + "properties/test")).text();
         expect(resp).to.equal('"on"');
 
-        resp = await (await fetch(uri + "actions/try", { method: "POST", body: "toggle" })).text();
+        let actionHttpResponse = await fetch(uri + "actions/try", { method: "POST", body: "toggle" });
+        resp = await actionHttpResponse.text();
+
+        expect(actionHttpResponse.status).to.equal(200);
         expect(resp).to.equal('"TEST"');
 
+        actionHttpResponse = await fetch(uri + "actions/try", { method: "POST", body: undefined });
         resp = await (await fetch(uri + "actions/try", { method: "POST", body: undefined })).text();
+
+        expect(actionHttpResponse.status).to.equal(200);
         expect(resp).to.equal('"TEST"');
 
         return httpServer.stop();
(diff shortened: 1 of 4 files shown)


packages/core/src/exposed-thing.ts
```diff
@@ -369,7 +369,7 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
                     new InteractionOutput(inputContent, form, this.actions[name].input),
                     options
                 );
-                if (result) {
+                if (result !== undefined) {
```
JKRhb, line 372: We probably need to scan the whole codebase for cases like these at some point – IIRC, there are a number of places where we check for "falsy" values instead of `undefined` or `null`. I think there is the possibility to make the eslint rules a bit stricter in this regard to help us with that.

JKRhb (approved):
Thank you, @relu91, the changes look good to me :) See a couple of minor suggestions/comments below.

packages/core/src/exposed-thing.ts
relu91: I would say that sometimes in fine to use the simpler `if` when we are expecting an object, but yeah a more stricter check might be good to spot problems like the one above. Let's create a tracking issue.

packages/binding-http/test/http-server-test.ts
```diff
@@ -213,6 +213,40 @@ class HttpServerTest {
...
+            });
+        });
+
+        await httpServer.expose(testThing);
+
+        const uri = `http://localhost:${httpServer.getPort()}/test/`;
+
+        debug(`Testing ${uri}`);
+
+        const resp = await await fetch(uri + "actions/noOutput", { method: "POST" });
```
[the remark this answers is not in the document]
relu91: fixed in [ee568df](https://github.com/eclipse-thingweb/node-wot/pull/1042/commits/ee568dffeaafb8ffce0e0ed4bf8affafc655121a)

packages/binding-http/test/http-server-test.ts
```diff
@@ -213,6 +213,40 @@ class HttpServerTest {
...
+                    forms: [],
+                },
+            },
+        });
+
+        testThing.setActionHandler("noOutput", (input: WoT.InteractionOutput) => {
+            return new Promise<undefined>((resolve, reject) => {
+                resolve(undefined);
+            });
+        });
```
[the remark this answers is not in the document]
relu91: fixed in [ee568df](https://github.com/eclipse-thingweb/node-wot/pull/1042/commits/ee568dffeaafb8ffce0e0ed4bf8affafc655121a)

packages/binding-http/test/http-server-test.ts
```diff
@@ -213,6 +213,36 @@ class HttpServerTest {
...
+
+        await httpServer.expose(testThing);
+
+        const uri = `http://localhost:${httpServer.getPort()}/test/`;
+
+        debug(`Testing ${uri}`);
+
+        const resp = await fetch(uri + "actions/noOutput", { method: "POST" });
+
+        expect(resp.status).to.equal(204);
```
danielpeintner, line 241: Just for the sake of completeness. Does it make sense to add

`expect(resp.status).to.equal(200);`

to one of the other action tests?

danielpeintner (approved):
PR looks good.

Very minor comment below that is up to you ...

packages/binding-http/test/http-server-test.ts
relu91: Yes, let me see if there is a place where it fits.

relu91: tackled in https://github.com/eclipse-thingweb/node-wot/pull/1042/commits/479d66d543d7d5b6878939a8516467fd8daff790 . Feel free to merge it if it is ok.

PR #1047, opened 2023-08-01
strict tsconfig checks in place for binding-file

danielpeintner:
relevant for to https://github.com/eclipse-thingweb/node-wot/issues/758

### DIFF
--- a/packages/binding-file/src/file-client.ts
+++ b/packages/binding-file/src/file-client.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -86,8 +86,9 @@ export default class FileClient implements ProtocolClient {
         error?: (error: Error) => void,
         complete?: () => void
     ): Promise<Subscription> {
-        error(new Error(`FileClient does not implement subscribe`));
-        return null;
+        return new Promise<Subscription>((resolve, reject) => {
+            reject(new Error(`FileClient does not implement subscribe`));
+        });
     }
 
     public async start(): Promise<void> {
--- a/packages/binding-file/tsconfig.json
+++ b/packages/binding-file/tsconfig.json
@@ -1,6 +1,8 @@
 {
     "extends": "../../tsconfig.json",
     "compilerOptions": {
+        "strict": true,
+        "strictFunctionTypes": true,
         "outDir": "dist",
         "rootDir": "src"
     },


packages/binding-file/src/file-client.ts
```diff
@@ -86,8 +86,9 @@ export default class FileClient implements ProtocolClient {
         error?: (error: Error) => void,
         complete?: () => void
     ): Promise<Subscription> {
-        error(new Error(`FileClient does not implement subscribe`));
-        return null;
+        return new Promise<Subscription>((resolve, reject) => {
```
relu91, line 89: I prefer throwing directly. 🤔 is it possible?

packages/binding-file/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
     "extends": "../../tsconfig.json",
     "compilerOptions": {
+        "strict": true,
+        "strictFunctionTypes": true,
```
relu91, line 5: I am doing some changes for other packages, is `strictFunctionTypes` required here? or is it sufficient to set `strict` to true?

relu91 (commented):
Great thank you, I have two comments before approving. One side note, I'm working on  `binding-http` from about 80 errors I'm down to six. Let's see if I can provide a PR for this afternoon.

packages/binding-file/src/file-client.ts
danielpeintner: Not sure what you mean by "throwing directly" ?

I just aligned this method with what is done in all the other methods in this class
see https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-file/src/file-client.ts

packages/binding-file/tsconfig.json
danielpeintner: I think what I would like to see in the end is a _global_ tsconfig.json file used for all packages. 
Hence I took the same settings that are currently used in `core` to not getting any surprise later (and yes, `strictFunctionTypes `might not be required here )

https://github.com/eclipse-thingweb/node-wot/blob/b25d863ab85774570bf203e126670d2975a0aa77/packages/core/tsconfig.json#L4-L5

packages/binding-file/src/file-client.ts
relu91: I mean in theory you don't need to create the promise object and reject it. You can `throw new Error("...")` directly. But I see know that you used the same pattern that was in the file. So for consistency, I'm ok with this, we can fix it later.

packages/binding-file/tsconfig.json
relu91: Checked the [doc](https://www.typescriptlang.org/tsconfig/#strict):

> Turning this on is equivalent to **enabling all** of the strict mode family options, which are outlined below.

So yeah in a followup PR we should remove this one here and the one in core (or everywhere else).

PR #1048, opened 2023-08-01
Strict checking for http-binding package

relu91:
Yet another big PR, really sorry but I couldn't make it shorter because otherwise, the code base would not even compile. 

relates to #758

### DIFF
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -31,7 +31,7 @@ export interface BasicCredentialConfiguration {
 export class BasicCredential extends Credential {
     private readonly username: string;
     private readonly password: string;
-    private readonly options: BasicSecurityScheme;
+    private readonly options: BasicSecurityScheme | undefined;
     /**
      *
      */
@@ -114,7 +114,7 @@ export class BasicKeyCredential extends Credential {
 
 export class OAuthCredential extends Credential {
     private token: Token | Promise<Token>;
-    private readonly refresh: () => Promise<Token>;
+    private readonly refresh: (() => Promise<Token>) | undefined;
 
     /**
      *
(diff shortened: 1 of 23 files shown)


packages/binding-http/src/credential.ts
```diff
@@ -167,9 +167,9 @@ export class TuyaCustomBearer extends Credential {
     protected key: string;
     protected secret: string;
     protected baseUri: string;
-    protected token: string;
-    protected refreshToken: string;
-    protected expireTime: Date;
+    protected token: string | undefined;
```
danielpeintner, line 170: I am just wondering.. would the following not work?

`protected token?: string;`

OR do you want to convey that `token` is **not** optional?

danielpeintner (approved):
I looked through the code and I did not spot any issue..

Thanks for all the work!

Below a minor question...

packages/binding-http/tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDir": "src"
+        "rootDir": "src",
+        "strict": true
```
danielpeintner, line 6: Do we want to add 

`"strictFunctionTypes": true`

as we do in the other tsconfigs?

danielpeintner: I saw the comment in https://github.com/eclipse-thingweb/node-wot/pull/1047#discussion_r1281537969
Hence we can tackle this comment as addressed

packages/binding-http/src/credential.ts
relu91: addressed with https://github.com/eclipse-thingweb/node-wot/pull/1048/commits/5469a3ada4a6da4a368388bc02a4aa94276caf38 (I checked the whole file to have consistency, but in other places we might use the old way)

danielpeintner (approved):
LGTM

danielpeintner (approved):

PR #1050, opened 2023-08-03
Use ts strict checking for binding modbus

relu91:
Yet another down. 

Relates to #758.

### DIFF
--- a/packages/binding-modbus/src/modbus-connection.ts
+++ b/packages/binding-modbus/src/modbus-connection.ts
@@ -15,13 +15,14 @@
 import ModbusRTU from "modbus-serial";
 import { ReadCoilResult, ReadRegisterResult } from "modbus-serial/ModbusRTU";
 import { ModbusEntity, ModbusFunction, ModbusForm } from "./modbus";
-import { Content, createLoggers, Endianness } from "@node-wot/core";
+import { Content, ContentSerdes, createLoggers, Endianness } from "@node-wot/core";
 import { Readable } from "stream";
 import { inspect } from "util";
 
 const { debug, warn, error } = createLoggers("binding-modbus", "modbus-connection");
 
 const configDefaults = {
+    connectionTimeout: 1000,
     operationTimeout: 2000,
     connectionRetryTime: 10000,
     maxRetries: 5,
@@ -100,7 +101,7 @@ class ModbusTransaction {
             } catch (err) {
                 warn(`Read operation failed on ${this.base}, len: ${this.quantity}, ${err}`);
                 // inform all operations and the invoker
-                this.operations.forEach((op) => op.failed(err));
+                this.operations.forEach((op) => op.failed(err instanceof Error ? err : new Error(JSON.stringify(err))));
                 throw err;
             }
         } else {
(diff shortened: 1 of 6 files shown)


packages/binding-modbus/src/modbus-connection.ts
```diff
@@ -461,12 +483,21 @@ export class PropertyOperation {
     done(base?: number, buffer?: Buffer): void {
         debug("Operation done");
 
+        if (!this.resolve || !this.reject) {
+            throw new Error("Invoked done before executing operation");
```
danielpeintner, line 487:
```suggestion
            throw new Error("Invoke done before executing operation");
```

I think that's what you meant, right?
OR

`Invoke called before executing operation`

packages/binding-modbus/src/modbus-connection.ts
```diff
@@ -490,6 +521,9 @@ export class PropertyOperation {
      */
     failed(reason: Error): void {
         warn(`Operation failed: ${reason}`);
+        if (!this.reject) {
+            throw new Error("Invoked reject before executing operation");
```
danielpeintner, line 525: Again not sure about the _passive_ form of invoked...

danielpeintner (commented):
Looks good in general.

Have some understanding problems w.r.t. some error messages. See below

packages/binding-modbus/src/modbus-connection.ts
relu91: Mmm not sure, what I meant is that the function `done` was invoked before executing the operation. Maybe it is better to be more explicit here: `Invoked function "done" before executing the Modbus operation` ?

danielpeintner: Your proposal makes sense.
An alternative *could* be 

`Function "done" was executed before the (this?) Modbus operation`

Either way is fine by me 👍

packages/binding-modbus/src/modbus-connection.ts
danielpeintner: see comment above

danielpeintner:
> ❗ Your organization is not using the GitHub App Integration. As a result you may experience degraded service beginning May 15th. Please [install the Github App Integration](https://github.com/apps/codecov) for your organization. [Read more](https://about.codecov.io/blog/codecov-is-updating-its-github-integration/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None).

I tried already several times to activate the GitHub integration but I do not know "what" blocks it?

image

Shall I open a ticket on Eclipse side?

danielpeintner:
> Shall I open a ticket on Eclipse side?

FYI: I opened https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3545

packages/binding-modbus/src/modbus-connection.ts
relu91: I changed it a little bit more in https://github.com/eclipse-thingweb/node-wot/pull/1050/commits/dbf99795f515e01401a5d17ab1a460046cb2f0e3. I hope it is ok, merge if it is ok! 🛩️

relu91:
@danielpeintner do you think we can merge it even if we have redcrosses in codecov?

danielpeintner (approved):

danielpeintner:
> @danielpeintner do you think we can merge it even if we have redcrosses in codecov?

We are now seeing these kind of "errors" in all our PRs now. I think we need to properly configure it and maybe show it as a warning or so.

Anyhow, w.r.t. this PR I think we can merge as is

PR #1051, opened 2023-08-03
Use ts strict checking for Websockets

relu91:
Another down. 

See #758


danielpeintner (approved):
Looks good, thanks @relu91!

PR #1052, opened 2023-08-07
AAS next

danielpeintner:
MAY include
* updates to TD conversion
* support TD to AAS/AID conversion


egekorkan:
After a talk with @Kaz040 and @danielpeintner, we have realized that TD -> AID will have some limitations at the moment. AID do not support multiple forms per affordance at the moment so we need to be selective. The proposal is for now to do the following:

If a TD has multiple forms, we follow the following implementation steps and give options to the developer:
1. Take the first form (default behavior)
2. Give option to pick form based on a certain preference
  - Protocol
  - Type of href: IPv4, IPv6, Domain name
  - Content Type
  - Security

I think that this form picking option can be almost put into a more generic place since I could also think of a WoT consumer needing a logic to pick such a form.

danielpeintner:
Some more questions w.r.t. alignment between TD and AID arises about `title`(s) and `description`(s):
* AID has the term `description` but in fact it contains descriptions arrays like `[{"language": "en", "text": "Counter"}, {"language": "de", "text": "Zähler"}]`
* AID does not have `descriptions` ...
* `title` uses just `string` like `{"idShort": "title", "valueType": "xs:string", "value": "Count"}`
* ...

--> seems inconsistent

danielpeintner:
The PR is read for review. It allows for 
* transforming AAS or AID submodel(s) to TD
* converting TD to AAS or AID submodel

Please have a look @egekorkan @sebastiankb @Kaz040 and others.
Note1: it is a BIG rewrite, hence the diff might not help a lot.
Note2: There are tests but I expect follow-up PRs.

The important API calls are in the following lines.

https://github.com/eclipse-thingweb/node-wot/blob/c06c67cfb58c7e1db5725226b136cb3423de0d85/packages/td-tools/src/util/asset-interface-description.ts#L43-L188

sebastiankb (approved):

PR #1053, opened 2023-08-08
Use ts strict checking for binding mqtt

egekorkan:
I had started on this but did not manage to move forward due to time constraints. I still wanted to make it public so that others can take the work.

@hasanheroglu maybe you can have a quick look? Otherwise, I will try to get back to it in two weeks :)

### DIFF
--- a/packages/binding-mqtt/src/mqtt-client.ts
+++ b/packages/binding-mqtt/src/mqtt-client.ts
@@ -40,7 +40,7 @@ export default class MqttClient implements ProtocolClient {
         this.scheme = "mqtt" + (secure ? "s" : "");
     }
 
-    private client: mqtt.MqttClient = undefined;
+    private client?: mqtt.MqttClient;
 
     public subscribeResource(
         form: MqttForm,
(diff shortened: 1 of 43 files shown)


packages/binding-mqtt/src/mqtt-client.ts
```diff
@@ -40,7 +40,7 @@ export default class MqttClient implements ProtocolClient {
         this.scheme = "mqtt" + (secure ? "s" : "");
     }
 
-    private client: mqtt.MqttClient = undefined;
+    private client: mqtt.MqttClient | undefined = undefined;
```
JKRhb, line 43: I think this could be simplified a bit :)

```suggestion
    private client?: mqtt.MqttClient;
```

relu91:
@danielpeintner should be ready to go. Before merging I'll do some rebase to clean up the commit messages.

danielpeintner (approved):

relu91:
argh I forgot to squash :/

egekorkan:
Thanks for doing it! :)

PR #1054, opened 2023-08-09
chore(binding-coap): use ts strict checking

JKRhb:
This PR should tick another box of #758 :)

I noticed only now, though, that I also made a slight API change to the `core` package in the context of this PR. Having another look at it, the change seems reasonable, though. If this PR should only be limited to the `binding-coap` package, then I could also revert the change and find a workaround for the time being.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -63,7 +63,7 @@ export default class CoapServer implements ProtocolServer {
     private readonly port: number;
     private readonly address?: string;
 
-    private mdnsIntroducer: MdnsIntroducer;
+    private mdnsIntroducer?: MdnsIntroducer;
 
     private readonly server: Server = createServer(
         { reuseAddr: false },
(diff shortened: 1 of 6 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -524,9 +525,9 @@ export default class CoapServer implements ProtocolServer {
         const listener = this.createContentListener(req, res, this.PROPERTY_DIR, affordanceKey);
 
         try {
-            await thing.handleObserveProperty(affordanceKey, listener, null);
```
JKRhb, line 527: @relu91 Hmm, so far the binding implementation always passed `null` to the `handleObserveProperty` and `handleUnobservProperty` methods. Not sure why that is the case. I will have another look at this in the binding itself and keep the API in core untouched then :)

JKRhb: I think the latest commits should now have provided the correct fix for the issue :) I think that this should also make the behavior of the server more predictable, as the form index was apparently not taken into account before when a property was observed.

relu91 (approved):
Thanks! I think it is ready to go 👍🏻 Maybe should we rebase the last three fixup commits?

JKRhb:
> Thanks! I think it is ready to go 👍🏻 Maybe should we rebase the last three fixup commits?

Awesome, thank you :) I just rebased the PR, should be ready to merge now :)

danielpeintner (approved):
Looks good, thanks!

PR #1056, opened 2023-08-11
chore(ci): re-enable Node 20

### LINKED ISSUE #1004: Node.js 20 workflow causes issues with modbus and mbus binding (related to serialport)
We need to investigate the reason why Node.js 20 is not working as expected

----
A proposal could be to run Node.js 16 and 18 only (for now) and we investigate why mbus and modbus are failing with Node.js 20.
What do you think?

_Originally posted by @danielpeintner in https://github.com/eclipse-thingweb/node-wot/issues/995#issuecomment-1553062342_

JKRhb:
This PR updates the node-mbus dependency, incorporating a fix for Node 20, and re-enables the corresponding CI workflow.

Fixes #1004.

### DIFF
--- a/.github/workflows/ci-full.yaml
+++ b/.github/workflows/ci-full.yaml
@@ -8,7 +8,7 @@ jobs:
         strategy:
             matrix:
                 os: [macos-latest, ubuntu-latest, windows-latest]
-                node-version: [16.x, 18.x]
+                node-version: [16.x, 18.x, 20.x]
 
         steps:
             - name: Checkout
--- a/.github/workflows/ci-full.yaml
+++ b/.github/workflows/ci-full.yaml
@@ -19,9 +19,6 @@ jobs:
               with:
                   node-version: ${{ matrix.node-version }}
                   cache: "npm"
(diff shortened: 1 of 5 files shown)


JKRhb:
Hmm. For some reason, the CI still fails on Node 20 (but for a different reason now?) I will investigate this a bit more.

JKRhb:
I have been able to identify the installation step of npm 7 as the last remaining cause of this issue :) A full fix should therefore be around the corner.

JKRhb:
The PR should now be ready for review :)

relu91 (approved):
Perfect! thank you @JKRhb 🎉. I'm also glad that we finally can get off that custom step for installing npm 7.  Good to go.

danielpeintner (changes requested):
Very good, thank you 👍

Can you remove also the following line in this PR
https://github.com/eclipse-thingweb/node-wot/blob/597f34dc6e32a3b7a68117d3ac51830c42a6df81/README.md?plain=1#L103

danielpeintner (approved):

PR #1059, opened 2023-08-11
chore(binding-mbus): enable strict type checking

danielpeintner:
another one of https://github.com/eclipse-thingweb/node-wot/issues/758

@LennyHEVS @Chxresubles can you take a look?
We are enabling strict checking for all node-wot packages

### DIFF
--- a/packages/binding-mbus/src/mbus-connection.ts
+++ b/packages/binding-mbus/src/mbus-connection.ts
@@ -156,7 +156,7 @@ export class MBusConnection {
         }
     }
 
-    async execute(op: PropertyOperation): Promise<Content | PromiseLike<Content>> {
+    async execute(op: PropertyOperation): Promise<Content | PromiseLike<Content> | undefined> {
         this.trigger();
         return op.execute();
     }
@@ -182,7 +182,7 @@ export class MBusConnection {
                 // inform all the operations that the connection cannot be recovered
                 this.queue.forEach((transaction) => {
                     transaction.operations.forEach((op) => {
-                        op.failed(`${error}`);
+                        op.failed(error instanceof Error ? error : new Error(JSON.stringify(error)));
                     });
                 });
             }
@@ -215,7 +215,9 @@ export class MBusConnection {
         } catch (err) {
             warn(`Read operation failed on unit ${transaction.unitId}. ${err}.`);
             // inform all operations and the invoker
-            transaction.operations.forEach((op) => op.failed(`${error}`));
+            transaction.operations.forEach((op) =>
+                op.failed(error instanceof Error ? error : new Error(JSON.stringify(error)))
+            );
             throw err;
         }
     }
@@ -278,11 +280,13 @@ export class PropertyOperation {
      * Trigger execution of this operation.
      *
      */
-    async execute(): Promise<Content | PromiseLike<Content>> {
-        return new Promise<Content>((resolve, reject) => {
-            this.resolve = resolve;
-            this.reject = reject;
-        });
+    async execute(): Promise<(Content | PromiseLike<Content>) | undefined> {
+        return new Promise(
+            (resolve: (value?: Content | PromiseLike<Content>) => void, reject: (reason?: Error) => void) => {
+                this.resolve = resolve;
+                this.reject = reject;
+            }
+        );
     }
 
     /**
@@ -315,12 +319,11 @@ export class PropertyOperation {
 
         const resp = new Content("application/json", Readable.from(JSON.stringify(payload)));
 
-        // resolve the Promise given to the invoking script
-        if (this.resolve) {
-            this.resolve(resp);
-        } else {
-            warn("resolve undefined");
+        if (!this.resolve) {
+            throw new Error("Function 'done' was invoked before executing the Mbus operation");
         }
+        // resolve the Promise given to the invoking script
+        this.resolve(resp);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


danielpeintner:
I tried to update the code, please have a look...

packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -182,21 +182,25 @@ export class MBusConnection {
                 // inform all the operations that the connection cannot be recovered
                 this.queue.forEach((transaction) => {
                     transaction.operations.forEach((op) => {
-                        op.failed(error);
+                        op.failed(`${error}`);
```
relu91, line 185: In Modbus we are using `op.failed(err instanceof Error ? err : new Error(JSON.stringify(err)))` I think we should be consistent.

packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -211,7 +215,7 @@ export class MBusConnection {
         } catch (err) {
             warn(`Read operation failed on unit ${transaction.unitId}. ${err}.`);
             // inform all operations and the invoker
-            transaction.operations.forEach((op) => op.failed(err));
+            transaction.operations.forEach((op) => op.failed(`${error}`));
```
relu91, line 218: Same as above use `op.failed(err instanceof Error ? err : new Error(JSON.stringify(err)))`

packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -311,7 +316,11 @@ export class PropertyOperation {
         const resp = new Content("application/json", Readable.from(JSON.stringify(payload)));
 
         // resolve the Promise given to the invoking script
-        this.resolve(resp);
+        if (this.resolve) {
+            this.resolve(resp);
+        } else {
+            warn("resolve undefined");
+        }
```
relu91, line 323: In Modbus we are throwing an error: 
```
   failed(reason: Error): void {
        warn(`Operation failed: ${reason}`);
        if (!this.reject) {
            throw new Error("Function 'failed' was invoked before executing the Modbus operation");
        }
        // reject the Promise given to the invoking script
        this.reject(reason);
    }
```
What do you think?

packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -322,6 +331,10 @@ export class PropertyOperation {
     failed(reason: string): void {
         warn("Operation failed:", reason);
         // reject the Promise given to the invoking script
-        this.reject(new Error(reason));
+        if (this.reject) {
+            this.reject(new Error(reason));
+        } else {
+            warn("reject undefined");
+        }
```
relu91, line 338: See the comment above.

relu91 (changes requested):
Since mbus code is pretty similar structurally to Modbus I noticed some dissimilarities. I think we can align the two implementations, but feel free to prove me wrong.

packages/binding-mbus/src/mbus-connection.ts
JKRhb: Maybe we could create a helper function to be consistent across packages here?

packages/binding-mbus/src/mbus-connection.ts
danielpeintner: done

packages/binding-mbus/src/mbus-connection.ts
danielpeintner: Alignment makes sense

packages/binding-mbus/src/mbus-connection.ts
danielpeintner: done also

packages/binding-mbus/src/mbus-connection.ts
danielpeintner: done

packages/binding-mbus/src/mbus-connection.ts
```diff
@@ -182,21 +182,25 @@ export class MBusConnection {
...
+                try {
+                    await this.executeTransaction(this.currentTransaction);
+                    this.currentTransaction = undefined;
+                    this.trigger();
+                } catch (err) {
+                    warn(`Transaction failed: ${err}`);
+                    this.currentTransaction = undefined;
+                    this.trigger();
+                }
             }
```
JKRhb, line 204: I guess the function could also return early here, right? That would get rid of the else clause and therefore another level of nesting.

```suggestion
            if (!this.currentTransaction) {
                warn(`Current transaction is undefined -> transaction not executed`);
                return;
            }

            try {
                await this.executeTransaction(this.currentTransaction);
                this.currentTransaction = undefined;
                this.trigger();
            } catch (err) {
                warn(`Transaction failed: ${err}`);
                this.currentTransaction = undefined;
                this.trigger();
            }
```

danielpeintner:
> Since mbus code is pretty similar structurally to Modbus I noticed some dissimilarities. I think we can align the two implementations

I wasn't aware that those two bindings are so similar. With my latest commit https://github.com/eclipse-thingweb/node-wot/pull/1059/commits/be3d6de0227ae6d5909cae3003b43163b525f7e8 I tried to align the code. Please have another look.

JKRhb (approved):

packages/binding-mbus/src/mbus-connection.ts
danielpeintner: Matter of taste but Yes, let's do it..

I grew up with MISRA that educated that _a function should have a single point of exit at the end_. 🙃

relu91 (approved):

relu91:
And with this, we fix #758.

packages/binding-mbus/src/mbus-connection.ts
JKRhb: Oh, yeah, I also did some research after you posted your initial comment and came across that page :D Interesting paradigm ;) I guess for JavaScript/TypeScript it might be a bit more idiomatic to use these kinds of "guard clauses", though, since we do not need to manually clean up memory, for example, before returning. In the end, it probably depends on the respective function, though, which approach is the most readable and maintainable one, as there are also many cases where one return statement at the end is the simplest approach.

Issue #1058, opened 2023-08-11
After integrating codecov we are getting failing CI builds


danielpeintner:
We should look into whether codecov integration can be configured to provide some details and maybe also warnings but not showing failures as it is currently done.

see for example https://github.com/eclipse-thingweb/node-wot/pull/1053

![image](https://github.com/eclipse-thingweb/node-wot/assets/11281461/903f0d13-412b-43ce-98c3-81d991707b4b)

danielpeintner:
I think this should do the trick

https://docs.codecov.com/docs#tips-and-tricks

PR #1060, opened 2023-08-11
chore: avoid blocking codecov status checks

danielpeintner:
see https://docs.codecov.com/docs#tips-and-tricks

fixes https://github.com/eclipse-thingweb/node-wot/issues/1058

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+    status:
+        project:
+            default:
+                informational: true
+        patch:
+            default:
+                informational: true
+# When modifying this file, please validate using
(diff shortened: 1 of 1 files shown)


danielpeintner:
Question: Mhh, how can we check whether it does what we asked for? 

In this PR there was no real change to complain 🤷‍♂️

relu91 (approved):
Let's try, but the question is when are we going to switch back to normal behaviour? after release?

JKRhb:
> Let's try, but the question is when are we going to switch back to normal behaviour? after release?

That sounds good :) Maybe we also try to increase code coverage a bit once we are done with the strict ts configs, then I think codecov will give a good incentive to always add tests when a new feature is integrated.

PR #1067, opened 2023-09-04
Send headers at connection for HTTP Server observe property

### LINKED ISSUE #1065: emitPropertyChange does not add content type to header
When working on https://github.com/eclipse-thingweb/node-wot/pull/1064, I have noticed that the code `thing.emitPropertyChange` doest not add the content type to the HTTP header. Funny enough, reading a property (not observing) or events gets the correct header.

relu91:
Fix #1065. We are now using the same policy that we used for events. 

Update: About tests, the current test suite is pretty messy. If you have ideas on how to start testing and better organizing the HTTP server feel free to write a proposal.

### DIFF
--- a/packages/binding-http/src/routes/property-observe.ts
+++ b/packages/binding-http/src/routes/property-observe.ts
@@ -66,6 +66,14 @@ export default async function propertyObserveRoute(
     if (req.method === "GET") {
         const listener = async (value: Content) => {
             try {
+                if (!res.headersSent) {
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):
As @relu91 mentioned it matches with what we do for events

https://github.com/eclipse-thingweb/node-wot/blob/7c0030949ca052704157ea69fb83c4b695bad05d/packages/binding-http/src/routes/event.ts#L68-L75

PR #1070, opened 2023-09-05
Add supports for multiple security schemes in http server

relu91:
This PR improves the exposing process of the `http-server`. In particular, it focuses on the ability to handle multiple security schemes per `http-server` instance. For example, now the `http-server` can be configure to be able to hosts Things without security and Things with basic schemes:
```js
const td1 = { } // automatically filled with nosec scheme
const td2 = {securityDefinitions: { basic: {scheme: "basic"}}, security: "basic" }

await (await WoT.produce(td1)).expose();
await (await WoT.produce(td2)).expose();
// ok 
```

**Note** that this does not mean that now we can use multiple security schemes in the `security` field or the `combo` SecurityScheme`. For multiple security schemas, I think we can avoid implementing that feature, mainly because it is already [deprecated in TD 1.1](https://w3c.github.io/wot-thing-description/#td-security-combo-deprecation). On the other hand, the combo scheme will require some additional work. Moreover, this does not introduce support for form-level security fields. It is quite tricky to achieve and we need first to support form templating in the expose method properly.

### DIFF
--- a/examples/browser/counter.html
+++ b/examples/browser/counter.html
@@ -44,7 +44,7 @@
                     readonly="readonly"
                     id="td_addr"
                     type="url"
-                    value="http://plugfest.thingweb.io:8083/counter"
+                    value="http://localhost:8080/counter"
                 /><!-- http://localhost:8080/counter -->
             </div>
         </div>
--- a/examples/browser/counter.html
+++ b/examples/browser/counter.html
@@ -44,7 +44,7 @@
                     readonly="readonly"
                     id="td_addr"
                     type="url"
-                    value="http://localhost:8080/counter"
+                    value="http://plugfest.thingweb.io:8083/counter"
                 /><!-- http://localhost:8080/counter -->
             </div>
         </div>
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -66,7 +66,7 @@ export default class HttpServer implements ProtocolServer {
     private readonly address?: string = undefined;
     private readonly baseUri?: string = undefined;
     private readonly urlRewrite?: Record<string, string> = undefined;
-    private readonly supportedSecurityScheme: string[] = ["nosec"];
+    private readonly supportedSecuritySchemes: string[] = ["nosec"];
     private readonly validOAuthClients: RegExp = /.*/g;
     private readonly server: http.Server | https.Server;
     private readonly middleware: MiddlewareRequestHandler | null = null;
@@ -176,7 +176,7 @@ export default class HttpServer implements ProtocolServer {
         if (config.security) {
             if (config.security.length > 1) {
                 // clear the default
-                this.supportedSecurityScheme = [];
+                this.supportedSecuritySchemes = [];
             }
             for (const securityScheme of config.security) {
                 switch (securityScheme.scheme) {
@@ -195,7 +195,7 @@ export default class HttpServer implements ProtocolServer {
                     default:
                         throw new Error(`HttpServer does not support security scheme '${securityScheme.scheme}`);
                 }
-                this.supportedSecurityScheme.push(securityScheme.scheme);
+                this.supportedSecuritySchemes.push(securityScheme.scheme);
             }
         }
     }
@@ -574,7 +574,7 @@ export default class HttpServer implements ProtocolServer {
                 );
             }
 
-            const isSupported = this.supportedSecurityScheme.find((supportedScheme) => {
+            const isSupported = this.supportedSecuritySchemes.find((supportedScheme) => {
                 const thingScheme = thing.securityDefinitions[secCandidate].scheme;
                 return thingScheme === supportedScheme.toLocaleLowerCase();
             });
@@ -582,7 +582,7 @@ export default class HttpServer implements ProtocolServer {
             if (!isSupported) {
                 throw new Error(
                     "Servient does not support thing security schemes. Current scheme supported: " +
-                        this.supportedSecurityScheme.join(", ")
+                        this.supportedSecuritySchemes.join(", ")
                 );
             }
             // We don't need to do anything else, the user has selected one supported security scheme.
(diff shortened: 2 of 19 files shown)


danielpeintner:
Thanks for your work @relu91 👍

No real review yet, just a note...
_changing the HTTP configuration from string to array is somewhat a breaking change but I can live with it ;-)_

`security?: TD.SecurityScheme;`
to
`security?: TD.SecurityScheme[];`

danielpeintner:
> Yes, I forgot to mention that important note. It sort of needed if we wanted to fix #873.

You might want to add "fixes https://github.com/eclipse-thingweb/node-wot/issues/873" to the first comment.

packages/binding-http/src/http-server.ts
```diff
@@ -606,14 +646,6 @@ export default class HttpServer implements ProtocolServer {
...
         });
 
-        // Set CORS headers
-        if (this.httpSecurityScheme !== "NoSec" && req.headers.origin) {
-            res.setHeader("Access-Control-Allow-Origin", req.headers.origin);
-            res.setHeader("Access-Control-Allow-Credentials", "true");
-        } else {
-            res.setHeader("Access-Control-Allow-Origin", "*");
-        }
-
```
danielpeintner, line 616: Note: I was surprised first "why" we no longer need CORS but the code has moved to "common"

packages/binding-http/src/routes/common.ts
```diff
@@ -79,3 +79,22 @@ export function isEmpty(obj: Record<string, unknown>): boolean {
...
+
+export function securitySchemeToHTTPHeader(scheme: string): string {
+    const [first, ...rest] = scheme;
+    // HTTP Authentication Scheme for OAuth does not contain the version number
+    // see https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml
+    if (scheme === "oauth2") return "OAuth";
+    return first.toUpperCase() + rest.join("").toLowerCase();
+}
+
+export function setCORSForThing(req: IncomingMessage, res: ServerResponse, thing: ExposedThing): void {
```
danielpeintner, line 91: Naming preference: setCorsForThing or setCORS4Thing

danielpeintner (commented):
I have some very minor questions/proposal... I can live with the *current* state also.

I just have one question/concern. Since we want to publish a release soon. Do you think we will see issues/regressions? I mean.. a lot has been changed and moved around... If you feel comfortable I am OK as well 👍

packages/binding-http/src/http-server.ts
relu91: Yeah, it is a tricky change, but in a nutshell: CORS management depends on the security schema used on that resource. Since now we don't have one single security strategy for the whole server, we need to handle CORS on each affordance and understand which security mechanisms is used for that specific thing (in the future for that specific form -> URL path).

relu91: Do you have any ideas to improve future devs about this choice? 🤔

danielpeintner: I just wanted to highlight it for other reviewers. I am fine with your decision 👍

packages/binding-http/src/http-server.ts
```diff
@@ -66,7 +66,7 @@ export default class HttpServer implements ProtocolServer {
     private readonly address?: string = undefined;
     private readonly baseUri?: string = undefined;
     private readonly urlRewrite?: Record<string, string> = undefined;
-    private readonly httpSecurityScheme: string = "NoSec"; // HTTP header compatible string
+    private readonly supportedSecurityScheme: string[] = ["nosec"];
```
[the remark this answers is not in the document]
relu91: Resolved in [3389714](https://github.com/eclipse-thingweb/node-wot/pull/1070/commits/33897149843e99183b7003cfe3e72412c96eaffb)

packages/binding-http/src/routes/common.ts
```diff
@@ -79,3 +79,22 @@ export function isEmpty(obj: Record<string, unknown>): boolean {
     }
     return true;
 }
+
+export function securitySchemeToHTTPHeader(scheme: string): string {
```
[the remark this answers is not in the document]
relu91: See [47554e5](https://github.com/eclipse-thingweb/node-wot/pull/1070/commits/47554e5e9ac059b8f8d8718f664a3786535654f2)

relu91:
> I just have one question/concern. Since we want to publish a release soon. Do you think we will see issues/regressions? I mean.. a lot has been changed and moved around... If you feel comfortable I am OK as well 👍

About this point, I was worried too. What I know is:
- OAuth2 example now is working
- All the tests are green
- Tried counter and other examples and they are all working (coffee machine has problems unrelated coming from URI variables)

examples/browser/counter.html
```diff
@@ -44,7 +44,7 @@ <h1>Counter Client Example in the Browser</h1>
                     readonly="readonly"
                     id="td_addr"
                     type="url"
-                    value="http://plugfest.thingweb.io:8083/counter"
+                    value="http://localhost:8080/counter"
```
danielpeintner, line 47: Do we want to keep the local or online counter instance?

packages/binding-http/test/http-server-oauth-tests.ts
```diff
@@ -79,7 +79,7 @@ class OAuthServerTests {
 
     @test async "should configure oauth"() {
         /* eslint-disable dot-notation */
-        this.server["httpSecurityScheme"].should.be.equal("OAuth");
+        this.server["supportedSecuritySchemes"].should.contain("oauth2");
         expect(this.server["oAuthValidator"]).to.be.instanceOf(EndpointValidator);
```
danielpeintner, line 83: Minor: since we are there we could resolve some of the eslint warnings and use the dot notation..

danielpeintner (commented):
I think the PR is fine. I added some minor comments..

 We just need to decide whether we want to merge it before or after the upcoming release

packages/binding-http/test/http-server-oauth-tests.ts
relu91: No, we can't because it is a private member.

examples/browser/counter.html
relu91: Right sorry, it was a leftover from local tests.

egekorkan:
Some notes:
- I definitely support this since it was a missing feature.
- I do not see issues to include it in the release. Is the main worry about not noticing an impact?
- Even though multiple security is deprecated, it will most probably come back in TD 2.0 the way it is done in OpenAPI atm. We will not have combo and different combinations of keywords will imply AND or OR combination.

relu91:
> Even though multiple security is deprecated, it will most probably come back in TD 2.0 the way it is done in OpenAPI atm. We will not have combo and different combinations of keywords will imply AND or OR combination.

Didn't know about this, I thought the tendency was to favor `combo` scheme. Well let's talk about this in the future. Btw I am also ok to merge given my comment above. @danielpeintner I leave to you the final choice.

danielpeintner (approved):

PR #1073, opened 2023-09-06
chore: use only base tsconfig strict

### LINKED ISSUE #758: `tsconfig.json` too benevolent

relu91:
Fix #758. I had to add additional changes because the `"strictFunctionTypes": false,` in the root tsconfig was hiding some problems in HTTP binding and examples. Even though the changes in the HTTP binding are more "missing" features of the find-my-way type, I think we can go over it and consider the issue fixed.


JKRhb (approved):

danielpeintner (approved):
LGTM.

While reviewing the PR I noticed differences in the `tsconfig.json` of some bindings.
In general the `compilerOptions` look like this.

```
  "compilerOptions": {
        "outDir": "dist",
        "rootDir": "src"
    }
```

Anyhow, there are some additions in some packages and I wonder whether we can align them

`"typeRoots": ["./node_modules/@types", "./typings"]`
... in `binding-coap`

`"resolveJsonModule": true`
... in `cli`, `core`, and `td-tools`

`"target": "ES2018"`
... in `examples`

`"types": ["node", "readable-stream"]`
... in `td-tools`

Do you think it is worthwhile investigating?

relu91:
Good points. Answering:
-` binding-coap`: It was a leftover of a workaround that we have for a bug in the typings of coap library. The commit above removes it. 
- `"resolveJsonModule": true"`: it is needed because this package requires some kind of JSON file (it can be the JSON schema or the package.json)
- `"target": "ES2018"` It is needed to have nice generated JS with await and async features. 
- `"types": ["node", "readable-stream"]` needed for webpack. 

I'm merging now.

PR #1074, opened 2023-09-07
Add static address to quickstart examples

egekorkan:
[The current deployed Things](http://plugfest.thingweb.io:8081/coffee-machine)http://[plugfest.thingweb.io:8081/coffee-machine](http://plugfest.thingweb.io:8081/coffee-machine) have wrong href due to automatic ip address detection. This fixes it.

With @relu91 we have discussed offline and we should simply move to a dockerized setup to manage all Things and their environment variables (like IP address, port etc.).

### DIFF
--- a/examples/quickstart/simple-coffee-machine.js
+++ b/examples/quickstart/simple-coffee-machine.js
@@ -24,6 +24,7 @@ servient.addServer(
         port: 8081,
     })
 );
+core_1.Helpers.setStaticAddress("plugfest.thingweb.io"); // comment this out if you are testing locally
 let waterAmount = 100;
 let beansAmount = 100;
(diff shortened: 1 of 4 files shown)


danielpeintner (approved):
I am okay with the change even though I think this is not _ideal_ for quickstart examples.

Can we come up with a better/other solution?
Is the issue gone in a dockerized setup? IF so, we should keep track of this issue to revert this change once docker is setup...

relu91:
In theory here, designing the examples using node-wot cli would have helped, but given the time constraints, we opted for this quick solution. After tpac we should review the examples and understand what we would like to show to newcomers and how.

relu91 (approved):

PR #1075, opened 2023-09-08
fix(examples/quickstart/smart-clock): change default port

relu91:
Change to an open port.


danielpeintner (approved):
LGTM.

Usage of the ports should be reflected in https://github.com/eclipse-thingweb/.github/blob/main/PORTS.md

Issue #1046, opened 2023-07-31
Strict boolean expressions


relu91:
> We probably need to scan the whole codebase for cases like these at some point – IIRC, there are a number of places where we check for "falsy" values instead of `undefined` or `null`. I think there is the possibility to make the eslint rules a bit stricter in this regard to help us with that.

_Originally posted by @JKRhb in https://github.com/eclipse-thingweb/node-wot/pull/1042#discussion_r1277735531_

As mentioned above, we are not currently verifying that our codebase uses the boolean expressions correctly. As we all know the nullish values can sometimes by tricked and led us to unwanted bugs (as the bugs that the PR above was solving). I tried to enable  `"@typescript-eslint/strict-boolean-expressions": "error"` into our eslint file, but that requires `strictNullChecks` to be enabled in `tsconfig.json`. Therefore, I suggest first completing the long standing #758 and then looking back to this issue. 

Packages to be fixed:
- [x] binding-coap
- [ ] binding-modbus
- [ ] browser-bundle
- [ ] binding-file
- [ ] binding-mqtt
- [ ] cli
- [ ] binding-netconf
- [x] core
- [ ] binding-opcua
- [x] examples
- [ ] binding-websocket
- [x] td-tools
- [x] binding-http
- [ ] binding-mbus

danielpeintner:
I suggest to fix the issue step by step like in https://github.com/eclipse-thingweb/node-wot/issues/758. That's why I added the "packages to be fixed list" above

PR #1077, opened 2023-09-08
chore(td-tools): enable eslint/strict-boolean-expressions and strictNullChecks

danielpeintner:


relu91:
I started working on this, but I gave up. Thank you for bringing this up! I think we should pay double attention to the changes... because sometimes we used `null` as synonym of `undefined`

packages/td-tools/src/td-parser.ts
```diff
@@ -222,20 +222,20 @@ export function serializeTD(thing: Thing): string {
     const copy = JSON.parse(JSON.stringify(thing));
 
     // clean-ups
-    if (!copy.security || copy.security.length === 0) {
+    if (copy.security === undefined || copy.security.length === 0) {
```
JKRhb, line 225:
```suggestion
    if (copy?.security.length === 0) {
```

packages/td-tools/src/td-parser.ts
```diff
@@ -222,20 +222,20 @@ export function serializeTD(thing: Thing): string {
...
         copy.securityDefinitions = {
             nosec_sc: { scheme: "nosec" },
         };
         copy.security = ["nosec_sc"];
     }
 
-    if (copy.forms && copy.forms.length === 0) {
+    if (copy.forms !== undefined && copy.forms.length === 0) {
```
JKRhb, line 232:
```suggestion
    if (copy?.forms.length === 0) {
```

packages/td-tools/src/td-parser.ts
```diff
@@ -265,11 +265,11 @@ export function serializeTD(thing: Thing): string {
             }
         }
     }
-    if (copy.events && Object.keys(copy.events).length === 0) {
+    if (copy.events !== undefined && Object.keys(copy.events).length === 0) {
         delete copy.events;
     }
 
-    if (copy.links && copy.links.length === 0) {
+    if (copy.links !== undefined && copy.links.length === 0) {
```
JKRhb, line 272:
```suggestion
    if (copy?.links.length === 0) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -241,7 +241,7 @@ export class ThingModelHelpers {
             case "http": {
                 return new Promise((resolve, reject) => {
                     http.get(uri, (res) => {
-                        if (!res.statusCode || res.statusCode !== 200) {
+                        if (res.statusCode === undefined || res.statusCode !== 200) {
```
JKRhb, line 244:
```suggestion
                        if (res?.statusCode !== 200) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -268,7 +268,7 @@ export class ThingModelHelpers {
                 return new Promise((resolve, reject) => {
                     https
                         .get(uri, (res) => {
-                            if (!res.statusCode || res.statusCode !== 200) {
+                            if (res.statusCode === undefined || res.statusCode !== 200) {
```
JKRhb, line 271:
```suggestion
                            if (res?.statusCode !== 200) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -508,7 +508,7 @@ export class ThingModelHelpers {
                 extendedModel.properties = {};
             }
             for (const key in properties) {
-                if (dest.properties && dest.properties[key]) {
+                if (dest.properties && dest.properties[key] !== undefined) {
```
JKRhb, line 511:
```suggestion
                if (dest?.properties[key] !== undefined) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -625,7 +625,7 @@ export class ThingModelHelpers {
         keys = keys.map((el) => el.replace("{{", "").replace("}}", ""));
         let isValid = true;
         let errors;
-        if (keys && keys.length > 0 && (map === undefined || map === null)) {
+        if (keys !== undefined && keys.length > 0 && (map === undefined || map === null)) {
```
JKRhb, line 628: With `foo == null`, you can actually check for `undefined` and `null` at the same time (from my knowledge, this is one of the few cases where `==` is valid to use).

```suggestion
        if ((keys ?? []).length > 0 && (map == null)) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -376,7 +376,7 @@ export class ThingModelHelpers {
         if (!options) {
             options = {} as CompositionOptions;
         }
-        if (!options.baseUrl) {
+        if (options.baseUrl === undefined) {
             options.baseUrl = ".";
         }
```
JKRhb, line 381: Statements like these might actually be able to be simplified like this:

```suggestion
        options.baseUrl ??= ".";
```

danielpeintner: You are right, but to be honest I still have *issues* understanding the nullish coalescing operator at the first look. Is it just me? :-)
Shall we go over and update all of them?

danielpeintner:
One more thing: Since there are lots of changes I am now more thinking about proposing PRs for each package step by step... otherwise it gets huge and hardly manageable.. 
What do people think?

danielpeintner:
> I started working on this, but I gave up. Thank you for bringing this up! I think we should pay double attention to the changes... because sometimes we used `null` as synonym of `undefined`

I think after doing some more work on it I noticed this issue as well.

In a sense 

`if (bla !== undefined) {`

is not enough and we would almost need 

`if (bla !== undefined && bla !== null) {`

which I really don't like...

Ideas/Proposals?

JKRhb:
> In a sense
> 
> `if (bla !== undefined) {`
> 
> is not enough and we would almost need
> 
> `if (bla !== undefined && bla !== null) {`
> 
> which I really don't like...
> 
> Ideas/Proposals?

I think in this case, you can (somewhat unintuitively) cover both with `bla != null`. When I integrated strict null checks in `node-coap`, that concept was very helpful in dealing with this scenario. As mentioned somewhere else, this is apparently one of the few scenarios where it is valid to use the `!=` operator instead of the `!==` operator.

packages/td-tools/src/thing-model-helpers.ts
danielpeintner: I don't think the proposed change leads to the same result ---> reverting for now..

danielpeintner:
@JKRhb I think some of the proposed changes are not correct and lead to issues. I suggest to keep them as is...
I think that's why the thing model helper tests are failing. I revert them ...

packages/td-tools/src/td-parser.ts
```diff
@@ -222,20 +222,20 @@ export function serializeTD(thing: Thing): string {
     const copy = JSON.parse(JSON.stringify(thing));
 
     // clean-ups
-    if (!copy.security || copy.security.length === 0) {
+    if (copy?.security.length === 0) {
```
danielpeintner, line 225: I also think this proposed change is not correct ... since it is OR not AND

danielpeintner: fixed with https://github.com/eclipse-thingweb/node-wot/pull/1077/commits/25d5390b5c9213128cd51404367da2bc98f0f6d9

danielpeintner:
@JKRhb @relu91 I am a bit lost and do not understand why some TM tests are failing. Can you spot the issue!?
I think I need a break 🙈

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -132,7 +132,7 @@ export class AssetInterfaceDescriptionUtil {
     public transformTD2SM(tdAsString: string, protocols?: string[]): string {
         const td: ThingDescription = TDParser.parseTD(tdAsString);
 
-        const aidID = td.id ? td.id : "ID" + Math.random();
+        const aidID = td.id != null ? td.id : "ID" + Math.random();
```
relu91, line 135:
```suggestion
        const aidID = td.id ?? "ID" + Math.random();
```

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -295,12 +295,12 @@ export class AssetInterfaceDescriptionUtil {
         endpointMetadata?: Record<string, unknown>
     ): string | [string, ...string[]] {
         const security: string[] = [];
-        if (endpointMetadata?.value && endpointMetadata.value instanceof Array) {
+        if (endpointMetadata?.value != null && endpointMetadata.value instanceof Array) {
```
relu91, line 298:
```suggestion
        if (endpointMetadata?.value instanceof Array) {
```

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -295,12 +295,12 @@ export class AssetInterfaceDescriptionUtil {
...
             for (const v of endpointMetadata.value) {
                 if (v.idShort === "security") {
-                    if (v.value && v.value instanceof Array) {
+                    if (v.value != null && v.value instanceof Array) {
```
relu91, line 301:
```suggestion
                    if (v.value instanceof Array) {
```

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -400,12 +399,16 @@ export class AssetInterfaceDescriptionUtil {
...
+            submodel instanceof Object &&
+            submodel.idShort != null &&
+            submodel.idShort === "AssetInterfacesDescription"
+        ) {
+            if (submodel.submodelElements != null && submodel.submodelElements instanceof Array) {
                 for (const submodelElement of submodel.submodelElements) {
                     if (submodelElement instanceof Object) {
                         logDebug("SubmodelElement.idShort: " + submodelElement.idShort);
-                        if (submodelRegex && typeof submodelRegex === "string" && submodelRegex.length > 0) {
+                        if (submodelRegex != null && typeof submodelRegex === "string" && submodelRegex.length > 0) {
```
relu91, line 411:
```suggestion
                        if (typeof submodelRegex === "string" && submodelRegex.length > 0) {
```

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -422,7 +425,7 @@ export class AssetInterfaceDescriptionUtil {
 
     private processSubmodelElement(smInformation: SubmodelInformation, submodelElement: Record<string, unknown>): void {
         // EndpointMetadata vs. InterfaceMetadata
-        if (submodelElement.value && submodelElement.value instanceof Array) {
+        if (submodelElement.value != null && submodelElement.value instanceof Array) {
```
relu91, line 428:
```suggestion
        if (submodelElement.value instanceof Array) {
```

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -446,7 +449,7 @@ export class AssetInterfaceDescriptionUtil {
                 if (smValue instanceof Object) {
                     if (smValue.idShort === "InterfaceMetadata") {
                         logInfo("InterfaceMetadata");
-                        if (smValue.value && smValue.value instanceof Array) {
+                        if (smValue.value != null && smValue.value instanceof Array) {
```
relu91, line 452:
```suggestion
                        if (smValue.value instanceof Array) {
```

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -625,7 +625,7 @@ export class ThingModelHelpers {
         keys = keys.map((el) => el.replace("{{", "").replace("}}", ""));
         let isValid = true;
         let errors;
-        if (keys && keys.length > 0 && (map === undefined || map === null)) {
+        if (keys != null && keys.length > 0 && (map === undefined || map === null)) {
```
relu91, line 628:
```suggestion
        if (keys?.length > 0 && (map === undefined || map === null)) {
```

relu91 (changes requested):
Some further simplifications.

packages/td-tools/src/thing-model-helpers.ts
relu91: Is this conversation resolved? Are we going to enable the rule for eslint? if that's so we should have a tracking issue.

JKRhb: I just opened #1092, so this conversation can be resolved :)

danielpeintner:
Note: I am not sure "why" but the CI gets stuck all the time in "Test with coverage report" with this PR. Local tests run fine...

relu91:
> Note: I am not sure "why" but the CI gets stuck all the time in "Test with coverage report" with this PR. Local tests run fine...

I was surprised too, let's see if #1093 get stuck too.

packages/td-tools/src/td-parser.ts
```diff
@@ -222,20 +222,20 @@ export function serializeTD(thing: Thing): string {
     const copy = JSON.parse(JSON.stringify(thing));
 
     // clean-ups
-    if (!copy.security || copy.security.length === 0) {
+    if (copy.security == null || copy.security.length === 0) {
```
JKRhb, line 225: I think one error in the CI is actually caused by this line:

```suggestion
    if (copy?.security.length === 0) {
```

danielpeintner: @JKRhb Can you explain why this should cause issues? Locally the tests are all "green".
I even think the change you are proposing has a different meaning.

The current code 
`    if (!copy.security || copy.security.length === 0) {`
enters the if branch IF either `copy.security` is `null` or `undefined ` but also IF the security array is of `length === 0`

Your code misses the first part, doesn't it ?

JKRhb: Hmm, yeah, you are right, I was mistaken with my suggestion, I thought the logic got reversed, but it should still be right. What got me to make the suggestion is that there is currently this error reported in the CI:

```
  1) the server side of servient
       should be able to add a Thing based on WoT.ThingFragment:
     TypeError: Cannot read properties of undefined (reading 'length')
      at Object.serializeTD (/home/runner/work/node-wot/node-wot/packages/td-tools/src/td-parser.ts:232:21)
      at ExposedThing.getThingDescription (src/exposed-thing.ts:126:30)
      at WoTServerTest.<anonymous> (test/ServerTest.ts:109:22)
      at fulfilled (test/ServerTest.ts:11:58)

```

However, I somehow made the suggestion to the wrong line. I would resolve this thread and try again 😅

packages/td-tools/src/td-parser.ts
```diff
@@ -222,20 +222,20 @@ export function serializeTD(thing: Thing): string {
...
         copy.securityDefinitions = {
             nosec_sc: { scheme: "nosec" },
         };
         copy.security = ["nosec_sc"];
     }
 
-    if (copy.forms && copy.forms.length === 0) {
+    if (copy?.forms.length === 0) {
```
JKRhb, line 232: After also testing locally, I now found out that the real issue is in this line :)

```suggestion
    if (copy.forms?.length === 0) {
```

danielpeintner: Yes, that makes sense! Sorry for not seeing it myself.. I still wonder why it works locally on my side!?

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -138,7 +138,7 @@ export class ThingModelHelpers {
     public static getModelVersion(data: ThingModel): string | undefined {
         if (
             "version" in data &&
-            data.version &&
+            data.version != null &&
             typeof data.version === "object" &&
             "model" in data.version &&
             typeof data.version.model === "string"
```
JKRhb, line 144: Maybe as a last-minute suggestion: I think these checks could be replaced with `typeof data?.version?.model === "string"` (or alternatively the whole method with `return data?.version?.model`).

JKRhb: I see, thank you for making the adjustment! :)

JKRhb (approved):

relu91 (approved):
Greenlight from my side too!

PR #1078, opened 2023-09-11
feat(coap-server): add support for URI variables

### LINKED ISSUE #1015: URI Variables in CoAP are not supported
[Smart coffee machine Thing](http://plugfest.thingweb.io:8083/smart-coffee-machine) produces CoAP forms even if there are uri variables for that affordance. But there is no support for that in CoAP so those forms are not usable at all. Either the support for URI variables should be implemented or forms should not be generated for coap in that case.

JKRhb:
Resolves #1015.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -189,7 +189,9 @@ export default class CoapServer implements ProtocolServer {
                 this.PROPERTY_DIR,
                 propertyName,
                 offeredMediaType,
-                opValues
+                opValues,
+                property.uriVariables,
+                thing.uriVariables
             );
 
             property.forms.push(form);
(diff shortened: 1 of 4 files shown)


egekorkan (approved):

JKRhb:
Thank you for your approval, @egekorkan :) In the last two commits, I applied some additional refactoring/simplification, the functionality itself should remain unchanged, though.

danielpeintner (approved):
LGTM

JKRhb:
> LGTM

Thank you! Is it okay for you to squash the fixup commits and then merge the PR? Or would you prefer the squash + merge approach?

danielpeintner:
I would "squash and merge" once we merge the PR.
I like this way best since it keeps the history of the PR till to the end and the final commit is clean when it gets merged into the master...

PR #1079, opened 2023-09-11
fix(binding-netconf): use optional chaining for non-null-assertion

JKRhb:
While working on #1078, I noticed that there is a warning emitted by eslint regarding the use of `!` for non-null-assertions. This PR simply replaces the `!`s with `?` (for optional chaining) which should avoid unwanted errors during runtime if one of the parameters should be undefined.


danielpeintner (approved):

JKRhb:
Given @danielpeintner's approval, I would merge this PR :)

PR #1080, opened 2023-09-12
chore: increase version numbers to 0.8.9 for next release

danielpeintner:


JKRhb (approved):
Looks good to me in general :) However, there are a lot of changes to package-lock.json file for some reason – I suppose these are caused by updates to packages within the semantic versioning changes, right?

danielpeintner:
> However, there are a lot of changes to package-lock.json file for some reason – I suppose these are caused by updates to packages within the semantic versioning ranges, right?

Yes, once we prepare for a new release we update the dependencies too..

see https://github.com/eclipse-thingweb/node-wot#development-internals
image


To be honest, I think it is good to update the dependencies from time to time..
Do you think this should be done differently?

relu91 (approved):

PR #1082, opened 2023-09-13
feat!(coap-server): adjust constructor to HTTP server

### LINKED ISSUE #1028: Discrepency between HTTP and CoAP Server Configuration
When creating an HTTP server, the constructor expects an object with parameters like port, e.g. `new HttpServer({port:8081})`. However, CoAP expects each parameter as an argument, e.g. `new CoapServer(port=5684)`. In the CLI, everything is an object as expected. I think that two configurations should be aligned and ideally an object makes more sense so that it looks similar to CLI configuration.

Note: When I provide an object to CoAP constructor, there are no complaints and the default port is used. For the HTTP, if I provide an argument that is not an object, I get an error.

JKRhb:
As pointed out by @egekorkan in #1028, the CoAP server currently has a different constructor API that differs from its HTTP counterpart.

This PR introduces a `CoapServerConfig` interface that features the previously defined `address` and `port` parameters in order to align it with the HTTP implementation (and probably also other binding implementations). Using this approach will also make it possible to extend the CoAP server more easily in the feature (e.g., with a security configuration).

As this PR introduces a breaking change, however, I suppose merging it should be postponed until the next minor(?) release.

Will resolve #1028.

### DIFF
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -70,7 +70,7 @@ export default class DefaultServient extends Servient {
         },
         http: {
             port: 8080,
-            selfSigned: false,
+            allowSelfSigned: false,
         },
         coap: {
             port: 5683,
@@ -131,11 +131,9 @@ export default class DefaultServient extends Servient {
                 // re-use httpServer (same port)
                 // this.addServer(new WebSocketServer(httpServer));
             }
-            if (this.config.coap) {
-                coapServer =
-                    typeof this.config.coap.port === "number"
-                        ? new CoapServer(this.config.coap.port)
-                        : new CoapServer();
+            const coapConfig = this.config.coap;
+            if (coapConfig != null) {
+                coapServer = new CoapServer(coapConfig);
                 this.addServer(coapServer);
             }
             if (this.config.mqtt) {
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -33,6 +33,7 @@ import slugify from "slugify";
 import { Readable } from "stream";
 import { MdnsIntroducer } from "./mdns-introducer";
 import { PropertyElement, DataSchema } from "wot-thing-description-types";
+import { CoapServerConfig } from "./coap";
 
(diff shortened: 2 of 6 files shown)


relu91 (approved):
LGTM

JKRhb:
> BTW CI is failing, you have to update also the examples :)

Should be fixed now :)

egekorkan:
A bit of a dangerous question but why aren't the cli package tests (and build) are failing? This effects the following:
- https://github.com/eclipse-thingweb/node-wot/blob/master/packages/examples/src/quickstart/smart-clock.ts
- https://github.com/eclipse-thingweb/node-wot/blob/master/packages/cli/src/cli-default-servient.ts#L137
- https://github.com/eclipse-thingweb/node-wot/blob/master/API.md?plain=1#L41C15-L41C15 (fine for now)
- https://github.com/eclipse-thingweb/node-wot/blob/master/examples/templates/exposed-thing/index.js (funny enough that this was actually fortune teller)


So in other words, we need to reflect this in other places

JKRhb:
Hmm, I guess the build for https://github.com/eclipse-thingweb/node-wot/blob/master/packages/cli/src/cli-default-servient.ts#L137 isn't failing since the corresponding method parameter is typed as `any`,

JKRhb:
The issue with the CLI package should now be addressed – however, the typing in that package seems to need some more work.

packages/cli/src/cli-default-servient.ts
```diff
@@ -70,7 +70,7 @@ export default class DefaultServient extends Servient {
         },
         http: {
             port: 8080,
-            selfSigned: false,
+            allowSelfSigned: false,
```
JKRhb, line 73: While I was at it, I corrected the key used here to actually correspond with the `HttpConfig` interface.

danielpeintner:
> As this PR introduces a breaking change, however, I suppose merging it should be postponed until the next minor(?) release.

We recently did have a _breaking_ change in `binding-http` as well in https://github.com/eclipse-thingweb/node-wot/pull/1070.
I think it is fine to go on...

Anyone opposed to it?

danielpeintner (approved):

PR #1083, opened 2023-09-14
Data Schema Fix for Quickstart example

egekorkan:
As discovered in W3C TPAC, we should fix the TD Data Schema otherwise error is returned with property read

### DIFF
--- a/examples/quickstart/README.md
+++ b/examples/quickstart/README.md
@@ -1,0 +1,10 @@
+# Quick Start Things
+
+This set of Things can be used to build mashup applications for a smart home scenario.
+These Things are:
+
+-   Simple coffee machine that can take an order to brew a coffee.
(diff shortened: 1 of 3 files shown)


relu91 (approved):

examples/quickstart/README.md
```diff
@@ -0,0 +1,10 @@
+# Quick Start Things
+
```
danielpeintner, line 2: Does it make sense to link to the possible consumer part? 
https://github.com/w3c/wot-cg/tree/main/Tutorials/quickstart

danielpeintner (approved):
LGTM

A note below...

examples/quickstart/README.md
egekorkan: I want to create it in ts and submit here but as a separate PR

Issue #1014, opened 2023-06-09
Meta operations in CoAP


egekorkan:
In comparison to HTTP, Thing-level forms with CoAP do not have meta-level operations (see [counter TD](http://plugfest.thingweb.io:8083/counter))

JKRhb:
Should we maybe turn this into a checklist or task list for all of the available meta operations? (Since a few have been implemented by #1086 now).

danielpeintner:
> Should we maybe turn this into a checklist or task list for all of the available meta operations?

Yes, that makes a lot of sense and provides us a good overview

PR #1084, opened 2023-09-14
refactor(coap-server): simplify form generation

JKRhb:
While preparing to implement #1014, I noticed that there is actually some room for simplification in the current CoAP server implementation. This PR makes the corresponding changes, getting rid of two unnecessary helper functions I previously introduced.


JKRhb:
> LGTM, just one question. You removed the port parameter because it was only used to log the Href Assignment and it could have been retrieved using `this.port` right?

Exactly :)

JKRhb:
Given @relu91's approval, I would merge this one, hope that's fine :)

PR #1086, opened 2023-09-15
feat(coap-server): add initial meta operation support

JKRhb:
Based on the refactoring made in #1084, this PR adds initial support for meta operations, namely the reading of multiple or all properties, to the CoAP server. This corresponds with issue #1014 – as only one kind of meta operation is supported yet, I think the issue can be kept open for now.

I haven't figured out the correct way to implement writing multiple properties yet, which is why I left it out of this PR for now. However, a method skeleton already exists (at the moment, it simply returns an error response with code `5.01` (Not Implemented)) which can be expanded in a follow-up PR.

As this PR depends on #1084, I will rebase once that PR has been merged to make reviewing a bit easier. Until then, I will keep this PR in draft status.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -34,6 +34,7 @@ import { Readable } from "stream";
 import { MdnsIntroducer } from "./mdns-introducer";
 import { PropertyElement, DataSchema } from "wot-thing-description-types";
 import { CoapServerConfig } from "./coap";
+import { DataSchemaValue } from "wot-typescript-definitions";
 
 const { debug, warn, info, error } = createLoggers("binding-coap", "coap-server");
 
@@ -171,6 +172,8 @@ export default class CoapServer implements ProtocolServer {
             for (const offeredMediaType of offeredMediaTypes) {
                 const base = this.createThingBase(address, port, urlPath);
 
+                this.fillInMetaPropertiesBindingData(thing, base, offeredMediaType);
+
                 this.fillInPropertyBindingData(thing, base, offeredMediaType);
                 this.fillInActionBindingData(thing, base, offeredMediaType);
                 this.fillInEventBindingData(thing, base, offeredMediaType);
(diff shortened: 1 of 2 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -182,17 +185,73 @@ export default class CoapServer implements ProtocolServer {
...
+            opValues.push("writemultipleproperties");
+        }
+
+        if (writableProperties === numberOfProperties) {
+            opValues.push("writeallproperties");
+        }
+
+        return opValues;
+    }
+
```
JKRhb, line 243: Note that these helper functions could also be moved to the core package (later?) to help with achieving unified behavior across binding implementations.

JKRhb:
> However, wouldn't it be better to never generate a TD with it?

Good point, I removed it for now alongside the `handleWriteMultipleProperties` method in order to not confuse people. However, note that in the HTTP server, the write meta operations are actually added to the `op` array despite having no corresponding handler method defined:

https://github.com/eclipse-thingweb/node-wot/blob/227be532bc98f5f6c8aeb4abc9de07cccca0ce71/packages/binding-http/src/http-server.ts#L346-L366

packages/binding-coap/src/coap-server.ts
```diff
@@ -182,17 +185,64 @@ export default class CoapServer implements ProtocolServer {
...
+        const opValues: string[] = [];
+
+        if (readableProperties > 0) {
+            opValues.push("readmultipleproperties");
+        }
+
+        if (readableProperties === numberOfProperties) {
+            opValues.push("readallproperties");
+        }
```
relu91, line 230: Actually, the spec is not clear about this semantic. One could still understand that the readallproperties operation just works on the set of readable properties of the thing. I guess for the time being is alright to leave as it is, but still other bindings implementations might act differently.

relu91 (approved):
Just a small comment below. Test coverage could be also improved but we can do it in a future PR.

danielpeintner (approved):

PR #1088, opened 2023-09-18
chore(binding-coap): enable strict-boolean-expressions and null checks

JKRhb:
This PR addresses one of the checkmarks in #1046.

### DIFF
--- a/packages/binding-coap/.eslintrc.json
+++ b/packages/binding-coap/.eslintrc.json
@@ -1,3 +1,6 @@
 {
-    "extends": "../../.eslintrc.js"
+    "extends": "../../.eslintrc.js",
+    "rules": {
+        "@typescript-eslint/strict-boolean-expressions": ["error"]
+    }
 }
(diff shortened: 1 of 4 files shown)


danielpeintner (approved):

JKRhb:
> I still didn't have proper time to review this, I want to be sure that is sufficient for check only for `null` and not for `undefined`. Given that in TS/JS world is way more common to leave values into the undefined state, I would like to be careful to only check for `null`

`foo != null` is basically a shorthand for `foo !== null && foo !== undefined` – as mentioned elsewhere, this is apparently the only case where using the `==` operator is valid. So when checking for `null` this way, you get the check for `undefined` for free and vice versa. (Having both `null` and `undefined` has obviously not been a good language design decision 😅).

I haven't found a really good documentation link to point here, but some further explanations can be found here: https://basarat.gitbook.io/typescript/recap/null-undefined#checking-for-either

relu91 (approved):
Silly me, I'm so used to the `!==` operator that I didn't even notice that you used the less strict version. Good to go.

PR #1093, opened 2023-09-21
docs(contributing): add a guide for checking of `undefined` or `null`

relu91:
Following [this comment](https://github.com/eclipse-thingweb/node-wot/pull/1088#issuecomment-1727706226), I thought it would have been good to have some references for future contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,17 +94,17 @@ How then should the contributor deal with nullable variables? For example:
 ```ts
 function(arg1: string | null | undefined) {
     // ERROR: not allowed by strict-boolean-expressions
-    if(!arg) { throw new Error("arg should be defined!))}
+    if (!arg) { throw new Error("arg should be defined!); }
 }
 ```
 
-Instead of checking for both null and `undefiend` values (`if(arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
-with ==, null and undefined are only equal to each other. Example:
+Instead of checking for both null and `undefined` values (`if (arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
+with `==`, `null` and `undefined` are only equal to each other. Example:
 
 ```ts
 function(arg1: string | null | undefined) {
     // OK
-    if(arg == null) { throw new Error("arg should be defined!))}
+    if (arg == null) { throw new Error("arg should be defined!); }
 }
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,27 @@ function startFoo() {
     error("This is an error message!");
 }
 ```
+### Checking for `undefined` or `null`
+
+In node-wot, we enabled [strict boolean expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/). In summary, this means that in the 
+the code base is not allowed to use non-boolean expressions where a boolean is expected (see the [examples](https://typescript-eslint.io/rules/strict-boolean-expressions/#examples)). 
+How then should the contributor deal with nullable variables? For example:
+```ts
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -85,6 +85,31 @@ function startFoo() {
...
+
+```ts
+function(arg1: string | null | undefined) {
+    // ERROR: not allowed by strict-boolean-expressions
+    if(!arg) { throw new Error("arg should be defined!))}
+}
+```
+
+Instead of checking for both null and `undefiend` values (`if(arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
+with ==, null and undefined are only equal to each other. Example:
```
JKRhb, line 102:
```suggestion
Instead of checking for both null and `undefined` values (`if (arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
with `==`, `null` and `undefined` are only equal to each other. Example:
```

CONTRIBUTING.md
```diff
@@ -85,6 +85,31 @@ function startFoo() {
...
+### Checking for `undefined` or `null`
+
+In node-wot, we enabled [strict boolean expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/). In summary, this means that in the
+the code base is not allowed to use non-boolean expressions where a boolean is expected (see the [examples](https://typescript-eslint.io/rules/strict-boolean-expressions/#examples)).
+How then should the contributor deal with nullable variables? For example:
+
+```ts
+function(arg1: string | null | undefined) {
+    // ERROR: not allowed by strict-boolean-expressions
+    if(!arg) { throw new Error("arg should be defined!))}
```
JKRhb, line 97:
```suggestion
    if (!arg) { throw new Error("arg should be defined!); }
```

CONTRIBUTING.md
```diff
@@ -85,6 +85,31 @@ function startFoo() {
...
+}
+```
+
+Instead of checking for both null and `undefiend` values (`if(arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
+with ==, null and undefined are only equal to each other. Example:
+
+```ts
+function(arg1: string | null | undefined) {
+    // OK
+    if(arg == null) { throw new Error("arg should be defined!))}
```
JKRhb, line 107:
```suggestion
    if (arg == null) { throw new Error("arg should be defined!); }
```

JKRhb (approved):
See just a couple of nits below, otherwise this looks very good and should be really helpful for contributors :)

CONTRIBUTING.md
relu91:
```suggestion
with ==, null and undefined are equal to each other. Example:
```

CONTRIBUTING.md
```diff
@@ -85,6 +85,31 @@ function startFoo() {
...
+
+```ts
+function(arg1: string | null | undefined) {
+    // ERROR: not allowed by strict-boolean-expressions
+    if (!arg) { throw new Error("arg should be defined!); }
+}
+```
+
+Instead of checking for both null and `undefined` values (`if (arg !== undefined && arg !== null)`) the preferred solution is to use `!=` or `==` operator. Interestingly in JavaScript
+with `==`, `null` and `undefined` are only equal to each other. Example:
```
relu91, line 102:
```suggestion
with `==`, `null` and `undefined` are equal to each other. Example:
```

@danielpeintner @JKRhb is that fine?

danielpeintner: Yes 👍

danielpeintner (approved):
PR looks good!

I just wonder whether we should wait till we merge since we just checked 2 (out of the many) checkboxes for https://github.com/eclipse-thingweb/node-wot/issues/1046

relu91:
I'll merge as guide for future PR regarding the issue #1046

PR #1094, opened 2023-09-21
test(binding-coap): `await` all calls of `coapServer.stop()`

JKRhb:
In the context of #1086 and #1077, I noticed that in the CoAP tests, not all invocations of `coapServer.stop()` are actually explicitly being awaited which might cause the CI to hang if there is an unexpected error.

This PR makes a couple of minor adjustments to improve the cleanup of resources and fix this potential problem (although I am not sure to which extent it is actually a problem to begin).

### DIFF
--- a/packages/binding-coap/test/coap-server-test.ts
+++ b/packages/binding-coap/test/coap-server-test.ts
@@ -336,29 +336,32 @@ class CoapServerTest {
             null,
         ];
 
-        const promises = contentFormats.map((contentFormat) => new Promise<void>((resolve) => {
-            const req = request(uri);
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):

relu91 (approved):

relu91:
> As I remembered that using both EventEmitter and async can be a bit problematic, I now wrapped the test code in question into Promises, which also had the benefit of slightly better performance, if I am not mistaken :) If your approval stands, then we could merge this one :)

Just saw [this video](https://www.youtube.com/watch?v=Ra7Ji9LmG9o) yesterday 😄

PR #1095, opened 2023-09-22
chore(examples): enable eslint/strict-boolean-expressions and strictNullChecks

danielpeintner:

### DIFF
--- a/examples/scripts/countdown.js
+++ b/examples/scripts/countdown.js
@@ -80,7 +80,7 @@ WoT.produce({
                 const listToDelete = [];
                 for (const id of countdowns.keys()) {
                     const as = countdowns.get(id);
-                    if (as !== undefined && as.output !== undefined) {
+                    if ((as === null || as === void 0 ? void 0 : as.output) !== undefined) {
                         const prev = as.output;
                         as.output--;
                         console.log("\t" + id + ", from " + prev + " to " + as.output);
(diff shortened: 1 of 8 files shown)


examples/scripts/countdown.js
```diff
@@ -80,7 +80,7 @@ WoT.produce({
                 const listToDelete = [];
                 for (const id of countdowns.keys()) {
                     const as = countdowns.get(id);
-                    if (as !== undefined && as.output !== undefined) {
+                    if ((as === null || as === void 0 ? void 0 : as.output) !== undefined) {
```
danielpeintner, line 83: Interesting to see what the tool-chain makes out of
`if (as?.output !== undefined) {`

Note: This is not a change in this PR.. we might have forgotten to update the JS in a previous PR..

packages/examples/src/scripts/smart-coffee-machine.ts
```diff
@@ -396,7 +396,7 @@ Assumes one medium americano if not specified, but time and mode are mandatory f
             const paramsp = (await params.value()) as Record<string, unknown>; //  : any = await Helpers.parseInteractionOutput(params);
 
             // Check if uriVariables are provided
-            if (paramsp && typeof paramsp === "object" && "time" in paramsp && "mode" in paramsp) {
+            if (paramsp != null && typeof paramsp === "object" && "time" in paramsp && "mode" in paramsp) {
```
danielpeintner, line 399: Note: What is also "interesting" here is that we cannot simply remove the 
`paramsp != null`
check.

the following code 
```
const paramsp = null;
if (typeof paramsp === "object") {
    console.log("null is object")
}
```

prints `null is object`

relu91 (approved):
Sorry, there we go! lgtm.

PR #1096, opened 2023-09-22
chore(core): enable eslint/strict-boolean-expressions & strictNullChecks

danielpeintner:

### DIFF
--- a/packages/core/src/codecs/json-codec.ts
+++ b/packages/core/src/codecs/json-codec.ts
@@ -24,7 +24,7 @@ export default class JsonCodec implements ContentCodec {
     private subMediaType: string;
 
     constructor(subMediaType?: string) {
-        if (!subMediaType) {
+        if (subMediaType == null) {
             this.subMediaType = ContentSerdes.DEFAULT; // 'application/json'
         } else {
             this.subMediaType = subMediaType;
--- a/packages/core/src/codecs/json-codec.ts
+++ b/packages/core/src/codecs/json-codec.ts
@@ -24,11 +24,7 @@ export default class JsonCodec implements ContentCodec {
     private subMediaType: string;
 
     constructor(subMediaType?: string) {
-        if (subMediaType == null) {
-            this.subMediaType = ContentSerdes.DEFAULT; // 'application/json'
-        } else {
-            this.subMediaType = subMediaType;
-        }
+        this.subMediaType = subMediaType ?? ContentSerdes.DEFAULT;
     }
 
     getMediaType(): string {
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -297,9 +297,7 @@ class InternalEventSubscription extends InternalSubscription {
             throw new Error(`ConsumedThing '${this.thing.title}' does not have event ${this.name}`);
         }
 
-        if (options.formIndex == null) {
-            options.formIndex = this.matchingUnsubscribeForm();
-        }
+        options.formIndex ??= this.matchingUnsubscribeForm();
 
         const { form } = this.thing.getClientFor(te.forms, "unsubscribeevent", Affordance.EventAffordance, options);
         if (form == null) {
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -166,10 +166,10 @@ class InternalPropertySubscription extends InternalSubscription {
 
     public async unobserveProperty(options: WoT.InteractionOptions = {}): Promise<void> {
         const tp = this.thing.properties[this.name];
-        if (!tp) {
+        if (tp == null) {
             throw new Error(`ConsumedThing '${this.thing.title}' does not have property ${this.name}`);
         }
-        if (!options.formIndex) {
+        if (options.formIndex == null) {
             options.formIndex = this.matchingUnsubscribeForm();
         }
         const { form } = this.thing.getClientFor(tp.forms, "unobserveproperty", Affordance.PropertyAffordance, options);
@@ -295,11 +295,11 @@ class InternalEventSubscription extends InternalSubscription {
 
     public async unsubscribeEvent(options: WoT.InteractionOptions = {}): Promise<void> {
         const te = this.thing.events[this.name];
-        if (!te) {
+        if (te == null) {
             throw new Error(`ConsumedThing '${this.thing.title}' does not have event ${this.name}`);
         }
 
-        if (!options.formIndex) {
+        if (options.formIndex == null) {
             options.formIndex = this.matchingUnsubscribeForm();
         }
 
@@ -317,7 +317,7 @@ class InternalEventSubscription extends InternalSubscription {
     private matchingUnsubscribeForm(): number {
         const refForm = this.thing.events[this.name].forms[this.formIndex];
         // Here we have to keep in mind that op default is ["subscribeevent", "unsubscribeevent"]
-        if (!refForm.op || (Array.isArray(refForm.op) && refForm.op.includes("unsubscribeevent"))) {
+        if (refForm.op == null || (Array.isArray(refForm.op) && refForm.op.includes("unsubscribeevent"))) {
             // we can re-use the same form for unsubscribe
             return this.formIndex;
         }
@@ -450,7 +450,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         for (const s of security) {
             const ws = this.securityDefinitions[s + ""]; // String vs. string (fix wot-typescript-definitions?)
             // also push nosec in case of proxy
-            if (ws) {
+            if (ws != null) {
                 scs.push(ws);
             }
         }
@@ -458,7 +458,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
     }
 
     ensureClientSecurity(client: ProtocolClient, form: TD.Form | undefined): void {
-        if (this.securityDefinitions) {
+        if (this.securityDefinitions != null) {
             const logStatement = () =>
                 debug(`ConsumedThing '${this.title}' setting credentials for ${client} based on thing security`);
 
@@ -471,7 +471,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
                     this.getSecuritySchemes(form.security),
                     this.getServient().retrieveCredentials(this.id)
                 );
-            } else if (this.security && Array.isArray(this.security) && this.security.length > 0) {
+            } else if (this.security != null && Array.isArray(this.security) && this.security.length > 0) {
                 logStatement();
                 client.setSecurity(
                     this.getSecuritySchemes(this.security as string[]),
@@ -551,7 +551,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
     async readProperty(propertyName: string, options?: WoT.InteractionOptions): Promise<WoT.InteractionOutput> {
         // TODO pass expected form op to getClientFor()
         const tp = this.properties[propertyName];
-        if (!tp) {
+        if (tp == null) {
             throw new Error(`ConsumedThing '${this.title}' does not have property ${propertyName}`);
         }
 
@@ -559,7 +559,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         if (!form) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable form`);
         }
-        if (!client) {
+        if (client == null) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`);
         }
         debug(`ConsumedThing '${this.title}' reading ${form.href}`);
@@ -618,14 +618,14 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
     ): Promise<void> {
         // TODO pass expected form op to getClientFor()
         const tp = this.properties[propertyName];
-        if (!tp) {
+        if (tp == null) {
             throw new Error(`ConsumedThing '${this.title}' does not have property ${propertyName}`);
         }
         let { client, form } = this.getClientFor(tp.forms, "writeproperty", Affordance.PropertyAffordance, options);
         if (!form) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable form`);
         }
-        if (!client) {
+        if (client == null) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`);
         }
         debug(`ConsumedThing '${this.title}' writing ${form.href} with '${value}'`);
@@ -661,14 +661,14 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         options?: WoT.InteractionOptions
     ): Promise<WoT.InteractionOutput> {
         const ta = this.actions[actionName];
-        if (!ta) {
+        if (ta == null) {
             throw new Error(`ConsumedThing '${this.title}' does not have action ${actionName}`);
         }
         let { client, form } = this.getClientFor(ta.forms, "invokeaction", Affordance.ActionAffordance, options);
         if (!form) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable form`);
         }
-        if (!client) {
+        if (client == null) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`);
         }
         debug(
@@ -714,14 +714,14 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         options?: WoT.InteractionOptions
     ): Promise<Subscription> {
         const tp = this.properties[name];
-        if (!tp) {
+        if (tp == null) {
             throw new Error(`ConsumedThing '${this.title}' does not have property ${name}`);
         }
         const { client, form } = this.getClientFor(tp.forms, "observeproperty", Affordance.PropertyAffordance, options);
         if (!form) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable form`);
         }
-        if (!client) {
+        if (client == null) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`);
         }
         if (this.observedProperties.has(name)) {
@@ -772,14 +772,14 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         options?: WoT.InteractionOptions
     ): Promise<Subscription> {
         const te = this.events[name];
-        if (!te) {
+        if (te == null) {
             throw new Error(`ConsumedThing '${this.title}' does not have event ${name}`);
         }
         const { client, form } = this.getClientFor(te.forms, "subscribeevent", Affordance.EventAffordance, options);
         if (!form) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable form`);
         }
-        if (!client) {
+        if (client == null) {
             throw new Error(`ConsumedThing '${this.title}' did not get suitable client for ${form.href}`);
         }
         if (this.subscribedEvents.has(name)) {
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -173,7 +173,7 @@ class InternalPropertySubscription extends InternalSubscription {
             options.formIndex = this.matchingUnsubscribeForm();
         }
         const { form } = this.thing.getClientFor(tp.forms, "unobserveproperty", Affordance.PropertyAffordance, options);
-        if (!form) {
+        if (form == null) {
             throw new Error(`ConsumedThing '${this.thing.title}' did not get suitable form`);
         }
 
@@ -219,7 +219,7 @@ class InternalPropertySubscription extends InternalSubscription {
         for (let i = 0; i < forms.length; i++) {
             let score = 0;
             const form = forms[i];
-            if (form.op === operation || (Array.isArray(form.op) && form.op.includes(operation))) {
+            if (form.op === operation || (form?.op?.includes(operation) === true && Array.isArray(form.op) === true)) {
                 score += 1;
             }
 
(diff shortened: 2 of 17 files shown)


packages/core/src/consumed-thing.ts
```diff
@@ -166,10 +166,10 @@ class InternalPropertySubscription extends InternalSubscription {
...
     public async unobserveProperty(options: WoT.InteractionOptions = {}): Promise<void> {
         const tp = this.thing.properties[this.name];
-        if (!tp) {
+        if (tp == null) {
             throw new Error(`ConsumedThing '${this.thing.title}' does not have property ${this.name}`);
         }
-        if (!options.formIndex) {
+        if (options.formIndex == null) {
             options.formIndex = this.matchingUnsubscribeForm();
         }
```
JKRhb, line 174:
```suggestion
        options.formIndex ??= this.matchingUnsubscribeForm();
```

packages/core/src/consumed-thing.ts
```diff
@@ -295,11 +295,11 @@ class InternalEventSubscription extends InternalSubscription {
...
         const te = this.thing.events[this.name];
-        if (!te) {
+        if (te == null) {
             throw new Error(`ConsumedThing '${this.thing.title}' does not have event ${this.name}`);
         }
 
-        if (!options.formIndex) {
+        if (options.formIndex == null) {
             options.formIndex = this.matchingUnsubscribeForm();
         }
```
JKRhb, line 304:
```suggestion
        options.formIndex ??= this.matchingUnsubscribeForm();
```

packages/core/src/codecs/json-codec.ts
```diff
@@ -24,7 +24,7 @@ export default class JsonCodec implements ContentCodec {
     private subMediaType: string;
 
     constructor(subMediaType?: string) {
-        if (!subMediaType) {
+        if (subMediaType == null) {
             this.subMediaType = ContentSerdes.DEFAULT; // 'application/json'
         } else {
             this.subMediaType = subMediaType;
```
JKRhb, line 30: Could be simplified to

```ts
this.subMediaType = subMediaType ?? ContentSerdes.DEFAULT;
```

packages/core/src/exposed-thing.ts
```diff
@@ -251,8 +251,8 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
     setPropertyObserveHandler(name: string, handler: WoT.PropertyReadHandler): WoT.ExposedThing {
         debug(`ExposedThing '${this.title}' setting property observe handler for '${name}'`);
 
-        if (this.properties[name]) {
-            if (!this.properties[name].observable) {
+        if (this.properties[name] != null) {
+            if (!(this.properties[name].observable === true)) {
```
JKRhb, line 255:
```suggestion
            if (this.properties[name].observable !== true) {
```

packages/core/src/exposed-thing.ts
```diff
@@ -275,8 +275,8 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
     setPropertyUnobserveHandler(name: string, handler: WoT.PropertyReadHandler): WoT.ExposedThing {
         debug(`ExposedThing '${this.title}' setting property unobserve handler for '${name}'`);
 
-        if (this.properties[name]) {
-            if (!this.properties[name].observable) {
+        if (this.properties[name] != null) {
+            if (!(this.properties[name].observable === true)) {
```
JKRhb, line 279:
```suggestion
            if (this.properties[name].observable !== true) {
```

packages/core/src/exposed-thing.ts
```diff
@@ -398,9 +399,10 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
                 debug(`ExposedThing '${this.title}' calls registered readHandler for Property '${propertyName}'`);
                 Helpers.validateInteractionOptions(this, this.properties[propertyName], options);
                 const result: WoT.InteractionInput | void = await readHandler(options);
-                const form = this.properties[propertyName].forms
-                    ? this.properties[propertyName].forms[options.formIndex]
-                    : { contentType: "application/json" };
+                const form =
+                    this.properties[propertyName].forms != null
+                        ? this.properties[propertyName].forms[options.formIndex]
+                        : { contentType: "application/json" };
```
JKRhb, line 405:
```suggestion
                const form = this.properties[propertyName]?.forms[options.formIndex] ?? { contentType: "application/json" };
```

packages/core/src/exposed-thing.ts
```diff
@@ -428,7 +430,7 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
             for (const propertyName of propertyNames) {
                 // Note: currently only JSON DataSchema properties are supported
                 const form = this.properties[propertyName].forms.find(
-                    (form) => form.contentType === ContentSerdes.DEFAULT || !form.contentType
+                    (form) => form.contentType === ContentSerdes.DEFAULT || form.contentType == null
```
JKRhb, line 433:
```suggestion
                    (form) => form?.contentType === ContentSerdes.DEFAULT ?? true
```

packages/core/src/exposed-thing.ts
```diff
@@ -477,15 +479,16 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
             }
             Helpers.validateInteractionOptions(this, this.properties[propertyName], options);
             const writeHandler = this.__propertyHandlers.get(propertyName)?.writeHandler;
-            const form = this.properties[propertyName].forms
-                ? this.properties[propertyName].forms[options.formIndex]
-                : {};
+            const form =
+                this.properties[propertyName].forms != null
+                    ? this.properties[propertyName].forms[options.formIndex]
+                    : {};
```
JKRhb, line 491:
```suggestion
            const form = this.properties[propertyName]?.forms[options.formIndex] ?? {};
```

packages/core/src/exposed-thing.ts
```diff
@@ -510,7 +513,7 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
         for (const propertyName in valueMap) {
             // Note: currently only DataSchema properties are supported
             const form = this.properties[propertyName].forms.find(
-                (form) => form.contentType === "application/json" || !form.contentType
+                (form) => form.contentType === "application/json" || form.contentType == null
```
JKRhb, line 516:
```suggestion
                (form) => form?.contentType === "application/json" ?? true
```

packages/core/src/helpers.ts
```diff
@@ -351,7 +351,7 @@ export default class Helpers implements Resolver {
         uriVariables: { [k: string]: DataSchema } = {}
     ): Record<string, unknown> {
         const params: Record<string, unknown> = {};
-        if ((url === undefined && url == null) || (!uriVariables && !globalUriVariables)) {
+        if ((url === undefined && url == null) || (uriVariables == null && globalUriVariables == null)) {
```
JKRhb, line 354:
```suggestion
        if (url == null || (uriVariables == null && globalUriVariables == null)) {
```

packages/core/src/protocol-helpers.ts
```diff
@@ -113,16 +113,15 @@ export default class ProtocolHelpers {
...
+            propertyName != null &&
+            uriScheme != null &&
+            td?.properties != null &&
+            td.properties[propertyName] != null &&
+            td.properties[propertyName].forms != null &&
             Array.isArray(td.properties[propertyName].forms)
         ) {
             for (const form of td.properties[propertyName].forms) {
-                if (form.href && form.href.startsWith(uriScheme) && form.contentType) {
+                if (form.href && form.href.startsWith(uriScheme) && form.contentType != null) {
```
JKRhb, line 124:
```suggestion
                if (form.href?.startsWith(uriScheme) ?? false && form.contentType != null) {
```

packages/core/src/protocol-helpers.ts
```diff
@@ -138,16 +137,15 @@ export default class ProtocolHelpers {
...
-            actionName &&
-            uriScheme &&
-            td.actions &&
-            td.actions[actionName] &&
-            td.actions[actionName].forms &&
+            actionName != null &&
+            uriScheme != null &&
+            td?.actions != null &&
+            td.actions[actionName] != null &&
+            td.actions[actionName].forms != null &&
```
JKRhb, line 144:
```suggestion
            td?.actions[actionName]?.forms != null &&
```

packages/core/src/protocol-helpers.ts
```diff
@@ -163,16 +161,15 @@ export default class ProtocolHelpers {
...
-            eventName &&
-            uriScheme &&
-            td.events &&
-            td.events[eventName] &&
-            td.events[eventName].forms &&
+            eventName != null &&
+            uriScheme != null &&
+            td?.events != null &&
+            td.events[eventName] != null &&
+            td.events[eventName].forms != null &&
```
JKRhb, line 168:
```suggestion
            uriScheme != null &&
            td?.events[eventName]?.forms != null &&
```

packages/core/src/protocol-helpers.ts
```diff
@@ -240,17 +237,17 @@ export default class ProtocolHelpers {
...
-                        chunks[0] &&
+                        chunks[0] != null &&
                         (chunks[0] instanceof Array || chunks[0] instanceof Buffer || chunks[0] instanceof Uint8Array)
                     ) {
                         resolve(Buffer.concat(chunks as Array<Buffer | Uint8Array>));
-                    } else if (chunks[0] && typeof chunks[0] === "string") {
+                    } else if (chunks[0] != null && typeof chunks[0] === "string") {
                         resolve(Buffer.from(chunks.join()));
                     } else {
                         resolve(Buffer.from(chunks as Array<number>));
```
JKRhb, line 253: Hmm, I am wondering here: What happens if `chunks[0]` is in fact `undefined`? Would that cause an error in the `else` clause?

packages/core/src/wot-impl.ts
```diff
@@ -87,7 +87,7 @@ export default class WoTImpl {
 
             debug(
                 `WoTImpl consuming TD ${
-                    newThing.id ? "'" + newThing.id + "'" : "without id"
+                    newThing.id != null ? "'" + newThing.id + "'" : "without id"
```
JKRhb, line 90:
```suggestion
                    newThing.id != null ? `'${newThing.id}'` : "without id"
```

packages/core/src/protocol-helpers.ts
```diff
@@ -352,7 +349,7 @@ export default class ProtocolHelpers {
         // If a form index hint is gived, you it. Just check the form actually supports the op
         if (interaction.forms !== undefined && formIndex !== undefined && interaction.forms.length > formIndex) {
             const form = interaction.forms[formIndex];
-            if (form && (operationName === undefined || form.op?.includes(operationName))) {
+            if (form != null && (operationName === undefined || form.op?.includes(operationName) === true)) {
```
JKRhb, line 352:
```suggestion
            if (form != null && (operationName == null || form.op?.includes(operationName) === true)) {
```

packages/core/src/exposed-thing.ts
```diff
@@ -355,16 +355,17 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
...
             if (handler != null) {
                 debug(`ExposedThing '${this.title}' calls registered handler for Action '${name}'`);
                 Helpers.validateInteractionOptions(this, this.actions[name], options);
-                const form = this.actions[name].forms
-                    ? this.actions[name].forms[options.formIndex]
-                    : { contentType: "application/json" };
+                const form =
+                    this.actions[name].forms != null
+                        ? this.actions[name].forms[options.formIndex]
+                        : { contentType: "application/json" };
```
JKRhb, line 368:
```suggestion
                const form = this.actions[name].forms[options.formIndex] ?? { contentType: "application/json" };
```

packages/core/src/consumed-thing.ts
```diff
@@ -450,19 +450,19 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
...
     }
 
     ensureClientSecurity(client: ProtocolClient, form: TD.Form | undefined): void {
-        if (this.securityDefinitions) {
+        if (this.securityDefinitions != null) {
             const logStatement = () =>
                 debug(`ConsumedThing '${this.title}' setting credentials for ${client} based on thing security`);
 
-            if (form && Array.isArray(form.security) && form.security.length > 0) {
+            if (form != null && Array.isArray(form.security) && form.security.length > 0) {
```
JKRhb, line 465: I think this if block can be further simplified if a `security` auxiliary variable is introduced like so:

```suggestion
            const security = form?.security;
            if (Array.isArray(security) && security.length > 0) {
```

packages/core/src/consumed-thing.ts
```diff
@@ -471,7 +471,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
                     this.getSecuritySchemes(form.security),
                     this.getServient().retrieveCredentials(this.id)
                 );
-            } else if (this.security && Array.isArray(this.security) && this.security.length > 0) {
+            } else if (this.security != null && Array.isArray(this.security) && this.security.length > 0) {
```
JKRhb, line 474: If I am not mistaken, the `null` check for `this.security` is actually not necessary:

```suggestion
            } else if (Array.isArray(this.security) && this.security.length > 0) {
```

packages/core/src/consumed-thing.ts
```diff
@@ -471,7 +471,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
...
                 logStatement();
                 client.setSecurity(
                     this.getSecuritySchemes(this.security as string[]),
```
JKRhb, line 477: This cast could then also be removed:

```suggestion
                    this.getSecuritySchemes(this.security),
```

packages/core/src/consumed-thing.ts
```diff
@@ -219,7 +219,7 @@ class InternalPropertySubscription extends InternalSubscription {
         for (let i = 0; i < forms.length; i++) {
             let score = 0;
             const form = forms[i];
-            if (form.op === operation || (Array.isArray(form.op) && form.op.includes(operation))) {
+            if (form.op === operation || (form?.op?.includes(operation) === true && Array.isArray(form.op) === true)) {
```
JKRhb, line 222: I think theoretically, the `&& Array.isArray(form.op) === true` part could also be removed here. However, then the check would also return `true` for strings such as `unobservepropertyfoo` – would that be a realistic scenario we should deal with?

JKRhb (approved):
Thank you, looks good to me already :) See a couple of mostly optional suggestions that further simplify a couple of lines below.

packages/core/src/protocol-helpers.ts
JKRhb: In general, I think if you introduce a helper variable above (e.g., `const firstChunk = chunks[0];`) and use it in the if checks (e.g., `typeof firstChunk === "string"`), the `chunks[0] != null` can be removed.

JKRhb: Doing so would yield something like 

```ts
                    const firstChunk = chunks[0];
                    if (
                        firstChunk instanceof Array ||
                        firstChunk instanceof Buffer ||
                        firstChunk instanceof Uint8Array
                    ) {
                        resolve(Buffer.concat(chunks as Array<Buffer | Uint8Array>));
                    } else if (typeof firstChunk === "string") {
                        resolve(Buffer.from(chunks.join()));
                    } else {
                        resolve(Buffer.from(chunks as Array<number>));
                    }
```

I guess the type safety of the method in general should be revisited in the future, though, using only the first element as an indicator for the type of everything will probably work most of the time, but is not rock solid, I suppose.

packages/core/src/codecs/json-codec.ts
danielpeintner: https://github.com/eclipse-thingweb/node-wot/pull/1096/commits/994e1a7861af07dd058cae4c4e410c60051ab647

packages/core/src/exposed-thing.ts
danielpeintner: https://github.com/eclipse-thingweb/node-wot/pull/1096/commits/73fc64272b8bb2c0b57d87a1bdee7e82036e4471

packages/core/src/exposed-thing.ts
danielpeintner: I would rather like to keep the current version 🙃

packages/core/src/exposed-thing.ts
danielpeintner: https://github.com/eclipse-thingweb/node-wot/commit/73fc64272b8bb2c0b57d87a1bdee7e82036e4471

packages/core/src/exposed-thing.ts
JKRhb: Sure, that's fine :)

packages/core/src/exposed-thing.ts
danielpeintner: I also find it better as is..

packages/core/src/protocol-helpers.ts
danielpeintner: Should the following be sufficient!?

`if (form.href?.startsWith(uriScheme) && form.contentType != null) {`

JKRhb: Hmm, I think in this case `form.href?.startsWith(uriScheme)` would have a type of `boolean?` (or rather `boolean | undefined`). How about `form.href?.startsWith(uriScheme) === true` as an alternative?

packages/core/src/protocol-helpers.ts
danielpeintner: I think there are definitively many parts that can (and maybe should) be improved.. but I guess for the sake of avoiding issues I kept the general structure...

packages/core/src/protocol-helpers.ts
```diff
@@ -139,9 +139,7 @@ export default class ProtocolHelpers {
         if (
             actionName != null &&
             uriScheme != null &&
-            td?.actions != null &&
-            td.actions[actionName] != null &&
-            td.actions[actionName].forms != null &&
+            td?.actions[actionName]?.forms != null &&
```
JKRhb, line 142: Hmm, okay, sorry, this seems to have introduced an error

packages/core/src/consumed-thing.ts
danielpeintner: In this case I tend to stick to the existing form ...

packages/core/src/protocol-helpers.ts
danielpeintner: Jupp, I am reverting one step

packages/core/src/protocol-helpers.ts
```diff
@@ -163,16 +159,13 @@ export default class ProtocolHelpers {
...
         if (
-            td &&
-            eventName &&
-            uriScheme &&
-            td.events &&
-            td.events[eventName] &&
-            td.events[eventName].forms &&
+            eventName != null &&
+            uriScheme != null &&
+            td?.events[eventName]?.forms != null &&
```
danielpeintner, line 164: reverting one step

packages/core/src/consumed-thing.ts
danielpeintner: Let's keep it as is..

packages/core/src/protocol-helpers.ts
danielpeintner: What is the difference between (your proposal)
`form.href?.startsWith(uriScheme) === true`
and (mine)
`form.href?.startsWith(uriScheme)`

I am not sure why `=== true` is needed since startsWith returns always a boolean already?

JKRhb: Oh, yeah, you are right, due to the `form.href` in the origin, I thought `href` could be `undefined`, but that obviously cannot be the case :) So I think we should be able to change that to `form.href.startsWith(uriScheme)` (removing the `?`).

packages/core/src/protocol-helpers.ts
```diff
@@ -138,16 +137,13 @@ export default class ProtocolHelpers {
...
-            td &&
-            actionName &&
-            uriScheme &&
-            td.actions &&
-            td.actions[actionName] &&
-            td.actions[actionName].forms &&
+            actionName != null &&
+            uriScheme != null &&
+            td?.actions[actionName]?.forms != null &&
             Array.isArray(td.actions[actionName].forms)
```
JKRhb, line 143: This change should now actually work (tested it locally):

```suggestion
            td.actions != null &&
            Array.isArray(td.actions[actionName]?.forms)
```

packages/core/src/protocol-helpers.ts
JKRhb: Okay, after double-checking locally, this actually works:

```suggestion
                if (form.href.startsWith(uriScheme) && form.contentType != null) {
```

danielpeintner: https://github.com/eclipse-thingweb/node-wot/pull/1096/commits/567fe1ee9446df947d9a15d1f66e9f25da43d7bd

Please speak up if you think it is wrong

JKRhb (approved):

packages/core/src/protocol-helpers.ts
JKRhb: I think the `?` after the `href` could actually be removed, but the current version would also be fine :)

packages/core/src/protocol-helpers.ts
```diff
@@ -163,16 +159,13 @@ export default class ProtocolHelpers {
...
-            td &&
-            eventName &&
-            uriScheme &&
-            td.events &&
-            td.events[eventName] &&
-            td.events[eventName].forms &&
+            eventName != null &&
+            uriScheme != null &&
+            td?.events[eventName]?.forms != null &&
             Array.isArray(td.events[eventName].forms)
```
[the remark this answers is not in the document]
danielpeintner: outdated but I think handled alreay

relu91:
I was about to approve this, but there is a [leftover comment](https://github.com/eclipse-thingweb/node-wot/pull/1096#discussion_r1334387148) from @JKRhb, to a quick glance I would commit it. Is that ok @danielpeintner?

danielpeintner:
> I was about to approve this, but there is a [leftover comment](https://github.com/eclipse-thingweb/node-wot/pull/1096#discussion_r1334387148) from @JKRhb, to a quick glance I would commit it. Is that ok @danielpeintner?

Sorry, I missed that.... I think it is fine now..

relu91 (approved):

Issue #911, opened 2023-01-18
`ExposedThing` initialization process not compliant


relu91:
The Scripting API defines a [Expand Thing Init](https://w3c.github.io/wot-scripting-api/#expand-an-exposedthinginit), but currently, our `ExposedThing` stops at step 3. We should align it with the spec. 

p.s. This causes https://github.com/eclipse/thingweb.node-wot/issues/873, because the HTTP server receives a Thing object without any scheme defined when the CLI module tries to create the Servient Thing using a very simple ExposeThingInit without a security scheme.

relu91:
To be precise we sort of implemented some later steps like 8.1, but we are not generating the title automatically. In general, the whole process needs to be reviewed and aligned.

danielpeintner:
I think we should also create some tests to confirm that we follow all steps for "expand an ExposedThingInit"

relu91:
While solving this issue keep in mind [this thread](https://github.com/eclipse-thingweb/node-wot/pull/1097#discussion_r1341938185) from the PR above.

Issue #558, opened 2021-10-11
Change default merge strategy


relu91:
On PR #519 @erossignon [suggested](https://github.com/eclipse/thingweb.node-wot/pull/519#issuecomment-938353707) changing the default merge strategy from "Allow merge commits" to "Allow rebase merging". Rebasing would allow having a more linear history in the master branch where the current approach is to have merge commits of different PRs. I work with both approaches in different repositories and I have to admit I like both of them.

My take is that the current approach becomes not useful when you have _a lot of_ feature branches (a.k.a. PRs) in a short amount of time. I personally like the fact that you can understand from which PR a change was generated and use this information to look back in Github the discussion history. But of course, when you merge 20/30 Prs a week the thing is not really manageable. But this is not the case yet, looking at node-wot git graph I can still tell what's going on:
![image](https://user-images.githubusercontent.com/3170227/136756637-1220d93f-93ef-472e-8e46-0acb1bb6cb2d.png)

Please share you thoughts about this topic below :)

egekorkan:
There is also this part in the Contributing.md about this : https://github.com/eclipse/thingweb.node-wot/blob/master/CONTRIBUTING.md#pull-requests-and-feature-branches

It is not exactly the same but some might be doing this while others are not. It would be good to say that rebase is required so that we have a single style of PRs? Not a strong opinion though :)

egekorkan:
Also see https://github.com/eclipse/thingweb.node-wot/issues/559

mkovatsc:
The guidelines combined what is used in other Eclipse projects and what we used for inner source.

A squash commit is good, when the PR is limited to a single feature or fix. A review before merging could ensure this. Then it could be as clean and modular as the rebase.

danielpeintner:
@mkovatsc do you mean we should decide on a PR based manner?

mkovatsc:
No, I meant being more strict with PRs (that they only add a single feature / fix a single bug) when changing to squash and merge.

danielpeintner:
I totally agree with your statement. We tried to do that in the past.. and surely failed *sometimes* which gives room to improve.

Anyhow, even if a PR tackled just one aspect but was developed over several commits the history on master does not convey that. Hence, I prefer having a *single* commit that essentially says what was fixed/improved in a given PR and leads to s single entry in the history.

Note: projects like Java, JavaFX etc do that as well.

relu91:
[...]

PR #1097, opened 2023-09-22
chore(binding-http): enable eslint/strict-boolean-expressions

JKRhb:
Ticks one of the boxes of #1046 :)

### DIFF
--- a/packages/binding-http/src/codecs/tuya-codec.ts
+++ b/packages/binding-http/src/codecs/tuya-codec.ts
@@ -17,17 +17,28 @@ import { ContentCodec } from "@node-wot/core";
 import * as TD from "@node-wot/td-tools";
 import { DataSchemaValue } from "wot-typescript-definitions";
 
+interface TuyaOutput {
+    success?: boolean;
+    msg?: string;
+    result?: {
+        code?: string;
+    }[];
+}
+
 export default class HttpTuyaCodec implements ContentCodec {
     getMediaType(): string {
         return "application/json+tuya";
     }
 
     bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
-        const parsedBody = JSON.parse(bytes.toString());
-        if (!parsedBody.success) throw new Error(parsedBody.msg ? parsedBody.msg : JSON.stringify(parsedBody));
-        for (const key in parsedBody.result) {
-            if (parsedBody.result[key].code === schema["tuya:PropertyName"]) {
-                return parsedBody.result[key].value;
+        const parsedBody: TuyaOutput = JSON.parse(bytes.toString());
+        if (parsedBody.success !== true) {
+            throw new Error(parsedBody.msg != null ? parsedBody.msg : JSON.stringify(parsedBody));
+        }
+
+        for (const value of Object.values(parsedBody.result ?? {})) {
+            if (value.code === schema["tuya:PropertyName"]) {
+                return value;
             }
         }
         throw new Error("Property not found");
--- a/packages/binding-http/src/codecs/tuya-codec.ts
+++ b/packages/binding-http/src/codecs/tuya-codec.ts
@@ -32,7 +32,9 @@ export default class HttpTuyaCodec implements ContentCodec {
 
     bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
         const parsedBody: TuyaOutput = JSON.parse(bytes.toString());
-        if (parsedBody.success !== true) {
+        const success = parsedBody.success ?? false;
+
+        if (!success) {
             throw new Error(parsedBody.msg != null ? parsedBody.msg : JSON.stringify(parsedBody));
         }
 
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -214,7 +214,9 @@ export class TuyaCustomBearer extends Credential {
             url = `${this.baseUri}/token/${this.refreshToken}`;
         }
         const data: TokenResponse = await (await fetch(url, request)).json();
-        if (data.success === true) {
+        const success = data.success ?? false;
+
+        if (success) {
             this.token = data.result?.access_token;
             this.refreshToken = data.result?.refresh_token;
 
--- a/packages/binding-http/src/credential.ts
+++ b/packages/binding-http/src/credential.ts
@@ -163,6 +163,18 @@ export interface TuyaCustomBearerCredentialConfiguration {
     secret: string;
 }
 
+interface TokenResponse {
+    success?: boolean;
+    result?: {
+        // eslint-disable-next-line camelcase
+        access_token?: string;
+        // eslint-disable-next-line camelcase
+        refresh_token?: string;
+        // eslint-disable-next-line camelcase
+        expire_time?: number;
+    };
+}
+
 export class TuyaCustomBearer extends Credential {
     protected key: string;
     protected secret: string;
@@ -184,14 +196,15 @@ export class TuyaCustomBearer extends Credential {
             await this.requestAndRefreshToken(isTokenExpired);
 
         const url: string = request.url;
-        const body = request.body ? request.body.read().toString() : "";
-        const headers = this.getHeaders(true, request.headers.raw(), body, url, request.method);
+        const body = request.body?.read().toString();
+        const method = request.method;
+        const headers = this.getHeaders(true, request.headers.raw(), body, url, method);
         Object.assign(headers, request.headers.raw());
-        return new Request(url, { method: request.method, body: body !== "" ? body : undefined, headers });
+        return new Request(url, { method, body: body !== "" ? body : undefined, headers });
     }
 
     protected async requestAndRefreshToken(refresh: boolean): Promise<void> {
-        const headers = this.getHeaders(false, {}, "");
+        const headers = this.getHeaders(false, {});
         const request = {
             headers,
             method: "GET",
@@ -200,36 +213,49 @@ export class TuyaCustomBearer extends Credential {
         if (refresh) {
             url = `${this.baseUri}/token/${this.refreshToken}`;
         }
-        const data = await (await fetch(url, request)).json();
-        if (data.success) {
-            this.token = data.result.access_token;
-            this.refreshToken = data.result.refresh_token;
-            this.expireTime = new Date(Date.now() + data.result.expire_time * 1000);
+        const data: TokenResponse = await (await fetch(url, request)).json();
+        if (data.success === true) {
+            this.token = data.result?.access_token;
+            this.refreshToken = data.result?.refresh_token;
+
+            const expireTime = data.result?.expire_time;
+            if (expireTime != null) {
+                this.expireTime = new Date(Date.now() + expireTime * 1000);
+            }
         } else {
             throw new Error("token fetch failed");
         }
     }
 
-    private getHeaders(NormalRequest: boolean, headers: unknown, body: string, url?: string, method?: string) {
+    private getHeaders(NormalRequest: boolean, headers: unknown, body?: string, url?: string, method?: string) {
         const requestTime = Date.now().toString();
         const replaceUri = this.baseUri.replace("/v1.0", "");
-        const _url = url ? url.replace(`${replaceUri}`, "") : undefined;
+        const _url = url?.replace(replaceUri, "");
         const sign = this.requestSign(NormalRequest, requestTime, body, _url, method);
         return {
             t: requestTime,
             client_id: this.key,
             sign_method: "HMAC-SHA256",
             sign,
-            access_token: this.token || "",
+            access_token: this.token ?? "",
         };
     }
 
-    private requestSign(NormalRequest: boolean, requestTime: string, body: string, path = "", method?: string): string {
-        const bodyHash = crypto.createHash("sha256").update(body).digest("hex");
+    private requestSign(
+        NormalRequest: boolean,
+        requestTime: string,
+        body?: string,
+        path = "",
+        method?: string
+    ): string {
+        const bodyHash = crypto
+            .createHash("sha256")
+            .update(body ?? "")
+            .digest("hex");
         let signUrl = "/v1.0/token?grant_type=1";
         const headerString = "";
         let useToken = "";
-        const _method = method || "GET";
+        const _method = method ?? "GET";
         if (NormalRequest) {
             useToken = this.token ?? "";
             const pathQuery = queryString.parse(path.split("?")[1]);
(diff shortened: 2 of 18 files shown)


packages/binding-http/src/oauth-token-validation.ts
```diff
@@ -60,11 +60,11 @@ function extractTokenFromRequest(request: http.IncomingMessage) {
     const url = new URL(request.url ?? "", `http://${request.headers.host}`);
     const queryToken = url.searchParams.get("access_token");
 
-    if (!headerToken && !queryToken) {
+    if (headerToken == null && queryToken == null) {
         throw new Error("Invalid request: only one authentication method is allowed");
```
danielpeintner, line 64: Somehow the error message does not seem right.. or confuses my.. the code checks whether both are null.. but the error says just one method is allowed?

JKRhb: Oh, good point! Seems like the logic was intended to check for the exact opposite, right (i.e., `headerToken != null && queryToken != null`)?

danielpeintner: I think so too but let's wait for @relu91 to comment...

JKRhb:
(Rebased the PR against the latest state of `master`.)

packages/binding-http/src/oauth-token-validation.ts
relu91: yes, exactly the translation is wrong. It should be as [suggested](https://github.com/eclipse-thingweb/node-wot/pull/1097/files#r1335868880).

JKRhb: Since it is more than a refactoring but a change of the actual logic, should we maybe make this change via a separate PR? Or is fine to keep it in here?

danielpeintner: I think in this case I am fine with integrating it as part of this PR also. But either way is fine..

JKRhb: Okay, then we should be ready to merge this one :)

packages/binding-http/src/codecs/tuya-codec.ts
```diff
@@ -17,17 +17,28 @@ import { ContentCodec } from "@node-wot/core";
...
     }
 
     bytesToValue(bytes: Buffer, schema: TD.DataSchema, parameters: { [key: string]: string }): DataSchemaValue {
-        const parsedBody = JSON.parse(bytes.toString());
-        if (!parsedBody.success) throw new Error(parsedBody.msg ? parsedBody.msg : JSON.stringify(parsedBody));
-        for (const key in parsedBody.result) {
-            if (parsedBody.result[key].code === schema["tuya:PropertyName"]) {
-                return parsedBody.result[key].value;
+        const parsedBody: TuyaOutput = JSON.parse(bytes.toString());
+        if (parsedBody.success !== true) {
```
relu91, line 35: is this allowed?
```suggestion
        if (!(parsedBody.success ?? false)) {
```

[source](https://typescript-eslint.io/rules/strict-boolean-expressions/): See last correct example

packages/binding-http/src/credential.ts
```diff
@@ -200,36 +213,49 @@ export class TuyaCustomBearer extends Credential {
         if (refresh) {
             url = `${this.baseUri}/token/${this.refreshToken}`;
         }
-        const data = await (await fetch(url, request)).json();
-        if (data.success) {
-            this.token = data.result.access_token;
-            this.refreshToken = data.result.refresh_token;
-            this.expireTime = new Date(Date.now() + data.result.expire_time * 1000);
+        const data: TokenResponse = await (await fetch(url, request)).json();
+        if (data.success === true) {
```
relu91, line 217: See above

packages/binding-http/src/routes/property.ts
```diff
@@ -106,7 +106,7 @@ export default async function propertyRoute(
             res.end(message);
         }
     } else if (req.method === "PUT") {
-        if (!property.readOnly) {
+        if (property.readOnly !== true) {
```
relu91, line 109: See above

packages/binding-http/src/http-server.ts
```diff
@@ -253,8 +252,10 @@ export default class HttpServer implements ProtocolServer {
 
     /** returns server port number and indicates that server is running when larger than -1  */
     public getPort(): number {
-        if (this.server.address() && typeof this.server.address() === "object") {
-            return (<AddressInfo>this.server.address()).port;
+        const address: AddressInfo | string | null = this.server?.address?.();
+
+        if (typeof address === "object") {
+            return address?.port ?? -1;
```
relu91, line 258: Noticed that this getPort method is different from the http-client-basic-test, can you explain why?

packages/binding-http/src/http-server.ts
```diff
@@ -298,28 +299,25 @@ export default class HttpServer implements ProtocolServer {
...
 
     private addUrlRewriteEndpoints(form: TD.Form, forms: Array<TD.Form>): void {
-        if (this.urlRewrite) {
-            for (const inUri in this.urlRewrite) {
-                const toUri = this.urlRewrite[inUri];
-                if (form.href.endsWith(toUri)) {
+        if (this.urlRewrite != null) {
+            for (const [inUri, toUri] of Object.entries(this.urlRewrite)) {
+                // TODO: Without === true, this is currently an eslint false positiv
+                if (form.href.endsWith(toUri) === true) {
```
relu91, line 320: Should we just disable the rule? or is it maybe that endsWith can result in undefined? 🤔

packages/binding-http/src/http-server.ts
```diff
@@ -336,10 +334,10 @@ export default class HttpServer implements ProtocolServer {
             let anyProperties = false;
             for (const propertyName in thing.properties) {
                 anyProperties = true;
-                if (!thing.properties[propertyName].readOnly) {
+                if (thing.properties[propertyName]?.readOnly !== true) {
```
relu91, line 337: See above

packages/binding-http/src/http-server.ts
```diff
@@ -336,10 +334,10 @@ export default class HttpServer implements ProtocolServer {
...
                     allReadOnly = false;
                 }
-                if (!thing.properties[propertyName].writeOnly) {
+                if (thing.properties[propertyName].writeOnly !== true) {
```
relu91, line 340: See above

packages/binding-http/src/http-server.ts
```diff
@@ -377,13 +375,13 @@ export default class HttpServer implements ProtocolServer {
                     form,
                     (tdTemplate.properties?.[propertyName] ?? {}) as PropertyElement
                 );
-                if (thing.properties[propertyName].readOnly) {
+                if (thing.properties[propertyName].readOnly === true) {
```
relu91, line 378: See above
```suggestion
                if (thing.properties[propertyName].readOnly ?? false) {
```

packages/binding-http/src/http-server.ts
```diff
@@ -377,13 +375,13 @@ export default class HttpServer implements ProtocolServer {
...
                     form.op = ["readproperty"];
                     const hform: HttpForm = form;
                     if (hform["htv:methodName"] === undefined) {
                         hform["htv:methodName"] = "GET";
                     }
-                } else if (thing.properties[propertyName].writeOnly) {
+                } else if (thing.properties[propertyName].writeOnly === true) {
```
relu91, line 384: See above
```suggestion
                } else if (thing.properties[propertyName].writeOnly ?? false) {
```

packages/binding-http/src/http-server.ts
```diff
@@ -565,35 +563,33 @@ export default class HttpServer implements ProtocolServer {
             return;
         }
 
-        if (thing.securityDefinitions) {
```
relu91, line 568: Given the comment, this should probably be: `if(thin.security == null)`. Right?

relu91 (changes requested):
I selected request changes but they are mostly suggestions and comments. I'm open to changing my mind on the suggested changes, let me know!

packages/binding-http/src/http-server.ts
JKRhb: Very good point, I copied over the code from there to here :)

In general, though, these getters should be revisited later IMHO as returning `-1` does not seem like a very good practice to me. Maybe that is why I forgot to also provide a "proper" fix here :/

Another thing I noticed, by the way, is that we should consider using getters and setters as defined by JavaScript for this kind of method in the future (see, e.g., [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) for an example) which is more idiomatic and simplifies some of the APIs.

The method implementation would then become something like this:

```ts
    class HttpServer {
        public get port(): number {
            return this.#port;
        }
    }
```

While the respective properties could be accessed like so (effectively getting rid of the `get` prefix):

```ts
    const serverPort = httpServer.port;
```

packages/binding-http/src/http-server.ts
JKRhb: As `thing.security` is a mandatory field, I think this part of the code covers the `thing.security.length === 0` case (i.e., the test for `thing.security.length > 0` above does not return true and we did not return from the method early).

This could definitely be a bit clearer, though – maybe we can perform some additional cleanups after this PR has been merged?

packages/binding-http/src/codecs/tuya-codec.ts
JKRhb: I think both versions are fine here – in the end, it is probably a question of readability 🤔 Due to otherwise doubled negation, I actually tend to the `!== true` version in this instance but I would also be fine with nullish coalescing :)

JKRhb: In order to maximize readability, I now introduced an `success` auxiliary variable – so I think, this should now be resolved :)

packages/binding-http/src/http-server.ts
JKRhb: I think this issue as well as others should get resolved by #1103, I will therefore remove the `=== true` check as well as the TODO :)

JKRhb:
Hmm, the persisting linting errors are still puzzling me a bit, as I don't get the same errors on my own machine. I think I now remember, though, that they were the reason why I used the `=== true` approach in the first place. It seems to me, though, that some of the errors are caused by the TS config and using `ES6` as the compile target.

I think the easiest approach would be to stick to `=== true` for now and maybe revisit this issue later.

JKRhb:
Since the updated eslint packages did not solve the issue and to make the history a bit cleaner, I dropped the commits merged in from #1103 and rebased against `master`.

packages/binding-http/src/http-server.ts
relu91: +1 for both considerations. However, I would address those in a separate PR. Regarding this PR we can consider this issue resolved.

packages/binding-http/src/http-server.ts
relu91: It is a mandatory field in the TD but not in the `ThingExposeInit`. In my understanding, there is nothing so far to check if the `security` is defined. Probably, due to #911. Currently, we just ask the bindings to handle the `ExposeThingInit` see https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/servient.ts#L38 .  So `thing.security` might be also `null` at that point in time.   Do the new changes take care of this too? 🤔 .

JKRhb: Ohh, I see, then this case might not be covered yet! I will double-check this :)

JKRhb: Hmm, the parameter passed into the method is of type `security: string | [string, ...string[]]` – so it is mandatory, but I suppose at least one thing that is missing is the case when security is a single string?

packages/binding-http/src/http-server.ts
JKRhb: I now managed to get rid of the `=== true` approach using some explicit typing :)

packages/binding-http/src/http-server.ts
JKRhb: I tried to provide a fix now in 7a0bd0240de661a9e2df5981038a48cfceb92ba4 – I think that should work for now, however, it seems to me as if the handling of `security` and `securityDefinitions` needs to be revisited in general.

JKRhb: Aparently, the code has been correct before, so it reverted the change. Seems to me as if the typing of the `security` field in the `ExposedThing` class/interface might need to be fixed.

JKRhb:
Rebased against `master` to resolve a conflict – with the latest changes, I think/hope this PR should be ready for merging now :)

packages/binding-http/src/http-server.ts
relu91: I know that it is counter-intuitive and I agree that the bug is somewhere else (e.g. #911). If you go upwards in the call hierarchy you'll notice that after producing method nobody checks if `security` is defined, tomorrow I'll try with a practical example. However for the time being we can merge the PR and track this issue in #911 .

relu91 (approved):
All the comments above have been resolved. Merging. Thanks @JKRhb !!!

JKRhb:
Awesome, thank you @relu91! :)

Maybe one thing, though: It's probably best to use the squashing strategy when merging PRs like this one, as the `fixup!` commits I am using are actually intended to be squashed eventually (see, e.g., [here](https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash)). After the discussions we had recently, I refrained from doing so which now, unfortunately, led to some of those commits "leaking" into the main branch. Maybe we could discuss our workflow with handling PRs more generally at one of the upcoming committer meetings? Potential solutions could be to let the person opening the PR merge the PR after receiving the required approvals, provided that it is a committer, or let them also give a final green light before the merge.

Personally, I would prefer if we would allow to also perform some clean-up in the PR (i.e., rebasing, squashing, and reordering of commits) itself *after* the reviewers gave their approval since having a clean commit history in the main branch is always a nice thing to have, I think. But maybe we could discuss this also at some point (in not too much detail) in a synchronous manner (which would also probably lead to the resolution of #558).

danielpeintner:
As said before.. rebasing, squashing, and reordering of commits while working on a PR makes it very difficult to follow what has changed between commits.

I guess the only thing we need to remember (when "actually" merging the PR) is set/choose the right option to "squash all commits".
I think on GitHub we can specify the default strategy ... we might ask Eclipse to do this setting... IF EVERYONE AGREES WITH ... which isn't clear from reading https://github.com/eclipse-thingweb/node-wot/issues/558

Shall we continue the discussion in https://github.com/eclipse-thingweb/node-wot/issues/558

JKRhb:
> As said before.. rebasing, squashing, and reordering of commits while working on a PR makes it very difficult to follow what has changed between commits.

You are completely right, that should only be done once all reviews have been addressed. Sorry once more for not following that principle before.

> I guess the only thing we need to remember (when "actually" merging the PR) is set/choose the right option to "squash all commits". I think on GitHub we can specify the default strategy ... we might ask Eclipse to do this setting... IF EVERYONE AGREES WITH ... which isn't clear from reading #558
> 
> Shall we continue the discussion in #558

:+1:

PR #1098, opened 2023-09-22
chore: move common devDependencies to top-level package.json

JKRhb:
Out of curiosity, I tried out if you can reduce redundancy in the `package.json` files by including shared `devDependencies` only in the top-level `package.json` file – and apparently, that works :)

This might require some double-checking but it could simplify updating shared devDependencies like TypeScript, eslint, and prettier in the future.


egekorkan:
It looks ok for the mqtt package but in other ones I also see new packages introduced. Why is it the case?

JKRhb:
> It looks ok for the mqtt package but in other ones I also see new packages introduced. Why is it the case?

In cases where a `devDependencies` field continues to be present, some entries were only used by the package in question (the diff might be a bit misleading here). We could also move those to the top level, but I thought it might make more sense to keep such devDependencies limited to the package that actually uses them.

JKRhb:
As some of the dependency versions got changed in the individual `package.json` files, I now reverted them back to the original versions to make it a bit clearer what has actually changed. A proper update should then be performed in a follow-up PR.

packages/binding-file/package.json
```diff
@@ -14,21 +14,7 @@
...
-        "eslint-config-standard": "^16.0.3",
-        "eslint-plugin-import": "^2.24.2",
-        "eslint-plugin-node": "^11.1.0",
-        "eslint-plugin-promise": "^5.1.0",
-        "eslint-plugin-unused-imports": "^1.1.4",
-        "prettier": "^2.3.2",
-        "typescript": "4.7.4",
-        "typescript-standard": "^0.3.36",
-        "wot-typescript-definitions": "0.8.0-SNAPSHOT.26"
+        "@node-wot/td-tools": "0.8.9"
```
danielpeintner, line 17: I am surprised to see still `td-tools` ...

relu91:
I'm wondering, don't we have a script dedicated to this kind of management of the dev deps? I talking about [this](https://github.com/eclipse-thingweb/node-wot/blob/master/utils/check_package_version_consistency.js), if it failed to check consistency what is there for?

packages/binding-file/package.json
JKRhb: Hmm, yeah, I think in the other packages they are used as a dependency and not as a devDependency, I think in this case it works as `td-tools` definitions are only used for typing. Or should `td-tools` generally be replaced with `wot-typescript-definitions`?

JKRhb:
> I'm wondering, don't we have a script dedicated to this kind of management of the dev deps? I talking about [this](https://github.com/eclipse-thingweb/node-wot/blob/master/utils/check_package_version_consistency.js), if it failed to check consistency what is there for?

Oh, that is a very good point! In theory, the script should have indicated this kind of redundancy before. However, I think the actual check is currently disabled due to this line:

https://github.com/eclipse-thingweb/node-wot/blob/bd247a0ac737daf0bb2e012d76124cc07e93c459/utils/check_package_version_consistency.js#L87C5-L87C44

I will investigate and update the script if possible :)

JKRhb:
The script should now be fixed by 0452ce838fe2c71074a94ca0613ca123d18968c5

utils/check_package_version_consistency.js
```diff
@@ -48,22 +58,25 @@ async function main() {
...
-            console.log(versionPackages);
-            nbErrors++;
+    for (const collectedDepedencies of [allDependencies, devDependencies]) {
+        for (const [module, versionPackages] of Object.entries(collectedDepedencies)) {
+            const versions = Object.keys(versionPackages);
+            if (versions.length !== 1) {
+                console.log("Warning module ", module, " has multiple versions ", versions.join(" "));
+                console.log(versionPackages);
+                nbErrors++;
+            }
```
JKRhb, line 68: Hmm, just noticed that this check does not need to be performed for the `devDependencies` as well, I will adjust these lines.

JKRhb: Fixed in 3c88f716f078632e34e1aabd570cc378c569dbb6

packages/binding-file/package.json
danielpeintner: I see, let's keep it as is (or as it was)

relu91:
Ok, thank you @JKRhb, my understanding is that in the future the script will pick up inconsistencies once again and if something is off the Github action will warn us. Is that correct?

JKRhb:
> Looks good to me!! we should document somewhere the stack that we are using for writing tests so that newcomers don't accidentally add new testing frameworks like `jest`.

Very good point! Should we open an issue for that?

danielpeintner (approved):
LGTM

PR #1103, opened 2023-09-30
chore(devDeps): upgrade eslint packages

JKRhb:
In the context of #1097, I noticed that eslint currently reports a couple of false positives when it comes to typing. Furthermore, when running `npm run lint`, it warns of missing compatibility with the currently used TypeScript version.

This PR simply updates all of the eslint-related packages and applies some (partially automated) fixes to the codebase.


relu91 (approved):
Thanks this was very well needed.

danielpeintner (approved):

PR #1104, opened 2023-09-30
refactor(core): use # instead __ as internal handler prefix

JKRhb:
In the context of #1097, I remembered that JavaScript nowadays has a built-in prefix for private methods and class fields (`#`).

As it turns out, fields with this prefix get ignored when calling `JSON.stringify`, so using it instead of `__` for the internal handler maps of the `ExposedThing` class enables us to get rid of some checks during serialization, which should make the code a bit more idiomatic and slightly increase its performance.

As the handler maps are now truly private, I had adjust some of the test code for it to still work. However, I think the code in question was rather trivial, so this change should hopefully be fine.


relu91 (approved):
PR looks good. I would say that those fields could be better modeled as protected (hence the private modifier would not work), but let's see if we find a use case for extending the exposed-thing in the future.

packages/core/src/exposed-thing.ts
```diff
@@ -66,19 +66,19 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
     };
 
     /** A map of property (read & write) handler callback functions */
-    __propertyHandlers: PropertyHandlerMap = new Map<string, PropertyHandlers>();
+    #propertyHandlers: PropertyHandlerMap = new Map<string, PropertyHandlers>();
```
danielpeintner, line 69: Maybe it is worth to mention that `#` prohibits that the data is serialized into JSON

danielpeintner (approved):
LGTM and thanks @JKRhb 

I just added the comment that it might make sense to put a statements why the # prefix has been used for the internal handlers..

packages/core/src/exposed-thing.ts
JKRhb: I tried to incorporate this in 2db8506c15a28996b2375c4c60d9ec18e9b1fffc – hope the comments are not too verbose now :/

PR #1105, opened 2023-10-02
refactor: use `structuredClone` for deep copying

JKRhb:
JavaScript nowadays has a built-in function for deep-cloning called `structuredClone` which makes it possible to get rid of the `JSON.parse(JSON.stringify(foo))` construction. I found a few places where the new function can be used, hopefully increasing readability and performance a bit.

Cases where the parse-stringify-approach still needs to be used *could* be revisited in the future to arrive at a slightly cleaner implementation.

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -371,7 +371,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         this.events = {};
         // Deep clone the Thing Model
         // without functions or methods
-        const clonedModel = JSON.parse(JSON.stringify(thingModel));
+        const clonedModel = structuredClone(thingModel);
         Object.assign(this, clonedModel);
         this.extendInteractions();
     }
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -369,10 +369,9 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         this.properties = {};
         this.actions = {};
         this.events = {};
-        // Deep clone the Thing Model
-        // without functions or methods
-        const clonedModel = structuredClone(thingModel);
-        Object.assign(this, clonedModel);
+
+        const deepClonedModel = structuredClone(thingModel);
+        Object.assign(this, deepClonedModel);
         this.extendInteractions();
     }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/consumed-thing.ts
```diff
@@ -371,7 +371,7 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         this.events = {};
         // Deep clone the Thing Model
         // without functions or methods
-        const clonedModel = JSON.parse(JSON.stringify(thingModel));
+        const clonedModel = structuredClone(thingModel);
```
danielpeintner, line 374: I was wondering how `structuredClone` handles functions on objects and I think it fails.. at least in my simple test it fails with 
`[ERR]: The object could not be cloned.`

Are we sure that there a no functions involved in all cases?
Apparently not since all tests fail but I would like to "ask" the question ;-)

JKRhb: There are a few places where I ran into the same error, so I limited the use of `structuredClone` to cases where "pure" objects without functions/methods are being used. So, especially for classes, this new function is apparently not really usable :/ But for regular objects, this new function is recommended over the "traditional" `JSON.parse(JSON.stringify(foo))` approach, as far as I've read.

packages/core/src/helpers.ts
```diff
@@ -220,7 +220,7 @@ export default class Helpers implements Resolver {
      * Helper function to remove reserved keywords in required property of TD JSON Schema
      */
     static createExposeThingInitSchema(tdSchema: unknown): SomeJSONSchema {
-        const tdSchemaCopy = JSON.parse(JSON.stringify(tdSchema));
+        const tdSchemaCopy = structuredClone(tdSchema) as SomeJSONSchema;
```
JKRhb, line 223: I added this as another instance suitable for using the `structuredClone` function – the typing is not that nice here, but I think casting should be fine here since we are performing type checks down below.

packages/core/src/consumed-thing.ts
danielpeintner: I see, the only thing that bothers me was the comment in the code
`// without functions or methods`

If there are no such "functions" we should remove that comment.
What do you think?

JKRhb: Very good point, I made some adjustments in 33af731e4a8a988414155718addc4702e05dd029 :) I also baked the deep cloning documentation into the variable names in these cases, getting rid of the `// Deep clone the Thing Model` comment :)

danielpeintner (approved):

packages/core/src/helpers.ts
relu91: I guess it is acceptable, we could even put it as the argument type given the fact that is just an internal helper function

relu91 (approved):
Good to go!

PR #1106, opened 2023-10-03
refactor(core): use private variables instead of function wrappers

JKRhb:
As another addition to #1104, this PR replaces a couple of internal getters (that have been used for avoiding serialization) with built-in private variables, making the code a bit cleaner.


danielpeintner (approved):
LGTM

Maybe it makes sense to add a comment why we use `#` but in this case I don't think I would do that since we would need to explain the rational all over the place.

I think it is good to go as is 👍

JKRhb:
I tried to address your feedback, @danielpeintner, in fcc3b70516b47223b2e6f8bf97ab057707f7c6e4, let me know if you see further room for improvement :)

relu91 (approved):
were they included in the TD? I've never noticed, otherwise, the comment is a little bit misleading.

JKRhb:
> were they included in the TD? I've never noticed, otherwise, the comment is a little bit misleading.

The result is basically the same as before, the only difference is that we do not need the getter construction anymore. We could also remove the comments as it is common JS behavior, I guess it depends on how useful the reminder of this behavior is.

danielpeintner:
Note: We used to have the workaround to wrap the  internal state into functions to not be stringified in a TD. The current approach  using private variables is much better in my regard. Hence I think the comment is fine 🤷‍♂️

PR #1107, opened 2023-10-09
refactor(td-tools): Aid updates

danielpeintner:
* use lowercase interaction name
* rework README and put focus on AID submodel usage


JKRhb (approved):
Changes look good to me, just spotted one nit in the `aid-to-td` example.

PR #1110, opened 2023-10-09
refactor: remove unnecessary `new Promise` wrappers

### LINKED ISSUE #569: Promise async code style

JKRhb:
This PR contributes to resolving #569. There are a few instances of the `return Promise` pattern left that were a bit difficult to resolve, though, so I am not sure whether we should keep the issue open?

### DIFF
--- a/packages/binding-coap/test/coap-server-test.ts
+++ b/packages/binding-coap/test/coap-server-test.ts
@@ -127,10 +127,8 @@ class CoapServerTest {
             },
         });
 
-        testThing.setActionHandler("try", (input: WoT.InteractionOutput) => {
-            return new Promise<string>((resolve, reject) => {
-                resolve("TEST");
-            });
+        testThing.setActionHandler("try", async (input: WoT.InteractionOutput) => {
+            return "TEST";
         });
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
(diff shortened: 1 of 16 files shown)


danielpeintner (approved):
I looked trough the changes and they seem all fine by me 👍

I would also argue after merging this PR we should close https://github.com/eclipse-thingweb/node-wot/issues/569.

relu91 (approved):
Looks good! regarding the issue linked, I think it might be difficult to be sure that every non-necessary promise it has been removed. I remember finding an [eslint extension](https://eslint-plugin-es.mysticatea.dev/rules/no-promise.html)... maybe we could resolve it once the extension is in place and the non-convertible cases are correctly opted out.

relu91:
Given we have two approvals I'm merging, let's wait for the eslint rule before closing #569

JKRhb:
That sounds good :) I updated the PR description before your merge so that #569 will stay open.

> Looks good! regarding the issue linked, I think it might be difficult to be sure that every non-necessary promise it has been removed. I remember finding an [eslint extension](https://eslint-plugin-es.mysticatea.dev/rules/no-promise.html)... maybe we could resolve it once the extension is in place and the non-convertible cases are correctly opted out.

Issue #1112, opened 2023-10-11
AID does not allow colons (i.e., `:`) in `idShort`


danielpeintner:
We will be getting AID files with the following `"idShort": "htv_methodName"` on `forms` level and need to convert it to `"htv:methodName"` for TD creation (and vice-versa when generating AIDs)

At the moment the following terms are supported:
* `htv_methodName`
* `htv_headers`
* `mqv_retain`
* `mqv_controlPacket`
* `mqv_qos`
* `modbus_function`
* `modbus_entity`
* `modbus_zeroBasedAddressing`
* `modbus_pollingTime`
* `modbus_timeout`
* `modbus_type`

PR #1113, opened 2023-10-12
refactor(td-tools): AID idShort does not allow ":" and uses "_" instead

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1112


relu91 (approved):
this column problem is interesting... it means that basically, it would not allow any extension to the TD (i.e., importing other ontologies, etc.)

danielpeintner:
> this column problem is interesting... it means that basically, it would not allow any extension to the TD (i.e., importing other ontologies, etc.)

There are other means in place/planned but I agree, this is somewhat limiting but from the implementation point of view we cannot do much about it...

Issue #882, opened 2022-11-29
Try to get back all Os's for testing + firestore


danielpeintner:
LGTM... just remember to have a tracking issue so that in the future we can restore the full checklist.

_Originally posted by @relu91 in https://github.com/eclipse/thingweb.node-wot/pull/878#pullrequestreview-1197116696_

JKRhb:
I suppose to the removal of the firestore binding, this issue can be closed, right?

danielpeintner:
> I suppose due to the removal of the firestore binding, this issue can be closed, right?

I don' think so. We used to have MacOS testing for our GHAs as well.
Question: Shall we bring them back? The downside is that our GHAs run longer.

What do people think?  Is Ubuntu testing enough?

JKRhb:
Oh, sorry, I think I slightly misread the title 😅 I just did a [test run](https://github.com/eclipse-thingweb/node-wot/actions/runs/6468614427) of the full CI pipeline, and apparently, macOS and Windows are still considerably slower than Linux (about six minutes for macOS and eight minutes for Windows).

Maybe we could adjust the workflows so that the full CI workflow is being run when a push to `master` occurs? I think at the moment, it only runs when triggered manually. However, as eight minutes is not as bad as it was before, I think, we could also consider bringing Windows and macOS back in general to have a greater platform coverage.

danielpeintner:
Note:
* https://github.com/eclipse-thingweb/node-wot/pull/1121 reintroduced macOS and Windows testing
* binding-firestore has been removed in the meanwhile

PR #1116, opened 2023-10-13
ci: reintroduce macOS and Windows testing on the `master` branch

JKRhb:
As discussed in #882, this PR reintroduces automated testing with Windows and macOS on the `master` branch. To limit the use of macOS and Windows to the `master` branch, I adjusted the testing matrix so that the two OSs are excluded if the current branch should not be `master`. This rendered the other workflow definition file obsolete, which is why I removed it.

While I was at it, I slightly adjusted the `on` section in the workflow file so that the workflow only runs in the `push` case when it is a push to the `master` branch. This resolves the slight annoyance that the CI ran twice for some reason in the case of a PR coming from a fork.

With this PR, we could consider #882 resolved. However, we could also leave it open and evaluate if macOS and Windows testing should be reintroduced for PRs as well.


danielpeintner (commented):
LGTM

The only use-case I have in mind is the following. At the moment MacOS and Windows are run only against master. Good so far, However, how can we (even manually) trigger the CI to run against all OS'es ... because we want to check that also in some special cases...

JKRhb:
I think with the `workflow_dispatch` case added, we should be able to run the workflow for certain branches :) At least that seems to be the case with the current full CI workflow: 

image

JKRhb:
Hmm. On second thought, that probably does not change the CI matrix. Maybe there could be another case added for that?

JKRhb:
> Hmm. On second thought, that probably does not change the logic within the CI matrix. Maybe there could be another case added for that?

Not entirely sure if it works, but if it does, the last commit should have added it in ;)

relu91:
I'm not entirely convinced about narrowing down the matrix to only work on merge. I would like to know immediately if a PR is breaking and fix the errors in the PR. I appreciate the effort, but my preference would be to wait longer and be sure that we have all greens.

JKRhb:
> I'm not entirely convinced about narrowing down the matrix to only work on merge. I would like to know immediately if a PR is breaking and fix the errors in the PR. I appreciate the effort, but my preference would be to wait longer and be sure that we have all greens.

Yeah, I think we can definitely go this route, as mentioned in the issue, the total time is now about 8 minutes which should be fine as, if something should be completely broken, we will notice it much earlier on `ubuntu-latest`. I will open another PR as an alternative to this one.

JKRhb:
I just opened #1121 as alternative :)

PR #1118, opened 2023-10-13
docs(README): use white logo on dark backgrounds

relu91:


README.md
```diff
@@ -1,4 +1,8 @@
-<img title="ThingWeb" alt="Thingweb logo" src="https://raw.githubusercontent.com/eclipse-thingweb/website/master/misc/thingweb_logo.svg" width="300px">
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/eclipse-thingweb/website/master/misc/thingweb_logo_for_dark_bg.svg">
+  <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/eclipse-thingweb/website/master/misc/thingweb_logo.svg">
+  <img title="ThingWeb" alt="Thingweb logo" src="" width="300px">
```
JKRhb, line 4: Thank you, @relu91, that is a great addition! (Not only because I am using the dark theme myself 😉)

However, inserting the code into the W3 validator (partly out of curiosity because of the brief discussion regarding the `width` during our meeting), it seems that the `px` should in fact be removed and also a fallback value for the `src` attribute should be provided.

As most browsers should support the current version, though, these suggestions are definitely optional and the current state is more than fine.

```suggestion
  <img title="ThingWeb" alt="Thingweb logo" src="https://raw.githubusercontent.com/eclipse-thingweb/website/master/misc/thingweb_logo.svg" width="300">
```

JKRhb (approved):
Looks good to me, just spotted two minor points for improvement :)

README.md
danielpeintner: We should take over these changes to other repos as well...

relu91: Oh thank you for checking @JKRhb, this is something I should also update in Zion readme. Great I'll commit your suggestions and then merge.

PR #1072, opened 2023-09-06
update opcua modules - remove unnecessary data transformation

erossignon:

PR #1119, opened 2023-10-13
refactor(binding-opcua): remove unnecessary data transformation

danielpeintner:
see https://github.com/eclipse-thingweb/node-wot/pull/1072


relu91 (approved):
LGTM, in my understanding, it is just the first step of many right?

JKRhb (approved):

PR #1124, opened 2023-10-16
Add quick start online Things to README.md

egekorkan:
Extending the readme with the other online Things we have. For the CG tutorial, we have stored the TDs in Zion and I am doing the same here but we can discuss that if needed. The reason for that was that we needed somewhere to store the MQTT TD. Alternatives could be: 
- Storing the TDs in this GitHub
- Storing them at `domus-directory` when it is here


danielpeintner (approved):
LGTM

Issue #1099, opened 2023-09-23
Support Objects in octet-stream codec


derwehr:
I'd like to propose supporting objects in octet streams, for example, for describing a device's payload data that sends packets with bit sequences containing measurements starting at different offsets. 
i.e., a device sends 2 bytes, where bit 1 & 2 are different flags and bit 3-16 are a temperature measurement.

Similiar to this:
```JSON
"measurements": {
  // ...
  "type": "object",
  "properties": {
    "flag1": {
      "type": "boolean",
      "offset": 0,
      "bitLength": 1,
      // ...
    },
    "flag2": {
      "type": "boolean",
      "offset": 1,
      "bitLength": 1,
      // ...
    },
    "temperature": {
      "type": "number",
      "offset": 2,
      "bitLength": 14
      "scale": 2, // number of digits to the right of the decimal point
    // ...
    }
  }
}
```

danielpeintner:
I hope I understand your proposal correctly.
You would like to describe the data exchange `measurements` with JSON but later one it should be transmitted more efficiently via an octect stream since you know each flag just requires _one_ bit etc, right?

At the moment the default encoding is JSON and as such `measurements` is transmitted. If you want to do additional things you need to make sure that it is clear "what" the contenttype is and you can also implement your own codec on top of it (using also existing codecs)

see https://github.com/eclipse-thingweb/node-wot#mediatype-support

derwehr:
I was thinking about the opposite way, where my device sends an octet stream and the JSON above is a possible description of the payload data in the device's TD.

More precisely, I would like to extend the octet stream to not throw an Error at https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/codecs/octetstream-codec.ts#L108, but call a new funtion `valueToObject`.

If that makes sense, I could try and provide a PR

danielpeintner:
I think in any case it needs to be balanced. Having said that, you would need a solution for `bytesToValue(...)` but also for `valueToBytes(...)`.
I wonder whether a generic solution for all different kinds of definitions can be established...

derwehr:
I thought, that with the data type, bit length, and bit offset defined in the TD, one could re-use the de/encoding method for the respective data type. Thus, shouldn't it be as generalizable as the `integer`, `float`, and `string` operations of the codec?

To extend the example from above
```JSON
"properties": {
  "measurements": {
    // ...
    "type": "object",
    "properties": {
      "flag1": {
[...]

PR #1125, opened 2023-10-16
Support Objects in octet-stream codec #1099

derwehr:
See #1099

As a first step, I added `ex:bitLength` and `ex:bitOffset` to all methods of the codec and extended the tests accordingly.

I'll add support for the `object` schema type next. But because I already changed a lot of code, there might be some material for discussion.


packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -115,23 +136,24 @@ export default class OctetstreamCodec implements ContentCodec {
...
         bytes: Buffer,
-        options: { dataLength: number; bigendian: boolean; signed: boolean }
+        options: { dataLength: number; bigEndian: boolean; signed: boolean }
     ): number {
-        const { dataLength, bigendian, signed } = options;
+        const { dataLength, bigEndian, signed } = options;
+
         switch (dataLength) {
-            case 1:
+            case 8:
```
danielpeintner, line 144: Out of curiosity. What is the big win using bit dataLength vs. byte dataLength?

derwehr: In the `integerToValue` method, the only point is to be consistent with using bit length throughout the codec where possible.
Further up at line 105, when reading only the bits specified at the bit dataLength, I am ensuring that the bit offset is a multiple of 8, so I could use the byte dataLength here, too.

derwehr:
I add a few more test cases to increase coverage
Would you like me to move the new features to another codec instead of extending `octet-steam`, or is there anything else I can change to possibly get this merged?

danielpeintner (approved):
I think it is definitely an improvement since we now support objects  (once a schema is provided for it)

Note: I only did a quick review but it looks good to me and the additional tests are useful also!

Let's wait till @relu91 or @JKRhb  have time to take a look.

BTW, Sorry for not reacting for such a long time

derwehr:
Thanks for the review.
By the way, one notable change I made is that I moved `byteSeq`, `signed` and `length` from the URI parameters into the data schema.
I'm unsure if you'd want that, but it seemed more natural. On second thought, that's a deprecation of the parameters, and maybe I should re-add them along with a check making sure they don't differ from the parameters provided by the data schema.
Please let me know which way you'd prefer.

danielpeintner:
> By the way, one notable change I made is that I moved `byteSeq`, `signed` and `length` from the URI parameters into the data schema.
> I'm unsure if you'd want that, but it seemed more natural. On second thought, that's a deprecation of the parameters, and maybe I should re-add them along with a check making sure they don't differ from the parameters provided by the data schema.

@relu91 @egekorkan @JKRhb do you have an opinion on that aspect?

I think the advantage of having these parameters as URI means one can tweak them at runtime while in the schema case they are fixed. I would even argue that they can be at both places and the URI parameters takes precedence if available.. but I guess this makes it more complex.

egekorkan:
> @relu91 @egekorkan @JKRhb do you have an opinion on that aspect?

Sadly none of the direction seems to be direction that the standards are developing towards. We will probably have protocol specific keywords for that but I can see why you would want to have these protocol independent since data encoding/coding happens in codec in node-wot, which is not aware of the protocol. I would definitely wait for @relu91 before merging.

relu91:
Hi @derwehr! 👋
First of all, thank you very much for your efforts!! I like the idea of improving the octet-stream codec with the ability to encode/decode objects. Sorry for all the friction, but this is a delicate topic because it is a moving target. The problem is that the Thing Description and Protocol binding task forces (part of the WoT WG) are discussing, **these days**, how to improve the mapping between IoT protocol payloads to application data. In practice, in the short future, we might need to change how payloads are processed in node-wot. 

Having said that we discussed it internally and decided that for now, it is good to merge this PR and leave the refactoring for later. Before merging though I might ask you one final job. Let me clarify one point. `byteSeq`, `signed`, and `length` were not URI parameters but rather **content-type** parameters. As the RFC specifies, content-types might have parameters, and third argument of the `bytesToValue` and `valueToBytes` functions is the content-type parameters parsed by the upper layer. See [here](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/content-serdes.ts#L140). Therefore, I would ask you if it is possible to revert the change you made and keep those in the content-type parameters. 

Now, I'll do a review of the rest of your implementation and let you know if there is something else to change before merging.

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -140,98 +166,132 @@ export default class OctetstreamCodec implements ContentCodec {
...
         }
     }
 
+    private objectToValue(bytes: Buffer, schema?: DataSchema): DataSchemaValue {
+        if (schema?.type !== "object") {
+            throw new Error("Schema must be of type 'object'");
+        }
+
+        const result: { [key: string]: unknown } = {};
+        for (const propertyName in schema.properties) {
```
relu91, line 207: to make this work the list of properties should always have the same order. Are we sure that this is consistent across different node instances? maybe it is safer to order them in alphabetical order?

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -140,98 +166,132 @@ export default class OctetstreamCodec implements ContentCodec {
...
+            const propertySchema = schema.properties[propertyName];
+            result[propertyName] = this.bytesToValue(bytes, propertySchema);
```
relu91, line 209: Now I think I understand why you moved the parameters to the schema. I still keep then as global parameters in the content-type and then apply them to all the subschema. It is very unlikely that the same platform will encode the properties in a different format. If it is BE it is BE for all the payloads.

relu91 (changes requested):
I skim through the code but it is hard to visually remove the changes that you did for the parameters. I left a couple of comments, but I think I review the code once again when you revert back to the previous handling of content-type parameters. Thanks!

packages/core/src/codecs/octetstream-codec.ts
derwehr: done :)

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -59,34 +59,57 @@ export default class OctetstreamCodec implements ContentCodec {
         debug("OctetstreamCodec parsing", bytes);
         debug("Parameters", parameters);
 
-        const bigendian = !(parameters.byteSeq?.includes(Endianness.LITTLE_ENDIAN) === true); // default to big endian
+        console.debug(parameters);
```
danielpeintner, line 62: Please remove or user proper logging

packages/core/src/codecs/octetstream-codec.ts
derwehr: I now updated this line to use `Object.getOwnPropertyNames`, which is ordered in Node version 6.4 and up, see https://node.green/#ES2015-misc-own-property-order

packages/core/src/codecs/octetstream-codec.ts
relu91: Yes please keep everything in the content-type for now.

relu91: I'd use the length parameter defined in the content type as an array length. If in the schema we are trying to access outside its ranges (either by specifying a bitOffset or a bitLenght) then it is a parsing error. If we want to downgrade gracefully we can take whatever is possible and log a warning.

derwehr: Perfect, I think that's taken care of by my most recent commit (28b135c346b922a6a22c7c34c8fd9324dda353a4) then

relu91 (approved):
Thank you very, much for your efforts! We can merge it and try to play with it. As a follow-up work, we should provide a short explanation about the new terms introduced.

danielpeintner:
Thank you @derwehr for your contribution 👍

I created https://github.com/eclipse-thingweb/node-wot/issues/1194 as follow-up.

PR #1126, opened 2023-10-17
chore: enable strictNullChecks globally

JKRhb:
As another step towards resolving #1046, this PR enables `strictNullChecks` globally and removes it from individual packages that extend the global `tsconfig.json` file.

PR #1127, opened 2023-10-17
refactor: enable strict-boolean-expressions globally

### LINKED ISSUE #1046: Strict boolean expressions

JKRhb:
Depends on #1126, resolves #1046.

### DIFF
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -98,7 +98,8 @@ export default class DefaultServient extends Servient {
 
         // apply flags
         if (clientOnly) {
-            this.config.servient ??= { clientOnly };
+            this.config.servient ??= {};
+            this.config.servient.clientOnly = true;
         }
 
         // set log level before any output
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -31,7 +31,7 @@ const { debug, error, info } = createLoggers("cli", "cli-default-servient");
 
 // Helper function needed for `mergeConfigs` function
 function isObject(item: unknown) {
-    return item && typeof item === "object" && !Array.isArray(item);
+    return item != null && typeof item === "object" && !Array.isArray(item);
 }
 
 /**
@@ -98,10 +98,7 @@ export default class DefaultServient extends Servient {
 
         // apply flags
         if (clientOnly) {
-            if (!this.config.servient) {
-                this.config.servient = {};
-            }
-            this.config.servient.clientOnly = true;
+            this.config.servient ??= { clientOnly };
         }
 
         // set log level before any output
@@ -111,7 +108,9 @@ export default class DefaultServient extends Servient {
         this.addCredentials(this.config.credentials);
 
         // remove secrets from original for displaying config (already added)
-        if (this.config.credentials) delete this.config.credentials;
+        if (this.config.credentials != null) {
+            delete this.config.credentials;
+        }
 
         // display
         debug("DefaultServient configured with");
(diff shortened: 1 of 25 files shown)


danielpeintner (approved):
That's a great clean-up! 🎉

packages/cli/src/cli.ts
```diff
@@ -307,17 +309,16 @@ const runScripts = async function (servient: DefaultServient, scripts: Array<str
     };
 
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const inspector = require("inspector");
     if (debug && debug.shouldBreak) {
         // Activate inspector only if is not already opened and wait for the debugger to attach
-        !inspector.url() && inspector.open(debug.port, debug.host, true);
+        inspector.url() != null && inspector.open(debug.port, debug.host, true);
```
relu91, line 314: Should this be:
```suggestion
        inspector.url() == null && inspector.open(debug.port, debug.host, true);
```

JKRhb: Thank you, good catch!

JKRhb:
Applied a last tiny improvement in [10f9498](https://github.com/eclipse-thingweb/node-wot/pull/1127/commits/10f949822632861ab238d7d181e04592f0880bcc) – hopefully, this one is now good to go :)

packages/cli/src/cli-default-servient.ts
```diff
@@ -98,10 +98,7 @@ export default class DefaultServient extends Servient {
 
         // apply flags
         if (clientOnly) {
-            if (!this.config.servient) {
-                this.config.servient = {};
-            }
-            this.config.servient.clientOnly = true;
+            this.config.servient ??= { clientOnly };
```
relu91, line 101: are you sure this works as the previous code? In my understanding previously if `this.config.servient` was already defined we just set `clientOnly` property to `true`. With the new code if it is defined we do not override it.

packages/cli/src/cli.ts
```diff
@@ -339,7 +340,7 @@ const runScripts = async function (servient: DefaultServient, scripts: Array<str
         });
     } else {
         // Activate inspector only if is not already opened and don't wait
-        debug && !inspector.url() && inspector.open(debug.port, debug.host, false);
+        debug != null && inspector.url() != null && inspector.open(debug.port, debug.host, false);
```
relu91, line 343: spotted another:
```suggestion
        debug != null && inspector.url() == null && inspector.open(debug.port, debug.host, false);
```

packages/cli/src/cli-default-servient.ts
JKRhb: Oh, you are right, then we guess we need to keep something like this here:

```suggestion
            this.config.servient ??= {};
            this.config.servient.clientOnly = true;
            
```

relu91 (approved):
good to go 🥳

Issue #1128, opened 2023-10-18
AID and ranges


danielpeintner:
AID uses special constructs/terms to represent ranges and min/max values

![image](https://github.com/eclipse-thingweb/node-wot/assets/11281461/60953f10-6efb-4a56-bf6e-fb639a5c8ff9)

PR #1129, opened 2023-10-18
feat: support special AID treatment for ranges

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1128

### DIFF
--- a/packages/td-tools/src/util/asset-interface-description.ts
+++ b/packages/td-tools/src/util/asset-interface-description.ts
@@ -340,12 +340,22 @@ export class AssetInterfaceDescriptionUtil {
                         for (const v of iv.value) {
                             // Binding
                             if (v.idShort === "href") {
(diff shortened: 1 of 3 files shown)


relu91 (approved):
LGTM.

relu91:
It seems we have dangling and random failing macos tests for node 18 😢 . But they are not related to this PR.

JKRhb:
> It seems we have dangling and random failing macos tests for node 18 😢 . But they are not related to this PR.

Yeah, I now restarted the workflow a second time, if it keeps failing then maybe we need to exclude this combination from the CI :/

JKRhb:
Okay, now it worked 🤔 OPC UA seems to be a bit unreliable here :/

relu91:
>Okay, now it worked 🤔 OPC UA seems to be a bit unreliable here :/

Yeah, we have to dig deeper, it might be some un-closed resources.

PR #1132, opened 2023-10-19
docs: update copyright year based on git log

danielpeintner:
Note: We should do it regularly and at latest on Dec 31 every year to get the copyright years right.

### DIFF
--- a/packages/binding-coap/src/coaps-client.ts
+++ b/packages/binding-coap/src/coaps-client.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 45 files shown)


egekorkan (approved):
Only changes are 2022 -> 2023, good to go

JKRhb (approved):

relu91 (approved):
If we have the script we can easily automate this with a GitHub action: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

danielpeintner:
> If we have the script we can easily automate this with a GitHub action: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

The "script" I use is a simple [Java class](https://github.com/eclipse-thingweb/node-wot/tree/master/utils/fix-copyright) that I had handy. Not sure if we should rewrite it in TS/JS and/or setup a scheduled workflow....

Issue #1139, opened 2023-10-24
TDParser - strange behavior with unusual base URI


danielpeintner:
I have encountered a _strange_ behavior with our TDParser.

Let's assume I have the following TD with the unusual `base`.

```
{
    "@context": "https://www.w3.org/2022/wot/td/v1.1",
    "title": "TestModbus",
    "base": "modbus+tcp://##addr##:502/##unitid##",
    "securityDefinitions": {
      "nosec_sc": {
        "scheme": "nosec"
      }
    },
    "security": "nosec_sc",
    "properties": {
      "voltage": {
        "title": "Voltage",
        "type": "number",
        "forms": [
          {
            "href": "/40001?quantity=2"
          }
        ]
      }
    }
  }
```

If I do `TDParser.parseTD(...)` I would assume that `href` for the property `voltage` is
* "/40001?quantity=2"
   or
* "modbus+tcp://##addr##:502/##unitid##/40001?quantity=2"

What I get is "modbus+tcp:///40001?quantity=2"

Somehow parts get stripped off!?

danielpeintner:
It is a problem of this line / library

https://github.com/eclipse-thingweb/node-wot/blob/247b0d48d69962263d994b91d1ff0097c5056be0/packages/td-tools/src/td-parser.ts#L211

What do people think? Is this a bug in the library?
I assume the URL is not valid (not sure if there is a good mean to check that).

I wonder whether we should really do such checks !?

danielpeintner:
The problem is BIGGER than I thought.. it seems that the library does not work for straightforwards base URLs.

e.g., PR  https://github.com/eclipse-thingweb/node-wot/pull/1140
base: "modbus+tcp://192.168.178.123:502/unitid"
relative href: "/40001?quantity=2"
leads to the following absolute URI "modbus+tcp://192.168.178.123:502/40001?quantity=2"

Somehow everything after the port 502 (in this case "/unitid") gets stripped off from the base.

egekorkan:
One part of the problem is the definition of the relative URL. If a relative URL starts with / , it replaces everything starting after port. However, it does not explain why even the hostname (`##addr##:502`) gets removed. The behaviour in the comment just above is correct.

danielpeintner:
> One part of the problem is the definition of the relative URL. If a relative URL starts with / , it replaces everything starting after port.

I see... I missed that 👍

danielpeintner:
[...]

PR #1140, opened 2023-10-24
Absolute URL with base and relative href

danielpeintner:
see https://github.com/eclipse-thingweb/node-wot/issues/1139


danielpeintner:
> By the way, that TD example and test using `##` should not be used. If it is a TM (which is the case here), we should use `{{ }}`

I agree, # has the meaning of fragments and is causing issues in URL processing.

Issue #1138, opened 2023-10-24
OPCUA test issues with MacOS


danielpeintner:
We recently re-introduced the CI tests for Windows and MacOS again (see https://github.com/eclipse-thingweb/node-wot/pull/1121),

By doing so we now have somewhat consistently failings tests for OPCUA on MacOS.
see for example https://github.com/eclipse-thingweb/node-wot/actions/runs/6625217554/job/17995749117?pr=1137#step:6:416

@erossignon do you have an idea what might causing these problems?

erossignon:
I'll take it, 
It seems the vm doesn't have the domain name properly installed
os.hostname() has a ".local" suffix added that causes problem.

erossignon:
After digging, it appears to me that the `/etc/hosts` file doesn't have an entry with the hostname of the machine by default.

Adding  `127.0.0.1  <hostname>` in the /etc/hosts fixes this.

danielpeintner:
resolved by https://github.com/eclipse-thingweb/node-wot/pull/1141

erossignon:
It's pretty annoying that this PR was not merged but has been replaced with a different one that does the same thing by a reviewer of this PR. This is not the find time that this is happening.

Are you still accepting external contributions to the project? Do you value the time and effort of your contributors ?

PR #1141, opened 2023-10-25
binding-opcua:test fix macOS endpoint

erossignon:


.github/workflows/ci.yaml
```diff
@@ -32,6 +32,11 @@ jobs:
             - name: Build
               run: npm run build
 
+            - name: Add missing hostname to /etc/hosts
```
relu91, line 35: Great, can we reference the issue #1138?

relu91 (changes requested):

danielpeintner:
@erossignon Thank you very much for looking into the issue.

As mentioned before I cannot run the opcua tests on Windows either since a while. Seems related to some company network settings. 
```

  1) OPCUA Client
       Y1-0 should read a topic with contentType= application/json:
     Error: Cannot connected to endpoint opc.tcp://<HOSTNAME_MY_MACHINE>:7890
msg = The connection may have been rejected by server,
```

For the record and maybe to help others as well.
By looking at your investigation and fix I did the same on windows updating `%WINDIR%\System32\drivers\etc\hosts` with the following line

`127.0.0.1 <HOSTNAME_MY_MACHINE>`

and the opcua tests are working again 👍
@egekorkan you seemed to have the same issue, right?

danielpeintner (approved):
LGTM

referring to the issue might help others understand the reason

.github/workflows/ci.yaml
JKRhb, line 35: Since it is not possible to add a comment to a workflow step, maybe a comment like this could be an alternative to also have documentation of the rationale for this step within the workflow file?

```suggestion
            # Workaround for issue #1138
            - name: Add missing hostname to /etc/hosts
```

JKRhb (approved):

.github/workflows/ci.yaml
danielpeintner: @relu91 I assume you are okay with the latest changes?

danielpeintner (approved):

relu91 (approved):
Good to go

Issue #1151, opened 2023-11-03
CLI not loading the script


relu91:
As demonstrated by the latest [live coding](https://www.youtube.com/live/2eUOk0oE5vo?si=ZLcT70nNHaj7NoX_&t=2054), when you call `wot-servient script.js` nothing happens and the script is not run.

danielpeintner:
I never run the CLI with `wot-servient`. Anyhow, running a sample like this

`node packages\cli\dist\cli.js examples\scripts\counter.js`

throws the following error.

```
Error: Cannot produce ExposedThing because " + #<Object> could not be cloned.
    at ...\packages\core\dist\wot-impl.js:111:24
    at new Promise (<anonymous>)
```

and my guess is that it has to do with https://github.com/eclipse-thingweb/node-wot/pull/1105.

I will try to take a look unless @JKRhb will look into... please let me know...

PR #1152, opened 2023-11-03
refactor: revert back to *old* way of cloning

danielpeintner:
Note: reverts parts of https://github.com/eclipse-thingweb/node-wot/pull/1105

fixes https://github.com/eclipse-thingweb/node-wot/issues/1151

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,14 @@
         },
         {
             "type": "node",
+            "request": "launch",
+            "name": "Launch Counter",
+            "program": "${workspaceRoot}/packages/cli/dist/cli.js",
+            "args": ["counter.js"],
+            "cwd": "${workspaceRoot}/examples/scripts/"
+        },
+        {
+            "type": "node",
             "request": "attach",
             "name": "Attach to Remote",
             "address": "localhost",
--- a/packages/cli/src/cli-default-servient.ts
+++ b/packages/cli/src/cli-default-servient.ts
@@ -123,7 +123,7 @@ export default class DefaultServient extends Servient {
         }
 
         let coapServer: CoapServer | undefined;
-        if (this.config.servient.clientOnly == null) {
+        if (this.config.servient.clientOnly === false) {
             if (this.config.http != null) {
                 const httpServer = new HttpServer(this.config.http);
                 this.addServer(httpServer);
--- a/packages/core/src/exposed-thing.ts
+++ b/packages/core/src/exposed-thing.ts
@@ -122,7 +122,9 @@ export default class ExposedThing extends TD.Thing implements WoT.ExposedThing {
         this.actions = {};
         this.events = {};
 
-        const deepClonedModel = structuredClone(thingModel);
+        // Deep clone the Thing Model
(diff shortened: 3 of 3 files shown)


danielpeintner:
The issue with starting `node packages\cli\dist\cli.js examples\scripts\counter.js` is gone but the exposed thing immediately shuts down and the "server" part is not active anymore...

danielpeintner:
We run into the case where no server is there...

https://github.com/eclipse-thingweb/node-wot/blob/8659709b76c346ce0c2d5ce768f113bb56e4777c/packages/core/src/servient.ts#L39-L44

2 questions:
* why does this happen?
* why is the warn statement not printed out ?

danielpeintner:
I found the problem, see https://github.com/eclipse-thingweb/node-wot/pull/1152/commits/2976374a3cfe3f2f1143776dc10260b4ad24025e

Question: I wonder whether the clientOnly flag is properly initialed...

https://github.com/eclipse-thingweb/node-wot/pull/1152/files#diff-d50fb68661fd819c0149ceef494c0015de6739c57acd49363e2ba04d6aa0eda3L99-L102

I would rather write something like..

```
        // apply flags
        this.config.servient ??= {};
        this.config.servient.clientOnly = clientOnly;
```

Any thoughts?

JKRhb:
Thank you, @danielpeintner, for providing this fix! After also doing some more research, I discovered that the problem might be related to `vm2`, since the error does not occur if you perform a `structuredClone` of an `ExposedThingInit` before passing it to the `produce` method (which I found pretty odd). Apparently, `vm2` wraps unprocessed JavaScript objects somehow, so that they are not compatible with `structuredClone` anymore. With the old parse-stringify approach, this problem does not seem to be present. 

I guess we could revisit using the more recommended `structuredClone` once again when a replacement for `vm2` has been added to the code. For now, maybe we need to make sure that the problem does not appear in other CLI-related contexts as well.

I am really sorry that this problem arose, I didn't anticipate this kind of behavior from the CLI :/ Really great that you discovered the solution so quickly, though!

JKRhb:
> Question: I wonder whether the clientOnly flag is properly initialed...
> 
> https://github.com/eclipse-thingweb/node-wot/pull/1152/files#diff-d50fb68661fd819c0149ceef494c0015de6739c57acd49363e2ba04d6aa0eda3L99-L102
> 
> I would rather write something like..
> 
> ```
>         // apply flags
>         this.config.servient ??= {};
>         this.config.servient.clientOnly = clientOnly;
> ```
> 
> Any thoughts?

That is definitely a much simpler solution and looks good to me! I think the current state resulted to a certain extent from the version we had before where the `servient` field was not always defined (or only when `clientOnly` was "truthy"). Maybe we need to double-check that this slight logic change would still work but other than that I am very much in favor of this change.

danielpeintner:
> That is definitely a much simpler solution and looks good to me!

@relu91 are you fine with the suggested change as part of this PR also?

relu91 (approved):
Yes good to go, and thanks for digging this down. VM2 is getting problematic :/

PR #1157, opened 2023-11-09
refactor(td-tools/AID): add AID JSON schema

danielpeintner:


sebastiankb (approved):

danielpeintner:
Note: I improved the code to do best effort approach if `base` is missing in TDs (which is required for AID) by looking at property forms.

sebastiankb:
thanks for the update. looks good for merge

Issue #1055, opened 2023-08-11
helper class fetch returning `unknown` ?


danielpeintner:
Should `fetch` return `ThingDescription` or `unknown`?

OR

shall we introduce yet another method `fetchTD` which returns `ThingDescription` and does TD validation also

_Originally posted by @danielpeintner in https://github.com/eclipse-thingweb/node-wot/pull/1049#discussion_r1281515189_

Issue #1062, opened 2023-08-19
Consume and read loop causes memory leak


relu91:
I discovered this memory leak a week ago, but only today I finally got what is the issue. Take this simple code:
```js
const { Servient } = require("@node-wot/core");
const { HttpClientFactory } = require("@node-wot/binding-http");
const { Helpers } = require("@node-wot/core");

let WoTHelpers;
let api;
async function main(){
    const servient = new Servient();
    servient.addClientFactory(new HttpClientFactory());
    api = await servient.start();
    WoTHelpers = new Helpers(servient);
    readloop();
}

async function readloop() {
    const td = await WoTHelpers.fetch("http://localhost:8080/test")
    const testThing = await api.consume(td);
    console.log("Consumed");

    const output = await testThing.readProperty("temperature")
    const value = await output.value();

    console.log(value);
    setTimeout(() => {
        main();
    }, 500);
}

main();
```

As you can see, we are simply polling a remote web thing and reading its `temperature` property. The loop seems pretty harmless but as time goes by the heap starts to inflate and after some days your service will crash for not having free available memory. 

Putting aside the reason why we are continuously fetching the consumed thing (more on that later), the root cause of this memory leak is `ajv` cache. Basically, every time we call the `value` function we call `ajv.compile(schema)` which, as a side effect, stores the result of the compilation inside an internal `Map` cache. The schema object is used as a key of this `Map` hence every time we fetch and consume the Thing Description we are creating a new schema object which for the map is a completely new key. This also triggers another compilation round which has an impact also in the CPU performance. 

One easy solution for our users is simply to always avoid fetch-consume loops (btw the same applies to simply consuming the same Thing Description in a loop because we are cloning it every time). However, in my use case, we were continuously fetching to have some small tolerance in transient services shutdowns ( e.g. if the remote thing updates the consumer service recovers automatically ) .

One solution could be to handle the caching of the Thing Description schemas ourselves (`ajv` provides a good API for managing schemas manually), but this would require some major efforts since, in the current architecture, the `InteractionOutput` has the responsibility of handling schema validation and it has a narrow view of the whole ConsumeThing. We would need to move the compilation of schemas responsibility up to the ConsumedThing object and pass it as a parameter to the  `InteractionOutput` object which only calls the validate function. 

Any thoughts?

egekorkan:
[...]

PR #1166, opened 2023-11-21
feat: add initial requestThingDescription implementation

JKRhb:
This PR adds an initial implementation for the `requestThingDescription` method.

The implementation details probably still need some additional discussion – looking forward to your feedback here :) For now, I extended the `ProtocolClient` interface with a new `requestThingDescription` method that is expected to return a `Content` object that is deserialized and (supposed to be ) validated by the `WoT` implementation.

I think this latter part also touches #1055 – in my opinion, there should be some validation happening so that library users can be sure that they always receive a valid Thing Description. Maybe the algorithm(s) in the Scripting API would need to be adjusted then as well.

For now, I only implemented the protocol-specific `requestThingDescription` method for HTTP and CoAP, if the general design works for you more protocols could be added in follow-up PRs.

### DIFF
--- a/packages/binding-file/src/file-client.ts
+++ b/packages/binding-file/src/file-client.ts
@@ -24,40 +24,48 @@ import path = require("path");
 
 const { debug, warn } = createLoggers("binding-file", "file-client");
 
+/**
+ * Used to determine the Content-Type of a file from the extension in its
+ * {@link filePath} if no explicit Content-Type is defined.
+ *
+ * @param filepath The file paththe Content-Type is determined for.
+ * @returns An appropriate Content-Type or `application/octet-stream` as a fallback.
+ */
+function mapFileExtensionToContentType(filepath: string) {
+    const fileExtension = path.extname(filepath);
+    debug(`FileClient found '${fileExtension}' extension`);
+    switch (fileExtension) {
+        case ".txt":
+        case ".log":
+        case ".ini":
+        case ".cfg":
+            return "text/plain";
+        case ".json":
+            return "application/json";
+        case ".jsontd":
+            return "application/td+json";
+        case ".jsonld":
+            return "application/ld+json";
+        default:
+            warn(`FileClient cannot determine media type for path '${filepath}'`);
+            return "application/octet-stream";
+    }
+}
+
 export default class FileClient implements ProtocolClient {
     public toString(): string {
         return "[FileClient]";
     }
 
+    private async readFile(filepath: string, contentType?: string): Promise<Content> {
+        const resource = fs.createReadStream(filepath);
+        const resourceContentType = contentType ?? mapFileExtensionToContentType(filepath);
+        return new Content(resourceContentType, resource);
+    }
+
     public async readResource(form: Form): Promise<Content> {
-        const filepath = form.href.split("//");
-        const resource = fs.createReadStream(filepath[1]);
-        const extension = path.extname(filepath[1]);
-        debug(`FileClient found '${extension}' extension`);
-        let contentType;
-        if (form.contentType != null) {
-            contentType = form.contentType;
-        } else {
-            // *guess* contentType based on file extension
-            contentType = "application/octet-stream";
-            switch (extension) {
-                case ".txt":
-                case ".log":
-                case ".ini":
-                case ".cfg":
-                    contentType = "text/plain";
-                    break;
-                case ".json":
-                    contentType = "application/json";
-                    break;
-                case ".jsonld":
-                    contentType = "application/ld+json";
-                    break;
-                default:
-                    warn(`FileClient cannot determine media type of '${form.href}'`);
-            }
-        }
-        return new Content(contentType, resource);
+        const filepath = form.href.split("//")[1];
+        return this.readFile(filepath, form.contentType);
     }
 
     public async writeResource(form: Form, content: Content): Promise<void> {
@@ -76,7 +84,7 @@ export default class FileClient implements ProtocolClient {
      * @inheritdoc
      */
     public async requestThingDescription(uri: string): Promise<Content> {
-        throw new Error("Not implemented");
+        return this.readFile(uri, "application/td+json");
     }
 
     public async subscribeResource(
--- a/packages/binding-file/src/file-client.ts
+++ b/packages/binding-file/src/file-client.ts
@@ -72,6 +72,13 @@ export default class FileClient implements ProtocolClient {
         throw new Error("FileClient does not implement unlink");
     }
 
+    /**
+     * @inheritdoc
+     */
+    public async requestThingDescription(uri: string): Promise<Content> {
+        throw new Error("Not implemented");
+    }
+
     public async subscribeResource(
         form: Form,
         next: (value: Content) => void,
--- a/packages/binding-file/src/file-client.ts
+++ b/packages/binding-file/src/file-client.ts
@@ -28,7 +28,7 @@ const { debug, warn } = createLoggers("binding-file", "file-client");
  * Used to determine the Content-Type of a file from the extension in its
  * {@link filePath} if no explicit Content-Type is defined.
  *
- * @param filepath The file paththe Content-Type is determined for.
+ * @param filepath The file path the Content-Type is determined for.
  * @returns An appropriate Content-Type or `application/octet-stream` as a fallback.
  */
 function mapFileExtensionToContentType(filepath: string) {
--- a/packages/binding-file/src/file-client.ts
+++ b/packages/binding-file/src/file-client.ts
@@ -64,7 +64,7 @@ export default class FileClient implements ProtocolClient {
     }
 
     public async readResource(form: Form): Promise<Content> {
-        const filepath = form.href.split("//")[1];
+        const filepath = new URL(form.href).pathname;
         return this.readFile(filepath, form.contentType);
     }
 
--- a/packages/binding-mqtt/src/mqtt-client.ts
+++ b/packages/binding-mqtt/src/mqtt-client.ts
@@ -161,6 +161,13 @@ export default class MqttClient implements ProtocolClient {
         }
     }
 
+    /**
+     * @inheritdoc
+     */
+    public async requestThingDescription(uri: string): Promise<Content> {
+        throw new Error("Method not implemented");
+    }
+
     public async start(): Promise<void> {
         // do nothing
     }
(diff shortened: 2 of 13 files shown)


packages/binding-mqtt/src/mqtt-client.ts
```diff
@@ -161,6 +161,13 @@ export default class MqttClient implements ProtocolClient {
         }
     }
 
+    /**
+     * @inheritdoc
+     */
+    public async requestThingDescription(uri: string): Promise<Content> {
+        throw new Error("Method not implemented");
```
egekorkan, line 168: Technically, this can be implemented since an MQTT Thing exposed by node-wot publishes its TD with retain. See https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-mqtt/src/mqtt-broker-server.ts#L141 It is not standardized afaik but I do not know what speaks against doing a subscription, getting a TD and unsubscribing. Of course it can be done in another PR, I leave it up to you @JKRhb

JKRhb: Oh, yeah, very good point, I implemented something similar in `dart_wot` so I can confirm that this is working :) Maybe I would open a follow-up PR for this feature, though, since it might require some discussion regarding the URL-to-topic mapping 🤔

egekorkan: Yup sure. Interestingly, this is also mandated by sparkplug with the birth messages (same as expose here). More reading is possible at https://sparkplug.eclipse.org/specification/version/3.0/documents/sparkplug-specification-3.0.0.pdf at page 21. Also, the newer version of it allows any payload to be published there (wink wink TD). 

In the end, as long as the developer provides the URL to fetch from, I think it is fine and no need for big discussions.

packages/core/src/wot-impl.ts
```diff
@@ -39,8 +40,20 @@ export default class WoTImpl {
...
     }
 
+    /** @inheritDoc */
     async requestThingDescription(url: string): Promise<WoT.ThingDescription> {
-        throw new Error("not implemented");
+        const uriScheme = new URL(url).protocol.split(":")[0];
+        const client = this.srv.getClientFor(uriScheme);
+        const result = await client.requestThingDescription(url);
+
+        const value = ContentManager.contentToValue({ type: result.type, body: await result.toBuffer() }, {});
```
relu91, line 49: Unrelated comment. I found it strange that the function ContentToValue does not accept directly the `result` variable which is type Content. Something to record in an issue for future improvements.

packages/binding-file/src/file-client.ts
```diff
@@ -72,6 +72,13 @@ export default class FileClient implements ProtocolClient {
         throw new Error("FileClient does not implement unlink");
     }
 
+    /**
+     * @inheritdoc
+     */
+    public async requestThingDescription(uri: string): Promise<Content> {
+        throw new Error("Not implemented");
```
relu91, line 79: Is there a blocker to implement this or is it just a time constraint? I think we can jus read the file pointed by the file URI.

relu91 (commented):
I would fix Ege's point and I left a couple of comments below. Moreover, I wondering if we can somehow track the different TODOs comments before merging. Maybe when we reched two accepts we can create the issues and do a quick update of the comments with a link to the issues? Finally, maybe I missed why there is so much discrepancy between `coaps` and `coap` implementation?

packages/core/src/wot-impl.ts
JKRhb: Yeah, that is indeed a bit strange, apparently `contentToValue` expects input that implements the `ReadContent` interface which is not compatible with the regular `Content` class 🤔 It would definitely be better to have one unified `Content` API that is used across the codebase. Opening an issue for that sounds like a good idea to me!

packages/binding-mqtt/src/mqtt-client.ts
JKRhb: Hmm, having another look at the MQTT client implementation after @relu91's comment I noticed that it actually does not even implement the `readResource` method where I think a similar logic would be used. So my proposal would be to implement both methods at the same time to arrive at a uniform implementation and keep the discussion focused – would that be okay for you, @relu91? Otherwise, adding the implementation here would also not be a problem, though :)

packages/binding-file/src/file-client.ts
JKRhb: I added an implementation now in 3a7c741e48a155b01039b3745db65fedd402755f :) I think there might actually be some more room for simplifying the `FileClient` implementation, though, since the file extension determination is not actually needed when we assume a default Content-Type of `application/json` for `Form`s. However, this would constitute a breaking change.

JKRhb:
> Finally, maybe I missed why there is so much discrepancy between `coaps` and `coap` implementation?

That is actually because we are currently using two different libraries for CoAP (`node-coap`) and CoAPS (`node-coap-client`) since `node-coap` does not support DTLS :/ A mid-term goal for me personally is definitely to arrive at a unified implementation that covers not only CoAP and CoAPS but also CoAP over TCP/TLS and WebSockets. For the time being, though, I think we need to live with this doubled CoAP implementation.

packages/core/src/wot-impl.ts
```diff
@@ -39,8 +41,19 @@ export default class WoTImpl {
...
-        throw new Error("not implemented");
+        const uriScheme = new URL(url).protocol.split(":")[0];
+        const client = this.srv.getClientFor(uriScheme);
+        const content = await client.requestThingDescription(url);
+
+        const value = ContentManager.contentToValue({ type: content.type, body: await content.toBuffer() }, TDSchema);
+
+        if (value instanceof Object) {
+            return value as WoT.ThingDescription;
+        }
```
JKRhb, line 54: I got rid of the TODO regarding validation now by passing the `TDSchema` to the `contentToValue` method alongside the TD content. I hope that makes sense – more validation could probably be added later.

packages/binding-file/src/file-client.ts
```diff
@@ -24,40 +24,48 @@ import path = require("path");
...
-                    break;
-                case ".jsonld":
-                    contentType = "application/ld+json";
-                    break;
-                default:
-                    warn(`FileClient cannot determine media type of '${form.href}'`);
-            }
-        }
-        return new Content(contentType, resource);
+        const filepath = form.href.split("//")[1];
```
danielpeintner, line 67: Would be failing if `href` split returns array with zero size, right? Are we fine with it?

danielpeintner (approved):
I think for an initial implementation it is fine as is.

I added some comments below

packages/binding-file/src/file-client.ts
```diff
@@ -24,40 +24,48 @@ import path = require("path");
 
 const { debug, warn } = createLoggers("binding-file", "file-client");
 
+/**
+ * Used to determine the Content-Type of a file from the extension in its
+ * {@link filePath} if no explicit Content-Type is defined.
+ *
+ * @param filepath The file paththe Content-Type is determined for.
```
danielpeintner, line 31: the comment is somehow broken

JKRhb: Thank you :) Should be fixed by 907b429a5e4400de62a5999e59a24c18850e62a4

packages/binding-file/src/file-client.ts
JKRhb: At the moment, this behavior is consistent with the previous version, I just refactored it a bit (previously, `filepath` was an array that got accessed this way twice). Adding a check or a fallback value sounds good to me, though – should this be done in this PR? Or should we follow up with another one to do so?

packages/core/src/wot-impl.ts
```diff
@@ -39,8 +41,19 @@ export default class WoTImpl {
         throw new Error("not implemented");
     }
 
+    /** @inheritDoc */
     async requestThingDescription(url: string): Promise<WoT.ThingDescription> {
-        throw new Error("not implemented");
+        const uriScheme = new URL(url).protocol.split(":")[0];
```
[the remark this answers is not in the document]
JKRhb: Good point, I now replaced this line with the `extractScheme` helper method in b67162da2984f44c487054258d9c74736b5779e8, which makes the implementation cleaner and more consistent with the rest of the codebase.

packages/binding-file/src/file-client.ts
danielpeintner: I "think" we can also do it in this PR

JKRhb: I now replaced the splitting in this line with accessing a parsed `URL` object. In this context, I noticed that the previous implementation might not have been completely correct since it used the *host* component of the parsed URL instead of the *path* component (which, at least according to [Wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme) is "never correct"). It seems like the README of the file binding also needs to be adjusted in this regard.

packages/binding-file/src/file-client.ts
relu91: It is good! I would like to keep that auto detection of content type... TD 2.0 might support other formats than the default.

packages/core/src/wot-impl.ts
relu91: As I recall `contentToValue` method does not do the validation. It uses the data schema only as a hint when converting. The proper way to handle this should use AJV here. But we have to think this correctly, given #1062 .

packages/binding-file/src/file-client.ts
relu91: I remember that parsing File URIs is problematic. A couple of unit tests might be good. Can you add them?

JKRhb: Sure, will do :)

packages/core/src/wot-impl.ts
JKRhb: In 5aeb45dde8507cb205e5f5c406e0b1f7528a15f1, I now incorporated the validation code used by @danielpeintner in #1169 instead

packages/binding-file/src/file-client.ts
JKRhb: I now included the file client code in a separate PR: #1175 Once we have merged that one, I would perform a rebase to keep a clean and reviewable diff in this PR :)

danielpeintner: Makes sense 👍

danielpeintner (approved):
LGTM

relu91 (approved):
Good enough for now. Let's improve it with future iterations!

PR #1169, opened 2023-11-23
feat(core): add helper function to fetch a TD

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1055

### DIFF
--- a/packages/binding-http/test/http-client-test.ts
+++ b/packages/binding-http/test/http-client-test.ts
@@ -22,12 +22,20 @@ import chai, { expect, should } from "chai";
 
 import * as http from "http";
 
-import { Content, DefaultContent, ContentSerdes, createLoggers, ProtocolServer } from "@node-wot/core";
+import {
+    Content,
+    DefaultContent,
+    ContentSerdes,
+    createLoggers,
+    ProtocolServer,
+    Servient,
+    Helpers,
+} from "@node-wot/core";
 
 import { Readable } from "stream";
 
 import HttpClient from "../src/http-client";
-import { HttpForm } from "../src/http";
+import { HttpForm, HttpClientFactory } from "../src/http";
 
 import express from "express";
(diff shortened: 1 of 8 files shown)


relu91:
Shouldn't we deprecate this helper function and favor the new #1166?

danielpeintner:
> Shouldn't we deprecate this helper function and favor the new #1166?

W.r.t. `fetchTD()` ... there should be just one piece of code dealing with it, true, point taken 👍.

I am wondering what we should do w.r.t. to the existing `fetch()`. Remove it? Helpers usually offers static functions, the only exception is `fetch()` which needs a `Servient` to work. I would think we should remove the function "on the long run"...

Moreover, there is the interface `Resolver` which is used by ThingModel code.. this is somewhat not very clear to me.

relu91:
Yes, I would deprecate the `fetch` and `fetchTD` functions in favor of `requestThingDescription`. 

> Moreover, there is the interface Resolver which is used by ThingModel code.. this is somewhat not very clear to me.

The interface is there to not depend on the Core from the ThingModelHelpers pov. Users can decide to use the ThingModel helpers without using the Core package and implement their own resolver. I think we can appreciate this choice when we move the tool to a new repo.

danielpeintner:
> Yes, I would deprecate the `fetch` and `fetchTD` functions in favor of `requestThingDescription`.

deprecate `fetch` is possible but at the moment it is still used.
Let's wait till https://github.com/eclipse-thingweb/node-wot/pull/1166 has been integrated and take a look to replace all usages.
Anyhow, ThingModel still uses `fetch`.. I believe.. but let's see...

deprecate `fetchTD` is not necessary since it hasn't been pushed yet.

danielpeintner:
Closing in favor of https://github.com/eclipse-thingweb/node-wot/pull/1183

PR #1174, opened 2023-11-28
feat(core): add semantic composed thing

danielpeintner:
Based on the work started in https://github.com/eclipse-thingweb/node-wot/tree/semantic-layer I would like to continue working on providing a `SemanticComposedThing` as a wrapper around `ConsumedThing`.

The idea is that consumers no longer need to know the property `name` to get the data they want. Instead it should be possible to use `@type` (e.g., read property with `"@type": "iot:BinarySwitchData"`)

Some thoughts:
* node-wot could return SemanticComposedThing by default so that people already _now_ can make use of this functionality
* this could be an extensions/input of the scripting API

Note: at the moment the PR is still in draft state since there are many open questions (see code) especially if we want to properly define input/output data. Comments are welcome!

### DIFF
--- a/packages/core/src/semantic-consumed-thing.ts
+++ b/packages/core/src/semantic-consumed-thing.ts
@@ -1,0 +1,89 @@
+/********************************************************************************
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the W3C Software Notice and
+ * Document License (2015-05-13) which is available at
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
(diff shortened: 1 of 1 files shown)


relu91:
Some early feedback without looking at the code. My gut feeling is that this should be an extension to node-wot. Sort of a richer framework that we can build on top of the raw API. I believe that node-wot should stay as close as possible to the Thing Description interaction model abstraction. Anyhow I think the extension is a great idea, but I'd like to see it as a standalone project that uses node-wot rather than another package inside this project.

danielpeintner:
> Some early feedback without looking at the code. My gut feeling is that this should be an extension to node-wot. Sort of a richer framework that we can build on top of the raw API.

This is good feedback and my response to it is as follows.
If we never "want" that to be standardized in the Scripting API I totally agree we should move  out the extension. If we "think" this could make it into the spec I would got for it.
Anyhow, for the time being I think I can develop it as part of node-wot and at latest when we plan to merge this PR we need to decide...

egekorkan:
> If we never "want" that to be standardized in the Scripting API

From some discussions in the TD/Binding TF at WoT WG, I remember others (@mjkoster if I am not wrong) having interest in this so it can indeed be standardized.

PR #1175, opened 2023-11-28
feat!: improve file client implementation

JKRhb:
As a spinoff from #1166, this PR extends the file client implementation with support for the `writeproperty` operation and improves the handling of `file://` URLs. As the determination of the Content-Type from the file extension was somewhat deviating from the standard (as this ignored the default Content-Type implied by a TD form), I removed it.

This last aspect might be a bit controversial, but I think this should be the better approach to have an implementation that is extendable with additional Content-Types by users who can simply register them with the `ContentSerdes`. However, I looking forward to feedback not only in this regard :)


JKRhb:
Had to apply some Windows-related fixes, the PR should now be ready for review :)

packages/binding-file/test/.gitignore
```diff
@@ -0,0 +1 @@
+test.*
```
danielpeintner, line 1: You need that since you create test files, right? Am I wrong?
Does it make sense to add a comment?

packages/binding-file/src/file-client.ts
```diff
@@ -17,51 +17,39 @@
...
+        const filePath = fileURLToPath(form.href);
+        const contentType = form.contentType ?? ContentSerdes.DEFAULT;
+
+        return this.readFromFile(filePath, contentType);
+    }
+
     public async writeResource(form: Form, content: Content): Promise<void> {
-        throw new Error("FileClient does not implement write");
+        const filePath = fileURLToPath(form.href);
+        content.toBuffer();
```
danielpeintner, line 47: needed?

danielpeintner (approved):
LGTM.

Added 2 comments...

packages/binding-file/README.md
```diff
@@ -92,7 +92,7 @@ servient.addClientFactory(new FileClientFactory(null));
 
 let wotHelper = new Helpers(servient);
 wotHelper
-    .fetch("file://TD.jsonld")
+    .fetch("file:///TD.jsonld")
```
relu91, line 95: Should we change this example using the new official `requestThingDescription` function?

packages/binding-file/src/file-client.ts
```diff
@@ -17,51 +17,39 @@
...
-        }
+    private async readFromFile(filePath: string, contentType: string) {
+        debug(`Reading file of Content-Type ${contentType} from path ${filePath}.`);
+        const resource = fs.createReadStream(filePath);
         return new Content(contentType, resource);
     }
 
+    public async readResource(form: Form): Promise<Content> {
+        const filePath = fileURLToPath(form.href);
+        const contentType = form.contentType ?? ContentSerdes.DEFAULT;
```
relu91, line 40: I know this is correct behavior, but it might be error-prone... people will try to serve files without expressing the content-type thinking that the client could figure it himself. Anyhow, just a comment no action is required here.

packages/binding-file/test/.gitignore
relu91: + 1 for a comment. I was surprised to find here the ignore too.

packages/binding-file/src/file-client.ts
danielpeintner: Maybe we can add a debug comment "saying that"... a la
... non contentType set, picked JSON by default

packages/binding-file/test/.gitignore
JKRhb: Sure, will add that :) In theory, we could also remove the `.gitignore` since the test files are cleaned up after the tests are finished, but I thought it might be good to avoid committing any test artifacts here.

packages/binding-file/README.md
JKRhb: Good point, I would probably add that to #1166, though, alongside the rest of the file-related implementation of the new method.

packages/binding-file/src/file-client.ts
JKRhb: Thank you, good catch :) Fixed by b6a364a7e10885f1e550070a966d6e222dd96d74

packages/binding-file/test/.gitignore
relu91: yes let's keep it to be 100% :)

JKRhb: Would this work for you?

```suggestion
# Ignore auxiliary files created by the FileClient implementation tests
test.*
```

packages/binding-file/src/file-client.ts
JKRhb: Tried to incorporate this in 14c178039c2537b12c97be60c80cb53dfa14270b, let me know if that works for you :)

In general, I was wondering if we should maybe move the default Content-Type determination to the `Form` class (maybe as part of a `get` method) to have a consistent behavior across the codebase. I would make an initial proposal in a follow-up PR that we could discuss there if that's okay for you.

packages/binding-file/README.md
JKRhb: Ah, okay, nevermind, as #1166 got already merged ;) Then I will add it to this PR instead :)

JKRhb: I added an initial update of the example in 1b79e91ea7696b86429f028fa0a0db84fdd88e73, haven't tested it yet, though.

JKRhb: I now performed some tests and actually found a few bugs which I fixed in 3793b6ecbb76883e91dd2bcf25dee88d87cea1c5.

packages/binding-file/src/file-client.ts
```diff
@@ -17,59 +17,43 @@
...
+
+        const resource = fs.createReadStream(filePath);
+        return new Content(contentType, resource);
     }
 
     public async readResource(form: Form): Promise<Content> {
-        const filepath = new URL(form.href).pathname;
-        return this.readFile(filepath, form.contentType);
+        const formContentType = form.contentType;
+        if (form.contentType == null) {
```
danielpeintner, line 42:
```suggestion
        if (formContentType == null) {
```

I know it is the same but since you have the variable already it makes more sense to me

JKRhb: Thank you, good catch :)

packages/binding-file/test/.gitignore
danielpeintner: Looks good. Maybe the comment can be even next to the test.* entry.. just in case other ignores are added below for some other reasons but we might not need to bother now 🙈

danielpeintner: I am fine either way also. Since it is your PR I let you decide ;-)

relu91 (approved):
Good to go now

PR #1176, opened 2023-11-29
refactor: use a default contentType of application/json in Forms

JKRhb:
As discussed in https://github.com/eclipse-thingweb/node-wot/pull/1175#discussion_r1409109652, this PR includes a proposal for making the `contentType` of the `Form` class non-nullable by defaulting to `application/json` when no Content-Type is passed to the constructor.

This change has the nice effect of getting rid of a lot of nullish coalescing which is currently required to arrive at the default value.


JKRhb:
Hmm, it seems as if for this to work, the `contentType` field on the [FormElementBase](https://github.com/w3c/wot-scripting-api/blob/521e7d1cefc49cebab16e001f258b538b51b16c7/typescript/thing-description/thing-description.d.ts#L283-L294) interface in the `wot-typescript-definitions` would need to be changed as well.

PR #1178, opened 2023-11-30
ci: add a total timeout of 30 minutes

JKRhb:
In the context of #1175, I noticed that the codecov step was somehow hanging for over 2 hours without causing the CI to fail. To address the issue, this PR proposes to add an additional "total" timeout value of 30 minutes for the whole job which should be enough to cover all runs that work as intended.


relu91 (approved):
ok 👍

PR #1180, opened 2023-12-04
chore: upgrade ts-node to 10.9.1

### LINKED ISSUE #1179: Setting debug level not working
On our vserver I am not able to set the debugging level with

`DEBUG=node-wot* npm start`

I get the following error. Is it a node-wot issue  or somewhere else?

```
node-wot/node_modules/typescript/lib/typescript.js:42537
        ts.Debug.assert(typeof typeReferenceDirectiveName === "string", "Non-string value passed to `ts.resolveTypeReferenceDirective`, likely by a wrapping package working with an outdated `resolveTypeReferenceDirectives` signature. This is probably not a problem in TS itself.");
                 ^
Error: Debug Failure. False expression: Non-string value passed to `ts.resolveTypeReferenceDirective`, likely by a wrapping package working with an outdated `resolveTypeReferenceDirectives` signature. This is probably not a problem in TS itself.

```

JKRhb:
Might resolve #1179.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11651,58 +11651,7 @@
                 "wot-servient": "bin/index.js"
             },
             "optionalDependencies": {
-                "ts-node": "10.1.0"
-            }
-        },
-        "packages/cli/node_modules/diff": {
-            "version": "4.0.2",
-            "resolved": "https://registry.npmjs.org/diff/-/diff-4.0.2.tgz",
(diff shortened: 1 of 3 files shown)


danielpeintner (approved):
Works on the Eclipse vserver as expected 👍

relu91 (approved):
LGTM

Issue #1159, opened 2023-11-10
CI: Combination of MacOS and MQTT failing


danielpeintner:
The recent days I noticed that MQTT tests are failing for for CI builds (I think mainly on MacOs).

e.g.,
* https://github.com/eclipse-thingweb/node-wot/actions/runs/6823627505/job/18557909297#step:8:401
* https://github.com/eclipse-thingweb/node-wot/actions/runs/6732890771/job/18300446252?pr=1149#step:7:401
* ...

Note: re-running the CI for failed jobs usually works.

pinging [@egekorkan](https://github.com/egekorkan) [@sebastiankb](https://github.com/sebastiankb) [@hasanheroglu](https://github.com/hasanheroglu) as MQTT mainatainers

danielpeintner:
Note: Once the issue is resolved we should remove the skipping tests in https://github.com/eclipse-thingweb/node-wot/pull/1181

danielpeintner:
Since the MQTT tests are still failing I wonder whether we should _re-activate_ https://github.com/eclipse-thingweb/node-wot/pull/1181

Note: Tests are not failing on MacOS only, also on Ubuntu/Windows etc..
see https://github.com/eclipse-thingweb/node-wot/actions/runs/7260776357/job/19780615145

danielpeintner:
It is still very much annoying that MQTT tests fail randomly. Are we okay to skip MQTT tests for now or do you prefer to see for each PR etc failings tests?

egekorkan:
Done as of #1272

PR #1181, opened 2023-12-04
test: skip MQTT broker test on macOS

JKRhb:
This PR proposes to skip one of the MQTT tests on macOS until the underlying issue (#1159) is fixed.

### DIFF
--- a/packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
+++ b/packages/binding-mqtt/test/mqtt-client-subscribe-test.integration.ts
@@ -109,7 +109,13 @@ describe("MQTT client implementation", () => {
         });
     }).timeout(20000);
 
-    it("should expose via broker using mqtts", (done: Mocha.Done) => {
+    it("should expose via broker using mqtts", function (done: Mocha.Done) {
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):
I agree, it is very annoying to see failings tests all the time.
The approach seems fine for now since we skip tests on MacOS only

danielpeintner:
@JKRhb is there a reason why it is still marked as "work in progress" ?

danielpeintner:
FYI: MQTT tests are failing on Windows also, see https://github.com/eclipse-thingweb/node-wot/actions/runs/7098842840/job/19321848426#step:8:414

relu91:
I'm working on MQTT client, let's put this on hold.

JKRhb:
> @JKRhb is there a reason why it is still marked as "work in progress" ?

The tests were failing somehow when I was working on them yesterday, but apparently, 3e5fecc actually resolved the issue 😅 But if you are working on an actual fix, @relu91, then we can also simply close this one – feel free to do so :)

JKRhb:
I think with #1184, we can consider this one superseded :)

PR #1183, opened 2023-12-05
refactor: favor WoT.requestThingDescription instead of general fetch

danielpeintner:
I think we should use `requestThingDescription()` instead of the generic `fetch()` everywhere.

Please let me know whether this change makes sense. 

fixes https://github.com/eclipse-thingweb/node-wot/issues/1055

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,20 +186,15 @@ const { HttpClientFactory } = require('@node-wot/binding-http');
 const servient = new Servient();
 servient.addClientFactory(new HttpClientFactory(null));
 
-WoT.requestThingDescription("http://localhost:8080/counter").then(async (td) => {
-    try {
-        const WoT = await servient.start();
-        // Then from here on you can consume the thing
-        let thing = await WoT.consume(td);
-        thing.observeProperty("count", async (data) => { console.log("count:", await data.value()); });
-        for (let i = 0; i < 5; i++) {
-            await thing.invokeAction("increment");
-        }
+servient.start().then(async (WoT) => {
+    const td = await WoT.requestThingDescription("http://localhost:8080/counter");
+    // Then from here on you can consume the thing
+    let thing = await WoT.consume(td);
+    thing.observeProperty("count", async (data) => { console.log("count:", await data.value()); });
+    for (let i = 0; i < 5; i++) {
+        await thing.invokeAction("increment");
     }
-    catch (err) {
-        console.error("Script error:", err);
-    }
-}).catch((err) => { console.error("Fetch error:", err); });
+}).catch((err) => { console.error("Error:", err); });
 ```
 
 If you execute both scripts you will see `count: ${count}` printed 5 times. We host a more complex version of this example at [http://plugfest.thingweb.io/examples/counter.html](http://plugfest.thingweb.io/examples/counter.html) and you can find the source code in the [counter example](./examples/browser) folder. You can also find more examples in the [examples folder](./examples/scripts) for JavaScript and in the [examples folder](./packages/examples/) for TypeScript. Finally, for your convenience, we host a set of online Things that you can use to test your applications. You can find more information about them in the [Online Things](#online-things) section.
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ servient.start().then(async (WoT) => {
     for (let i = 0; i < 5; i++) {
         await thing.invokeAction("increment");
     }
-}).catch((err) => { console.error("Error:", err); });
+}).catch((err) => { console.error(err); });
 ```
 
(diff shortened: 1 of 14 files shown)


JKRhb:
Thank you, @danielpeintner, that looks pretty good already :) Let us know if we can support you somehow.

danielpeintner:
https://github.com/eclipse-thingweb/node-wot/pull/1183/commits/598140cf8425ed215b783917906c4679d14a93c7 adds also the changes/updates w.r.t. documentation.

Something I noticed. We seem to use different import styles.

e.g. HTTP
https://github.com/eclipse-thingweb/node-wot/blob/7334567c2c531dfec25cab0b5d9a9aa943176167/packages/binding-http/README.md?plain=1#L29-L30

vs. CoAP
https://github.com/eclipse-thingweb/node-wot/blob/7334567c2c531dfec25cab0b5d9a9aa943176167/packages/binding-coap/README.md?plain=1#L29-L30

I think we should align the style. Any preference?

packages/binding-coap/README.md
```diff
@@ -26,18 +26,15 @@ The Thing Description is located under the following CoAP URI `coap://plugfest.t
...
 const { CoapClientFactory } = require("@node-wot/binding-coap");
 
 // create Servient and add CoAP binding
 const servient = new Servient();
 servient.addClientFactory(new CoapClientFactory());
 
-const wotHelper = new Helpers(servient);
-wotHelper
-    .fetch("coap://plugfest.thingweb.io:5683/testthing")
+WoT.requestThingDescription("coap://plugfest.thingweb.io:5683/testthing")
```
JKRhb, line 36: I think with the new approach, the `servient` needs to be started first t get the `WoT` object, right?

```ts
servient.start().then(async (WoT) => {
    const td = WoT.requestThingDescription("coap://plugfest.thingweb.io:5683/testthing");
```

danielpeintner: Good point, will fix soon.. I simply string replaced it 🙈

README.md
```diff
@@ -180,27 +180,21 @@ Now supposing you want to interact with the device, you have to consume its Thin
...
-        console.error("Script error:", err);
+
+servient.start().then(async (WoT) => {
+    const td = await WoT.requestThingDescription("http://localhost:8080/counter");
+    // Then from here on you can consume the thing
+    let thing = await WoT.consume(td);
+    thing.observeProperty("count", async (data) => { console.log("count:", await data.value()); });
+    for (let i = 0; i < 5; i++) {
+        await thing.invokeAction("increment");
     }
```
danielpeintner, line 196: Question: Should we keep the "internal" try catch ?

JKRhb: Hmm, since there is a `.catch` down below, keeping it as it is should be fine, I suppose 🤔 I think we could also consider removing the `"Error:"` prefix, since the type of error should be part of the output if I am not mistaken.

JKRhb (approved):
LGTM! :) Maybe one last thing, though: Should we add a [`@deprecated` tag](https://jsdoc.app/tags-deprecated) to the `fetch` method to guide people to the new `requestThingDescription` method?

danielpeintner:
> LGTM! :) Maybe one last thing, though: Should we add a [`@deprecated` tag](https://jsdoc.app/tags-deprecated) to the `fetch` method to guide people to the new `requestThingDescription` method?

I tend to think this is a _generic_ helper method that can be used for any other purpose as well.. we are not claiming is is meant to fetch TDs only. Hence I would argue we can/should keep it. What do others think?

relu91:
> I tend to think this is a generic helper method that can be used for any other purpose as well.. we are not claiming it is meant to fetch TDs only. Hence I would argue we can/should keep it as is. What do others think?

My take is that we have already the fetch function for that. Now it is supported in [Node.js](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html) too, so no need to define yet another generic fetching function.

relu91 (approved):
Other than the comment above I think we can merge it.

danielpeintner:
> My take is that we have already the fetch function for that. Now it is supported in [Node.js](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html) too, so no need to define yet another generic fetching function.

Is the `fetch` in node-wot really yet another fetch? It is about any protocol while the default fetch is about HTTP only?
Anyhow, If someone feels strongly about I can add [@deprecated tag](https://jsdoc.app/tags-deprecated) ...
Anyhow, we should definitely add a statement that we have a `WoT.requestThingDescription()` method that is better suited for requesting TDs. Any opinion? As said, I am fine either way ;-)

relu91:
> Is the fetch in node-wot really yet another fetch? It is about any protocol while the default fetch is about HTTP only?

yeah good point, but I would argue that the old `fetch` was actually an early version (or better prototype) of `requestThingDescription` function. Anyhow we can talk about this in an issue. Let's merge this meanwhile.

PR #1185, opened 2023-12-07
test(binding-file: use asynchronous unlink for file deletion

JKRhb:
This PR might resolve the issues we've been observing with the file client tests.


JKRhb:
The only thing left failing are the MQTT tests now but that should be resolved by #1184 – which PR should we merge first?

relu91 (approved):
Let's go with this if @danielpeintner is also ok.

danielpeintner (approved):

PR #1186, opened 2023-12-07
feat(core): implement `exploreDirectory` method

JKRhb:
This PR starts implementing the `exploreDirectory` method that has recently been added to the Scripting API specification.

### DIFF
--- a/packages/core/src/validation.ts
+++ b/packages/core/src/validation.ts
@@ -14,7 +14,7 @@
  ********************************************************************************/
 
 import TDSchema from "wot-thing-description-types/schema/td-json-schema-validation.json";
-import Ajv, { ValidateFunction } from "ajv";
+import Ajv from "ajv";
 
 export function createWotAjvInstance() {
     return (
@@ -33,7 +33,7 @@ export function createWotAjvInstance() {
 }
 
 const ajv = createWotAjvInstance();
-const validateTd = ajv.compile(TDSchema) as ValidateFunction;
+const validateTd = ajv.compile<WoT.ThingDescription>(TDSchema);
 
 export function isThingDescription(input: unknown): input is WoT.ThingDescription {
     return validateTd(input);
--- a/packages/core/src/validation.ts
+++ b/packages/core/src/validation.ts
@@ -13,28 +13,8 @@
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
 
-import TDSchema from "wot-thing-description-types/schema/td-json-schema-validation.json";
-import Ajv from "ajv";
-
-export function createWotAjvInstance() {
-    return (
-        new Ajv({ strict: false })
-            // RegExps taken from https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts
-            .addFormat(
-                "iri-reference",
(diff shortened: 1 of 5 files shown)


packages/core/src/wot-impl.ts
```diff
@@ -38,7 +38,15 @@ export default class WoTImpl {
 
     /** @inheritDoc */
     async exploreDirectory(url: string, filter?: WoT.ThingFilter): Promise<WoT.ThingDiscoveryProcess> {
-        throw new Error("not implemented");
+        const directoyThingDescription = await this.requestThingDescription(url);
+        const consumedDirectoy = await this.consume(directoyThingDescription);
+
+        const thingsProperty = await consumedDirectoy.readProperty("things");
+        const value = await thingsProperty.value();
```
JKRhb, line 45: Do you think this general approach (first requesting the directory's TD, then consuming it, and reading the `things` property) makes sense here? I think the [algorithm](https://w3c.github.io/wot-scripting-api/#the-exploredirectory-method) in the Scripting API specification is not going into much detail here.

danielpeintner: That's the way I understand it as well 👍

JKRhb:
I think this PR could now be ready for an initial review :) Even though there currently are some typescript quirks regarding the handling of `ReadableStream`s, interacting with a local instance of Zion to retrieve TDs already worked.

One major piece is currently missing the validation of the retrieved TDs – did we already have a function handy that could be used for that?

danielpeintner:
> One major piece is currently missing the validation of the retrieved TDs – did we already have a function handy that could be used for that?

[ThingModelHelpers](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/td-tools/src/thing-model-helpers.ts) has `isThingModel` and `validateThingModel`. We might want to have something similar for ThingDescription.

danielpeintner (commented):
Do you see a possibility to have tests? Using Zion or so would require the service to be always up which may be an issue, right?

relu91:
> Do you see a possibility to have tests? Using Zion or so would require the service to be always up which may be an issue, right?

Having a second look, doing tests with Zion means that we need to add them in the CLI because we would need to have also the binding templates up and running. It complicates a little bit the setup but it might be good to create a sort of end-to-end testing in a single place. In other words, we won't need to test the operations **per** binding but we could have a single code base in one place. Probably it is too much work for a single PR. What do you think?

JKRhb:
I think for now I would try to mock the directory API, but having integration or end-to-end testing with Zion in place in the mid-term sounds like a great idea :)

packages/core/src/validation.ts
```diff
@@ -0,0 +1,41 @@
...
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the W3C Software Notice and
+ * Document License (2015-05-13) which is available at
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+
+export function isThingDescription(input: unknown): input is WoT.ThingDescription {
+    // TODO: Improve validation function
```
danielpeintner, line 17: Does it make sense to also have proper validation based on the JSON schema (AJV) validation like we do for TMs.
https://github.com/eclipse-thingweb/node-wot/blob/0fd864b04c0badca2b8dbb90f01144614c36c37d/packages/td-tools/src/thing-model-helpers.ts#L154

Anyhow, definitely not in this PR and I am not even sure whether it makes sense int this case since we should be relaxed when consuming a TD and as strict as possible when creating a TD.

danielpeintner (approved):
I think this PR is a very good start and for me, it is okay to merge it as is 👍

packages/core/src/validation.ts
JKRhb: Thank you for your feedback! I actually had already started to change the validation approach so that it uses the JSON schema definition for validation and also added it to the PR now in the latest commits :) Hope that this is still fine

danielpeintner (approved):
Looks still good 👍

relu91 (commented):
I'm not sure if it is worth to approve the PR. Some remarks that I have:
- validation: I like the approach but it is a *half* refactoring, we didn't consistently check where ajv was used and updated the code accordingly. I think the refactoring is needed (and has serious implications with #1062) but it needs to be done systematically. 
-  `@ts-expect-error`: I'm not comfortable about disabling tsc without a proper explanation of what is causing the problem and **why** is a false positive.

Nonetheless, with this new function we are still in the exploratory space so if @danielpeintner wants to move on I won't block it! Just keep in mind the two issues (that need to be added in the issue tracker).

danielpeintner:
> Nonetheless, with this new function we are still in the exploratory space so if @danielpeintner wants to move on I won't block it! Just keep in mind the two issues (that need to be added in the issue tracker).

@JKRhb @relu91  what do you think. Shall we tackle the issues mentioned by @relu91 in this PR or is it better to merge this PR as is and create follow-up PRs....

JKRhb:
Thank you for your feedback, @relu91 and @danielpeintner! You raised valid concerns, @relu91, and I think we could provide some preliminary fixes in this PR:

> * validation: I like the approach but it is a _half_ refactoring, we didn't consistently check where ajv was used and updated the code accordingly. I think the refactoring is needed (and has serious implications with [Consume and read loop causes memory leak #1062](https://github.com/eclipse-thingweb/node-wot/issues/1062)) but it needs to be done systematically.

Yeah, I also had to think about the `ajv` issue, and fixing the underlying problem first definitely sounds like the better approach. As we are still in the experimental phase, I could just remove the body of the `isThingDescription` function or change it back to the very simple validation I had before (that is, only checking for the `@context`, the `title`, and the security definitions, for example). In a follow-up PR, we could then replace that validation logic with a more sophisticated one that might also utilize something like `ajv`.

> * `@ts-expect-error`: I'm not comfortable about disabling tsc without a proper explanation of what is causing the problem and **why** is a false positive.

Sure, I will investigate what causes the issue here :) It is a bit strange since it runs without a problem when ignoring the errors TypeScript is reporting here, but these might be related to the ES version we are targeting (note that I tried to use the "regular" streams instead of the Node.js ones). Should I try changing those to their Node.js counterpart?

JKRhb:
I think @relu91's feedback should now be addressed for the most part – since Node.js streams seem to have a different API, I simplified the `ThingDiscoveryProcess` implementation a lot, making the use of the `AsyncIterable` interface almost a bit pointless. However, we might be able to utilize its asynchronous nature a bit better once we add an implementation for the `discover` method.

Regarding the `ajv` issue: I now reverted the change and implemented two wrapper functions around the Helper methods we already had in `validation.ts` that could be expanded upon in the future. I hope this is an improvement, otherwise just let me know :) I also applied the newly introduced functions in `requestThingDescription` by the way to ensure a unified behavior.

JKRhb:
Added a tiny fix and two additional test cases in the latest commits.

relu91 (approved):
Thank you @JKRhb much better now!

relu91:
> I think @relu91's feedback should now be addressed for the most part – since Node.js streams seem to have a different API

Btw have you tried the implementation in the browser? is browserify able to translate the streams correctly?

danielpeintner:
I think we can merge this PR as is and look at some points raised in the comments later
* @relu91 check implementation in browser
* @danielpeintner function getLastValidationErrors() not in any case what one would expect ...

Did I miss anything?
I will create a follow-up issue..

PR #1188, opened 2023-12-10
fix(coap-server): ignore incoming requests with an invalid source port

### LINKED ISSUE #1182: CoapServer crashes with RangeError [ERR_SOCKET_BAD_PORT]
I recently noticed that the node-wot instance on our Eclipse vserver crashes once it deals with CoAP.

It seems to start fine on port 5683 but later it complaints about bad port.

```
  node-wot:binding-coap:coap-server:debug CoapServer on port 5683 received 'GET(3900) /' from 36.148.4.22:22 +58s
  node-wot:core:helpers:debug AddressHelper uses static plugfest.thingweb.io +58s
  node-wot:binding-coap:coap-server:debug CoapServer replied with '2.05' to 36.148.4.22:22 +1ms
  node-wot:binding-coap:coap-server:debug CoapServer on port 5683 received 'GET(3900) /' from 123.156.231.16:22 +886ms
  node-wot:core:helpers:debug AddressHelper uses static plugfest.thingweb.io +888ms
  node-wot:binding-coap:coap-server:debug CoapServer replied with '2.05' to 123.156.231.16:22 +1ms
  node-wot:binding-coap:coap-server:debug CoapServer on port 5683 received 'GET(3900) /' from 139.99.115.48:0 +2m
  node-wot:core:helpers:debug AddressHelper uses static plugfest.thingweb.io +2m
  node-wot:binding-coap:coap-server:debug CoapServer replied with '2.05' to 139.99.115.48:0 +2ms
  node-wot:cli:cli-default-servient:error Servient caught Asynchronous script error '/home/vserver/thingweb.node-wot/examples/testthing/testthing.js' and halted with RangeError [ERR_SOCKET_BAD_PORT]: Port should be > 0 and < 65536. Received type number (0).
  node-wot:cli:cli-default-servient:error     at new NodeError (node:internal/errors:405:5)
  node-wot:cli:cli-default-servient:error     at validatePort (node:internal/validators:409:11)
  node-wot:cli:cli-default-servient:error     at Socket.send (node:dgram:633:12)
```

@JKRhb maybe you have an idea what might be causing the problem....

JKRhb:
Resolves #1182.

This is more like a workaround for now to fix the crashes immediately but should be fixed at the library level eventually.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -435,7 +435,9 @@ export default class CoapServer implements ProtocolServer {
     }
 
     private async handleRequest(req: IncomingMessage, res: OutgoingMessage) {
-        if (req.rsinfo.port === 0) {
+        const sourcePort = req.rsinfo.port;
+        const hasInvalidPortRange =  sourcePort < 1 || sourcePort > 65535;
+        if (hasInvalidPortRange) {
             // Ignore requests with an invalid source port
             // See https://github.com/eclipse-thingweb/node-wot/issues/1182
             return;
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -436,7 +436,7 @@ export default class CoapServer implements ProtocolServer {
 
     private async handleRequest(req: IncomingMessage, res: OutgoingMessage) {
         const sourcePort = req.rsinfo.port;
-        const hasInvalidPortRange =  sourcePort < 1 || sourcePort > 65535;
+        const hasInvalidPortRange = sourcePort < 1 || sourcePort > 65535;
         if (hasInvalidPortRange) {
             // Ignore requests with an invalid source port
             // See https://github.com/eclipse-thingweb/node-wot/issues/1182
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -435,6 +435,12 @@ export default class CoapServer implements ProtocolServer {
     }
 
     private async handleRequest(req: IncomingMessage, res: OutgoingMessage) {
+        if (req.rsinfo.port === 0) {
+            // Ignore requests with an invalid source port
+            // See https://github.com/eclipse-thingweb/node-wot/issues/1182
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


relu91:
What happens to the ignored request? Do clients try to send another request? Also I'd add a followup issue to revert the change once it is solved in node-coap.

JKRhb:
> What happens to the ignored request? Do clients try to send another request?

If the request is confirmable, then clients will try to retransmit it until it has timed out. Usually, you could inform them that their request was unable to be processed by sending a reset message – but in this case, that is not possible as we would be running into the same problem again :/

> Also I'd add a followup issue to revert the change once it is solved in node-coap.

That sounds good!

danielpeintner:
> > Also I'd add a followup issue to revert the change once it is solved in node-coap.
> 
> That sounds good!

Created https://github.com/eclipse-thingweb/node-wot/issues/1190

packages/binding-coap/src/coap-server.ts
```diff
@@ -435,6 +435,12 @@ export default class CoapServer implements ProtocolServer {
     }
 
     private async handleRequest(req: IncomingMessage, res: OutgoingMessage) {
+        if (req.rsinfo.port === 0) {
```
danielpeintner, line 438: since `port` is of type `number` ... does it make sense to check for negative values also?
```suggestion
        if (req.rsinfo.port <= 0) {
```

danielpeintner (approved):
LGTM, just a suggestion see below...

Since it is a temporary fix I am okay either way

packages/binding-coap/src/coap-server.ts
JKRhb: Good point! Maybe we could even have a check for the full port range, to also reflect the original error message.

```suggestion
        const sourcePort = req.rsinfo.port;
        const hasInvalidPortRange =  sourcePort < 1 || sourcePort > 65535;
        if (hasInvalidPortRange) {
```

JKRhb: Took the freedom of already committing this in a640d6cfb8247f386d89d86e6a5f0fe8fb13625b, hope the change is fine for you as well 😅

Issue #1193, opened 2023-12-15
[CoAP] Exposing multiple Things on the same Servient produces an Error


FadySalama:
When I try to add multiple CoAP devices to the same servient defined like this 
```typescript
const coapServient = new Servient()
coapServient.addServer(new CoapServer({
    port: 5683
}))
```
it causes this error to occur 
```bash
Error: getsockname EBADF
    at Socket.address (node:dgram:775:11)
    at CoapServer.getPort (/node_modules/@node-wot/binding-coap/dist/coap-server.js:104:27)
    at CoapServer.<anonymous> (/node_modules/@node-wot/binding-coap/dist/coap-server.js:112:31)
    at Generator.next (<anonymous>)
    at /node_modules/@node-wot/binding-coap/dist/coap-server.js:31:71
    at new Promise (<anonymous>)
    at __awaiter (/node_modules/@node-wot/binding-coap/dist/coap-server.js:27:12)
    at CoapServer.expose (/node_modules/@node-wot/binding-coap/dist/coap-server.js:111:16)
    at /node_modules/@node-wot/core/dist/servient.js:51:40
    at Array.forEach (<anonymous>) {
  errno: -9,
  code: 'EBADF',
  syscall: 'getsockname'
}
```
The error does not happen if I expose only 1 CoAP device and only happens with CoAP. (I run the same code to generate HTTP instances as well and it works flawlessly). I was able to replicate the error in a Windows environment and in a dockerized Ubuntu environment running an Ubuntu host.

FadySalama:
A small update: I found what was causing this issue. I was calling `servient.start()` for each new device added to get the WoT object which did break things severely for CoAP and not so severely for HTTP. By calling servient.start() only once, the issues were solved and I could add multiple CoAP devices without any issues.
The problem is, there is nothing stopping me from calling `servient.start()` multiple times. How about ensuring that calling `servient.start()` always returns exactly one `WoT` object using a singelton pattern? Otherwise, I think it is safe to close this issue.

danielpeintner:
I opened https://github.com/eclipse-thingweb/node-wot/pull/1195 which prevents a Servient to be started a second time.

Hence in your case you would get an error ("Servient started already") the second time. 
Does this make sense?

PR #1195, opened 2023-12-19
refactor: avoid starting servient multiple times

danielpeintner:
I think it should **not** be possible to start a servient multiple times. Hence I added a simple check.

Moreover, I _think_ a servient should be usable only once. Hence after a `shuthown` it is no longer usable.
Anyhow, I am open to other opinions whether we should allow for a re-`start` again?

fixes https://github.com/eclipse-thingweb/node-wot/issues/1193

### DIFF
--- a/packages/core/src/servient.ts
+++ b/packages/core/src/servient.ts
@@ -30,7 +30,7 @@ export default class Servient {
     private things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
     private credentialStore: Map<string, Array<unknown>> = new Map<string, Array<unknown>>();
 
-    #started = false;
+    #startedWoT?: typeof WoT;
     #shutdown = false;
 
     /** add a new codec to support a mediatype; offered mediatypes are listed in TDs */
@@ -213,25 +213,29 @@ export default class Servient {
 
     // will return WoT object
     public async start(): Promise<typeof WoT> {
-        if (this.#started) {
-            throw Error("Servient started already");
+        if (this.#startedWoT != null) {
+            debug("Servient started already -> nop -> returning previous WoT implementation");
+            return this.#startedWoT;
         }
+        if (this.#shutdown) {
+            throw Error("Servient cannot be started (again) since it was already stopped");
+        }
+
         const serverStatus: Array<Promise<void>> = [];
         this.servers.forEach((server) => serverStatus.push(server.start(this)));
         this.clientFactories.forEach((clientFactory) => clientFactory.init());
 
         await Promise.all(serverStatus);
-        this.#started = true;
-        return new WoTImpl(this);
+        return (this.#startedWoT = new WoTImpl(this));
     }
 
     public async shutdown(): Promise<void> {
-        if (!this.#started) {
-            debug("Servient cannot be shutdown, wasn't even started");
-            return;
+        if (this.#startedWoT === undefined) {
+            throw Error("Servient cannot be shutdown, wasn't even started");
         }
         if (this.#shutdown) {
-            throw Error("Servient shutdown already");
+            debug("Servient shutdown already -> nop");
+            return;
         }
 
         this.clientFactories.forEach((clientFactory) => clientFactory.destroy());
@@ -239,5 +243,6 @@ export default class Servient {
         const promises = this.servers.map((server) => server.stop());
         await Promise.all(promises);
         this.#shutdown = true;
+        this.#startedWoT = undefined; // clean-up reference
     }
 }
--- a/packages/core/src/servient.ts
+++ b/packages/core/src/servient.ts
@@ -30,6 +30,9 @@ export default class Servient {
     private things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
     private credentialStore: Map<string, Array<unknown>> = new Map<string, Array<unknown>>();
 
+    private started = false;
+    private shutdowned = false;
+
     /** add a new codec to support a mediatype; offered mediatypes are listed in TDs */
     public addMediaType(codec: ContentCodec, offered = false): void {
         ContentManager.addCodec(codec, offered);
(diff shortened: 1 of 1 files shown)


packages/core/src/servient.ts
```diff
@@ -30,6 +30,9 @@ export default class Servient {
     private things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
     private credentialStore: Map<string, Array<unknown>> = new Map<string, Array<unknown>>();
 
+    private started = false;
+    private shutdowned = false;
```
JKRhb, line 34: Hmm, maybe we could also use `closed` as a variable name here?

packages/core/src/servient.ts
```diff
@@ -210,18 +213,30 @@ export default class Servient {
...
+        this.started = true;
         return new WoTImpl(this);
     }
 
     public async shutdown(): Promise<void> {
-        this.clientFactories.forEach((clientFactory) => clientFactory.destroy());
+        if (this.started) {
+            if (this.shutdowned) {
+                throw Error("Servient shutdowned already");
+            }
```
JKRhb, line 232: I think by reversing the logic here, we could get rid of one level of nesting and make the code a bit more readable.

```suggestion
        if (!this.started) {
            debug(`Servient cannot be shut down, wasn't even started`);
            return;
        }
        
        if (this.shutdowned) {
            throw Error("Servient shutdowned already");
        }
```

JKRhb (commented):
Thank you, @danielpeintner, the changes already look good to me :) See two comments/suggestions below.

egekorkan:
> Anyhow, I am open to other opinions whether we should allow for a re-start again?

I do not see a use case but I do not see issues for allowing that

danielpeintner:
> > Anyhow, I am open to other opinions whether we should allow for a re-start again?
> 
> I do not see a use case but I do not see issues for allowing that

The issues I see for allowing to restart is that I fear internal states "hanging around" that cause problems in the end 🙃

packages/core/src/servient.ts
danielpeintner: Sometimes I think people have different _taste_ of code. I am mostly fine with nesting while you dislike it ;-)
Anyhow, in this case I am fine either way. Fixed with https://github.com/eclipse-thingweb/node-wot/pull/1195/commits/b2b4388aeb71135e178769faa7a50008ad08f664

packages/core/src/servient.ts
danielpeintner: I don't like "shutdowned" either and had "destroyed" at first but what convinced me was to align it with the function names we use...

JKRhb: Hmm, but I guess the correct adjective form for "to shut down" would be "shut down", right? How about something like `isShutDown`? Since its private, we could probably also use `#shutdown` or `#shutDown` for this field without causing a conflict.

danielpeintner: like this https://github.com/eclipse-thingweb/node-wot/pull/1195/commits/08e85536a462be9c2fed6abd01304ac9f0827dc7

JKRhb (approved):

relu91 (changes requested):
I'm not sure I like the approach (should I throw or should I log and silently return). What about:
- In the start method: 
   1. throw if the servient was shut down
   2. silently return and `debug` message the servient was already started!
- In the shutdown method:
   1. throw if the servient was not even started (it's an invalid state after all)
   2. silently return and `debug` message the servient was already shut down

danielpeintner:
Thank you. I wasn't sure either.

> In the start method:
> 
>     1. throw if the servient was shut down
> 
>     2. silently return and `debug` message the servient was already started if it `#started` is true

I felt it is best to throw if it has been started already since I guess this is an actual error and as a developer I want to know it.

> In the shutdown method:
> 
>     1. throw if the servient was not even started (it's an invalid state after all)
> 
>     2. silently return and `debug` message the servient was already shut down  `#shudown` is true

Good point...  the more I think about I would do the opposite:
* if not started -> throw Error
* if shutdown already -> just debug... does not really harm

Do you have an other opinion? Throw in any case?

> Btw I'm okey with using `#shudown` and `#stated` but should we add the `private` modifier or is it still redundant (given they are really private fields for js)?

It is redundant and also shows as such in VS Code.

danielpeintner:
I am mostly okay with your reasoning.

There is one aspect that is a bit undesirable 

> * if the servient was started and start is called -> you are already in the required state -> nop

The `start `method requires us to return a `WoT` object.
Hence if we want to do what you are suggesting we have to keep a handle.. see the changes in https://github.com/eclipse-thingweb/node-wot/pull/1195/commits/0d0c1f0776432aee6548662f9ff672399e90c523

I am fine with doing so.

danielpeintner:
Note: with the lastest changes we have issues with CoAP tests (see [should reject requests for undefined meta operations](https://github.com/eclipse-thingweb/node-wot/actions/runs/7448073693/job/20261632064?pr=1195#step:9:228)).
Maybe it is good to get feedback from @JKRhb again

packages/core/src/servient.ts
```diff
@@ -210,18 +213,36 @@ export default class Servient {
...
 
         await Promise.all(serverStatus);
-        return new WoTImpl(this);
+        return (this.#startedWoT = new WoTImpl(this));
     }
 
     public async shutdown(): Promise<void> {
+        if (this.#startedWoT === undefined) {
+            throw Error("Servient cannot be shutdown, wasn't even started");
+        }
```
JKRhb, line 235: Seems like the CoAP tests are failing because they are calling the Servient's `shutdown` method without invoking `start` first. Should we maybe just return here? That would solve the problem. Or is it better to throw an error here?

packages/core/src/servient.ts
```diff
@@ -30,6 +30,9 @@ export default class Servient {
     private things: Map<string, ExposedThing> = new Map<string, ExposedThing>();
     private credentialStore: Map<string, Array<unknown>> = new Map<string, Array<unknown>>();
 
+    #startedWoT?: typeof WoT;
```
JKRhb, line 33: Maybe a field name like `wotInstance` could be a bit clearer here?

packages/core/src/servient.ts
danielpeintner: @relu91 ?
According to https://github.com/eclipse-thingweb/node-wot/pull/1195#issuecomment-1880833020 Cristiano prefers to throw ...

packages/core/src/servient.ts
danielpeintner: Sure 👍
I will wait till we are fine with the proposed changes and apply the change...

packages/core/src/servient.ts
relu91: Why are they calling stop without starting it? For the reasons above I would not jus silently return here.

packages/core/src/servient.ts
danielpeintner: fixed with https://github.com/eclipse-thingweb/node-wot/pull/1195/commits/aa1537187cad8789907972262b7ef15c2f55fc0d

packages/core/src/servient.ts
danielpeintner: resolved by https://github.com/eclipse-thingweb/node-wot/pull/1212

danielpeintner:
I _think_ we should now have a solution that works for all of us ;-)

relu91 (approved):
Good to go 👍🏻

PR #1204, opened 2023-12-21
ci: enable logging only when debug logging is enabled

JKRhb:
In the menu for rerunning failed actions, I noticed that there is an option for "enabling debug logging" as can be seen in this screenshot:

image

Although we have introduced automatic reruns with debug logging in #1189, I got the impression that this change is not that well usable in practice, partly because it makes the logs for failed actions very long, thereby increasing the difficulty of finding the place where something went wrong.

This PR proposes using debug logging only when it is enabled via the GitHub actions UI that can be seen above, making the choice of seeing the verbose output a bit more deliberate. This change would also half the wait time for feedback on a failed test while getting rid of the external dependency introduced for retrying the tests automatically.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,17 +47,14 @@ jobs:
               run: |
                   sudo echo "127.0.0.1" `hostname` | sudo tee -a /etc/hosts
 
+            - name: Configure debug logging
+              if: runner.debug == '1'
+              run: echo "DEBUG=node-wot*" >> $GITHUB_ENV
+              shell: bash
+
             - name: Test with coverage report
-              uses: nick-fields/retry@v2
-              with:
-                  timeout_minutes: 10
-                  max_attempts: 2
-                  command: npm run coverage:only
-                  # Rerun with debug logging enabled on error, but always fail the retry
-                  new_command_on_retry: |
-                      DEBUG=node-wot* npm run test:only
-                      exit 1
-                  shell: bash
+              timeout-minutes: 10
+              run: npm run coverage:only
 
             - name: Upload to codecov.io
               uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


JKRhb:
A demonstration of a run with debug logging enabled can be found here: https://github.com/eclipse-thingweb/node-wot/actions/runs/7287753854/job/19859204326

.github/workflows/ci.yaml
```diff
@@ -47,17 +47,14 @@ jobs:
               run: |
                   sudo echo "127.0.0.1" `hostname` | sudo tee -a /etc/hosts
 
+            - name: Configure debug logging
+              if: runner.debug == '1'
```
danielpeintner, line 51: I think it makes sense to describe what this flag is and how it can be enabled...

Not sure if some prose in ci.yaml makes sense or also pointing to this PR where one could see the GH screenshot etc

danielpeintner (changes requested):
I think the change/simplification makes sense.

Below a "request" ;-)

relu91 (approved):
Great! I discovered the option while we were debuggin MQTT and I didn't think we could exploit like this. Awesome! It is going still hard to catch that nasty MQTT problem but nonetheless I'd go on and merge the PR after @danielpeintner's concerns are resolved.

.github/workflows/ci.yaml
JKRhb, line 51: @danielpeintner Would something like this work for you? :)

```suggestion
            # Enables the output of log messages from the Node.js debug module
            # for workflow runs that are being retried with "debug logging"
            # enabled.
            #
            # See https://github.com/eclipse-thingweb/node-wot/pull/1204 for
            # more context regarding the rationale of this step.
            - name: Configure debug logging
              if: runner.debug == '1'
```

danielpeintner: Looks good!

danielpeintner (approved):

PR #1205, opened 2023-12-21
chore(binding-mqtt): remove unused debug import

JKRhb:
This is an extremely trivial PR that gets rid of the single warning eslint is currently reporting on the `master` branch.

Since the change is so simple, I will merge it right away.

### DIFF
--- a/packages/binding-mqtt/src/util.ts
+++ b/packages/binding-mqtt/src/util.ts
@@ -16,7 +16,7 @@ import { createLoggers } from "@node-wot/core";
 import { MqttQoS } from "./mqtt";
 import { IClientPublishOptions } from "mqtt";
 
-const { debug, warn } = createLoggers("binding-mqtt", "mqtt-util");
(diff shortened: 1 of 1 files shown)


packages/binding-mqtt/src/util.ts
```diff
@@ -16,7 +16,7 @@ import { createLoggers } from "@node-wot/core";
 import { MqttQoS } from "./mqtt";
 import { IClientPublishOptions } from "mqtt";
 
-const { debug, warn } = createLoggers("binding-mqtt", "mqtt-util");
+const { warn } = createLoggers("binding-mqtt", "mqtt-util");
```
danielpeintner, line 19: Mhh interesting. I would have said that the code checks we have in place mention unused imports...

JKRhb: The warning popped up in other PRs as an additional warning not related to code being changed, which is how I noticed it – but yeah, strange that it slipped through 🤔

Issue #1194, opened 2023-12-15
Follow-up octet-stream codec - short explanation


danielpeintner:
> Thank you very, much for your efforts! We can merge it and try to play with it. As a follow-up work, we should provide a short explanation about the new terms introduced.

_Originally posted by @relu91 in https://github.com/eclipse-thingweb/node-wot/pull/1125#pullrequestreview-1783480216_

derwehr:
I can write up explanations for the added terms along with examples. Where would be the correct place to add the descriptions?

danielpeintner:
> I can write up explanations for the added terms along with examples. Where would be the correct place to add the descriptions?

We have READMEs for each package. Hence I suggest to enhance the [core README](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/README.md) with a subsection of "Codecs" in which you can explain the octet-stream codec.

@relu91 @JKRhb I hope this is fine by you as well... otherwise we can move around the description

PR #1209, opened 2024-01-02
Follow-up octet-stream codec - short explanation #1194

derwehr:
I've added descriptions for the content type parameters and data schema terms of the `octet-stream` codec, and an example of how to describe an object with the new terms.

(related issue: https://github.com/eclipse-thingweb/node-wot/issues/1194)

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -24,6 +24,7 @@ see binding examples such as
 ## Codecs
 
 ### Octet-Stream
+
 The octet-stream codec enables deserialization and serialization of binary data. To encode and decode sequences of bytes, the octet-stream codec uses the following content type parameters:
 | Parameter | Description | Default |
 | --- | --- | --- |
@@ -40,15 +41,17 @@ Additionally, the octet-stream codec supports the data schema terms below for ad
 With the help of the terms and parameters above, the octet-stream codec can be used to serialize and deserialize objects containing bit-fields and sequences of bytes, like in the following example.
 
 #### Example
+
 To serialize the object `{ flag1: true, numberProperty: 99, stringProperty: "Web" }` to a sequence of bytes, the content type `application/octet-stream;length=4;signed=false;`, along with the following data schema can be used:
+
 ```json
 {
     "type": "object",
     "properties": {
         "flag1": { "type": "boolean", "ex:bitOffset": 0, "ex:bitLength": 1 },
         "numberProperty": { "type": "integer", "ex:bitOffset": 1, "ex:bitLength": 7 },
-        "stringProperty": { "type": "string", "ex:bitOffset": 2, "ex:bitLength": 24 },
-    },
+        "stringProperty": { "type": "string", "ex:bitOffset": 2, "ex:bitLength": 24 }
+    }
 }
 ```
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -21,6 +21,37 @@ see binding examples such as
 -   https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-http
 -   https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-coap
 
+## Codecs
+
+### Octet-Stream
+The octet-stream codec enables deserialization and serialization of binary data. To encode and decode sequences of bytes, the octet-stream codec uses the following content type parameters:
+| Parameter | Description | Default |
+| --- | --- | --- |
+| `length` | Number of bytes produced during serialization or consumed during deserialization | |
+| `signed` | Signedness of the data type, `true` or `false` | `true` |
+| `byteSeq` | Endianness, enum of `BIG_ENDIAN`, `LITTLE_ENDIAN`, `BIG_ENDIAN_BYTE_SWAP`, `LITTLE_ENDIAN_BYTE_SWAP` | `BIG_ENDIAN` |
+
+Additionally, the octet-stream codec supports the data schema terms below for addressing bit-fields:
+| Term | Description | Default |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


packages/core/README.md
```diff
@@ -21,6 +21,37 @@ see binding examples such as
...
+#### Example
+To serialize the object `{ flag1: true, numberProperty: 99, stringProperty: "Web" }` to a sequence of bytes, the content type `application/octet-stream;length=4;signed=false;`, along with the following data schema can be used:
+```json
+{
+    "type": "object",
+    "properties": {
+        "flag1": { "type": "boolean", "ex:bitOffset": 0, "ex:bitLength": 1 },
+        "numberProperty": { "type": "integer", "ex:bitOffset": 1, "ex:bitLength": 7 },
+        "stringProperty": { "type": "string", "ex:bitOffset": 2, "ex:bitLength": 24 },
+    },
```
JKRhb, line 51:
```suggestion
        "stringProperty": { "type": "string", "ex:bitOffset": 2, "ex:bitLength": 24 }
    }
```

packages/core/README.md
```diff
@@ -21,6 +21,37 @@ see binding examples such as
 -   https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-http
 -   https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-coap
 
+## Codecs
+
+### Octet-Stream
+The octet-stream codec enables deserialization and serialization of binary data. To encode and decode sequences of bytes, the octet-stream codec uses the following content type parameters:
```
JKRhb, line 27:
```suggestion

The octet-stream codec enables deserialization and serialization of binary data. To encode and decode sequences of bytes, the octet-stream codec uses the following content type parameters:
```

packages/core/README.md
```diff
@@ -21,6 +21,37 @@ see binding examples such as
...
+Additionally, the octet-stream codec supports the data schema terms below for addressing bit-fields:
+| Term | Description | Default |
+| --- | --- | --- |
+| `ex:bitLength` | Number of bits produced during serialization or consumed during deserialization | |
+| `ex:bitOffset` | Offset in bits from the beginning of the byte sequence | `0` |
+
+With the help of the terms and parameters above, the octet-stream codec can be used to serialize and deserialize objects containing bit-fields and sequences of bytes, like in the following example.
+
+#### Example
+To serialize the object `{ flag1: true, numberProperty: 99, stringProperty: "Web" }` to a sequence of bytes, the content type `application/octet-stream;length=4;signed=false;`, along with the following data schema can be used:
```
JKRhb, line 43:
```suggestion

To serialize the object `{ flag1: true, numberProperty: 99, stringProperty: "Web" }` to a sequence of bytes, the content type `application/octet-stream;length=4;signed=false;`, along with the following data schema can be used:
```

packages/core/README.md
```diff
@@ -21,6 +21,37 @@ see binding examples such as
...
+```json
```
JKRhb, line 44:
```suggestion

```json
```

JKRhb (approved):
The following suggestions can be applied to get rid of the prettier errors – otherwise LGTM :)

derwehr:
> The following suggestions can be applied to get rid of the prettier errors – otherwise LGTM :)

Thanks for the review. I commited the suggested changes

packages/core/README.md
```diff
@@ -21,6 +21,40 @@ see binding examples such as
...
+
+To serialize the object `{ flag1: true, numberProperty: 99, stringProperty: "Web" }` to a sequence of bytes, the content type `application/octet-stream;length=4;signed=false;`, along with the following data schema can be used:
+
+```json
+{
+    "type": "object",
+    "properties": {
+        "flag1": { "type": "boolean", "ex:bitOffset": 0, "ex:bitLength": 1 },
+        "numberProperty": { "type": "integer", "ex:bitOffset": 1, "ex:bitLength": 7 },
+        "stringProperty": { "type": "string", "ex:bitOffset": 2, "ex:bitLength": 24 }
```
danielpeintner, line 53:
```suggestion
        "stringProperty": { "type": "string", "ex:bitOffset": 8, "ex:bitLength": 24 }
```

I am not very much in the proposal but shouldn't it be 8 instead of 2 .. since 8 bits are used for flag1 and numberProperty.

In general I wonder how one can deal with the order of terms ... since in general a JSON object per se is unordered and hence you need the the right offset, don't you?

derwehr: Good find, thank you!

I'm unsure if this is what you're asking, but the `ex:bitOffset` term describes the serialized octet-stream, so the deserialized JSON object does not need to be ordered.

When deserializing the octet-stream `[0x63, 0x57, 0x65, 0x62]` with the schema defined in the readme, one might as well change the order of the object's properties to

```json
{
    "type": "object",
    "properties": {
        "stringProperty": { "type": "string", "ex:bitOffset": 8, "ex:bitLength": 24 }
        "flag1": { "type": "boolean", "ex:bitOffset": 0, "ex:bitLength": 1 },
        "numberProperty": { "type": "integer", "ex:bitOffset": 1, "ex:bitLength": 7 },
    }
}
```

Also `{ flag1: true, numberProperty: 99, stringProperty: "Web" }`  and `{ numberProperty: 99, flag1: true,  stringProperty: "Web" }`  both get serializied to `[0x63, 0x57, 0x65, 0x62]` with both of the schemas above.

Therefore, the Data Schema specifies the order of the bits in the octet-stream independently from the order in the JSON object.

relu91 (approved):
Thank you! good to go from my side.

danielpeintner (approved):

Issue #1177, opened 2023-11-29
Enable `@typescript-eslint/no-unnecessary-condition` rule


JKRhb:
In the context of #1176, I noticed that we currently perform a lot of checks and nullish coalescing that are actually not really necessary based on our current typings. In some cases, this makes the code less readable, in other cases this can pave the way for potential bugs since some possible null values are hidden behind typecasts.

`eslint` provides the rule `@typescript-eslint/no-unnecessary-condition` which makes it possible to detect and fix these problems. As another step toward more type safety, this rule could be added on a per-package basis again, which should ultimately make the library more reliable.

lchauha:
Hi @JKRhb I would like to work on this issue. Can you please assign this to me?

PR #1210, opened 2024-01-06
chore: add `guard-for-in` eslint rule

JKRhb:
(Not only) in the context of #1177, I noticed that some of the for loops use the `for ... in` syntax instead of the `for ... of` syntax. Apparently, the former style is discouraged and not recommended by eslint, for example (see https://eslint.org/docs/latest/rules/guard-for-in).

This PR replaces the code lines in question and enforces the new style by activating the `guard-for-in` eslint rule.

### DIFF
--- a/packages/binding-coap/src/mdns-introducer.ts
+++ b/packages/binding-coap/src/mdns-introducer.ts
@@ -73,8 +73,8 @@ export class MdnsIntroducer {
     private determineTarget(): string {
         const interfaces = networkInterfaces();
 
-        for (const iface in interfaces) {
-            for (const entry of interfaces[iface] ?? []) {
+        for (const iface of Object.values(interfaces ?? {})) {
+            for (const entry of iface ?? []) {
                 if (entry.internal === false) {
                     if (entry.family === this.ipAddressFamily) {
                         return entry.address;
--- a/packages/binding-netconf/src/codecs/netconf-codec.ts
+++ b/packages/binding-netconf/src/codecs/netconf-codec.ts
@@ -126,8 +126,9 @@ export default class NetconfCodec {
             let nsFound = false;
             let aliasNs = "";
             let value;
-            for (const key in properties) {
-                const el = properties[key];
+            // TODO: Use correct type for el
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            for (const [key, el] of Object.entries(properties) as [string, any]) {
                 const payloadField = payload[key];
                 if (payloadField == null) {
                     throw new Error(`Payload is missing '${key}' field specified in TD`);
--- a/packages/core/src/codecs/netconf-codec.ts
+++ b/packages/core/src/codecs/netconf-codec.ts
@@ -79,8 +79,9 @@ export default class NetconfCodec implements ContentCodec {
             let nsFound = false;
             let aliasNs = "";
             let value;
-            for (const key in properties) {
-                const el = properties[key];
+            // TODO: Use correct type for el
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            for (const [key, el] of Object.entries(properties) as [string, any]) {
                 if (payload[key] == null) {
                     throw new Error(`Payload is missing '${key}' field specified in TD`);
                 }
--- a/packages/core/src/codecs/octetstream-codec.ts
+++ b/packages/core/src/codecs/octetstream-codec.ts
@@ -604,11 +604,12 @@ export default class OctetstreamCodec implements ContentCodec {
         }
 
         result = result ?? Buffer.alloc(length);
-        for (const propertyName in schema.properties) {
+        // TODO: Use correct type for propertySchema
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        for (const [propertyName, propertySchema] of Object.entries(schema.properties) as [string, any]) {
             if (Object.hasOwnProperty.call(value, propertyName) === false) {
                 throw new Error(`Missing property '${propertyName}'`);
             }
-            const propertySchema = schema.properties[propertyName];
             const propertyValue = value[propertyName];
             const propertyOffset = parseInt(propertySchema["ex:bitOffset"]);
             const propertyLength = parseInt(propertySchema["ex:bitLength"]);
--- a/packages/td-tools/src/td-parser.ts
+++ b/packages/td-tools/src/td-parser.ts
@@ -19,13 +19,41 @@ import * as TDHelpers from "./td-helpers";
 
 import isAbsoluteUrl = require("is-absolute-url");
 import URLToolkit = require("url-toolkit");
-import { ThingContext } from "wot-thing-description-types";
+import { ThingContext, PropertyElement, ActionElement, EventElement } from "wot-thing-description-types";
 
 // TODO: Refactor and reuse debug solution from core package
 import debug from "debug";
 const namespace = "node-wot:td-tools:td-parser";
 const logDebug = debug(`${namespace}:debug`);
 const logWarn = debug(`${namespace}:warn`);
+
+type AffordanceElement = PropertyElement | ActionElement | EventElement;
+
+/**
+ * Initializes the affordances field of a thing with an empty object if its
+ * type should be incorrect or undefined.
+ *
+ * This avoids potential errors that could occur due to an undefined
+ * affordance field.
+ *
+ * @param thing The Thing whose affordance field is being adjusted.
+ * @param affordanceKey The key of the affordance field.
+ */
+function adjustAffordanceField(thing: Thing, affordanceKey: string) {
+    const affordance = thing[affordanceKey];
+
+    if (typeof affordance !== "object" || affordance == null) {
+        thing[affordanceKey] = {};
+    }
+}
+
+function adjustBooleanField(affordance: AffordanceElement, key: string) {
+    const currentValue = affordance[key];
+
+    if (currentValue === undefined || typeof currentValue !== "boolean") {
+        affordance[key] = false;
+    }
+}
 
 /** Parses a TD into a Thing object */
 export function parseTD(td: string, normalize?: boolean): Thing {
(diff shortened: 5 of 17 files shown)


danielpeintner:
Out of curiosity: Did you run into an actual problem or are you proposing the change because of the possible unexpected errors?

JKRhb:
> Out of curiosity: Did you run into an actual problem or are you proposing the change because of the possible unexpected errors?

I think there haven't been real problems so far, however, I think I've already read some time ago that using `for of` is a better approach than `for in` in general. So in that regard it is more like an optimization.

However, I've also noticed that there are several places where a pattern like

```js
  for (const key in object) {
    const value = object[key];
    
    useKeyAndValue(key, value); 
  }
```

is being used. In terms of readability and especially type safety, I think it is better to rather use something like


```js
  for (const [key, value] of Object.entries(object)) {
    useKeyAndValue(key, value); 
  }
```

instead, which is encouraged by activating the `guard-for-in` rule.

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -797,16 +790,20 @@ export class AssetInterfaceDescriptionUtil {
...
                         vi.secNamesForEndpoint = secNamesForEndpointMetadata.get(vi.endpointMetadata);
                     }
                     const form = this.createInteractionForm(vi, smInformation.endpointMetadataArray.length > 1);
-                    thing.properties[key].forms.push(form);
+                    forms.push(form);
                 }
+
+                thing.actions[key] = {
+                    // @ts-expect-error Here is a strange type mismatch present
+                    forms,
```
relu91, line 805: same comment as above.

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -818,16 +815,17 @@ export class AssetInterfaceDescriptionUtil {
             for (const [key, value] of smInformation.events.entries()) {
                 logInfo("Event " + key + " = " + value);
 
-                thing.events[key] = {};
-                thing.events[key].forms = [];
-
-                for (const vi of value) {
+                const forms = value.map((vi) => {
```
relu91, line 818: Why are we changing the paradigm here? didn't we use `for of` for the other affordances above?

relu91 (changes requested):
Some questions/required changes below.

packages/binding-coap/src/mdns-introducer.ts
```diff
@@ -73,8 +73,8 @@ export class MdnsIntroducer {
     private determineTarget(): string {
         const interfaces = networkInterfaces();
 
-        for (const iface in interfaces) {
-            for (const entry of interfaces[iface] ?? []) {
+        for (const iface of Object.values(interfaces ?? [])) {
```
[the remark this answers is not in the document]
JKRhb: Thank you, good catch :)

packages/td-tools/src/util/asset-interface-description.ts
```diff
@@ -682,8 +673,10 @@ export class AssetInterfaceDescriptionUtil {
             for (const [key, value] of smInformation.properties.entries()) {
                 logInfo("Property" + key + " = " + value);
 
-                thing.properties[key] = {};
-                thing.properties[key].forms = [];
+                thing.properties[key] = {
+                    // @ts-expect-error Here is a strange type mismatch present
+                    forms: [],
```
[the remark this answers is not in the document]
JKRhb: Hmm, yeah, that is true. I have now found a potential solution for achieving a type-safe implementation here. Should there be an error thrown if the `forms` array is empty?

packages/core/src/codecs/octetstream-codec.ts
```diff
@@ -535,11 +535,11 @@ export default class OctetstreamCodec implements ContentCodec {
         }
 
         result = result ?? Buffer.alloc(parseInt(parameters.length));
-        for (const propertyName in schema.properties) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        for (const [propertyName, propertySchema] of Object.entries(schema.properties) as [string, any]) {
```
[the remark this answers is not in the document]
JKRhb: > Missing TODO.

Thank you, should be fixed now :)

> can't we try to use the inferred type from the removed line below?
> 
> ```js
> const propertySchema = schema.properties[propertyName];
> ```

Unfortunately, the type of `propertySchema` was `any` here :/ `DataSchema` seems to be defined like this at the moment in the `wot-typescript-definitions`:

```
type DataSchema = { [key: string]: any; };
```

egekorkan (approved):
Fine regarding the changes to mqtt package

packages/td-tools/src/util/asset-interface-description.ts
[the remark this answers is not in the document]
JKRhb: See 4f2663fcc6096dbb61a96956b63ef479c5d7d99c for an intial proposal.

packages/td-tools/src/td-parser.ts
```diff
@@ -98,43 +98,21 @@ export function parseTD(td: string, normalize?: boolean): Thing {
...
-                prop.writeOnly = false;
-            }
-            if (prop.observable === undefined || typeof prop.observable !== "boolean") {
-                prop.observable = false;
-            }
-        }
+    for (const property of Object.values(thing.properties ?? {})) {
+        property.readOnly ??= false;
+        property.writeOnly ??= false;
+        property.observable ??= false;
```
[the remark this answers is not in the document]
danielpeintner: Good question. In theory for well-defined TDs it should not make any difference on the other hand the *more* complex "old" code fixes wrong types also. If it were me to decide I would stick with the old code.. but I do not have a very strong preference.


the same goes for actions and events

packages/binding-netconf/src/codecs/netconf-codec.ts
```diff
@@ -126,8 +126,9 @@ export default class NetconfCodec {
             let nsFound = false;
             let aliasNs = "";
             let value;
-            for (const key in properties) {
-                const el = properties[key];
+            // TODO: Use correct type for el
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            for (const [key, el] of Object.entries(properties) as [string, any]) {
```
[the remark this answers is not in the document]
JKRhb: Unfortunately, `DataSchema` is currently defined as an object whose values are typed as `any` :/ So in this case, we would need to cast. Note also that these are the `properties` of the `DataSchema` and not Property Affordances. In any case, the typing here needs to be revisited.

packages/td-tools/src/util/asset-interface-description.ts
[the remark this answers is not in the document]
JKRhb: With the `asset-interface-description.ts` file gone in the meantime, I will mark the respective threads as resolved :)

packages/td-tools/src/td-parser.ts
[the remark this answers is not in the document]
JKRhb: The code should now behave the same as before and should actually only be a refactoring :)

JKRhb:
Rebased this PR against the latest state of the `master` branch and applied some adjustments :) I think this should be ready for another review.

packages/td-tools/src/td-parser.ts
```diff
@@ -237,32 +240,20 @@ export function serializeTD(thing: Thing): string {
...
-            if (prop.writeOnly === undefined || typeof prop.writeOnly !== "boolean") {
-                prop.writeOnly = false;
-            }
-            if (prop.observable === undefined || typeof prop.observable !== "boolean") {
-                prop.observable = false;
-            }
+        for (const prop of Object.values(copy.properties)) {
+            prop.readOnly ??= false;
+            prop.writeOnly ??= false;
+            prop.observable ??= false;
```
danielpeintner, line 246: I am not 100% sure but I think the *new* code is not exactly doing the same. The old code fixed wrong types while the new one does not do that. Or am I wrong?

e.g., `"readOnly": "YES"` .. was changed to` "readOnly": false`

Question is also whether we want that to happen..

packages/td-tools/src/thing-model-helpers.ts
```diff
@@ -453,15 +447,15 @@ export class ThingModelHelpers {
         return tmpThingModels;
     }
 
-    private static getThingModelRef(data: Record<string, unknown>): Record<string, unknown> {
-        const refs = {} as Record<string, unknown>;
+    private static getThingModelRef(data: Record<string, unknown>): Record<string, string> {
```
danielpeintner, line 450: A general comment. Changes to ThingModel code should also go to https://github.com/eclipse-thingweb/td-tools/tree/main/node/thing-model later...

packages/td-tools/src/td-parser.ts
JKRhb: Thank you, that should be covered via a9d20792a0cebfb3f43aea0cc73730c7800a5dec now :)

danielpeintner (approved):
I looked over the changes and the they seem fine now. Thanks!

packages/core/src/codecs/netconf-codec.ts
JKRhb, line 1: By the way: I just realized that there are `netconf-codec.ts` files, one in the `core` package and one in the `binding-netconf` package. I wonder if we should remove one of the two (probably the one in the `core` package)?

JKRhb:
> I looked over the changes and the they seem fine now. Thanks!

Thank you! :) The PR would now be ready to be merged – however, I guess we should wait for #1302 et al. to resolve the issue with the eslint step.

packages/core/src/codecs/netconf-codec.ts
danielpeintner: uiii, strange and bad. I checked them 

* https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-netconf/src/codecs/netconf-codec.ts
  vs.
* https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/codecs/netconf-codec.ts

and they are similar but not the same at all 🙈

It seems the one in binding-netconf seems more sophisticated. @lukesmolo?

danielpeintner:
@relu91 are you okay with these changes?

packages/core/src/codecs/netconf-codec.ts
JKRhb: > I checked them
> 
>     * https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-netconf/src/codecs/netconf-codec.ts
>       vs.
> 
>     * https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/codecs/netconf-codec.ts
> 
> 
> and they are similar but not the same at all 🙈

Hmm, yeah, if I see it correctly, commit https://github.com/eclipse-thingweb/node-wot/commit/56797cdf422943a4db7f2b7d2abe506244e72b17 intended to move the codec to the binding directory but also left it in the core package 😅

relu91 (approved):

packages/core/src/codecs/netconf-codec.ts
relu91: I think this should be moved to its own issue.

danielpeintner: I created https://github.com/eclipse-thingweb/node-wot/issues/1304.
@JKRhb shall we fix it independently and you merge in master in this PR? Would that be okay for you?

JKRhb: As this one is already merged, I think we can continue in #1305 independently :)

danielpeintner: https://github.com/eclipse-thingweb/node-wot/pull/1305 has been merged. @JKRhb amybe you can merge in master and we can continue with merging this PR ...

packages/td-tools/src/thing-model-helpers.ts
danielpeintner: I created https://github.com/eclipse-thingweb/td-tools/issues/33

packages/core/src/codecs/netconf-codec.ts
danielpeintner: Ahh, I missed that the PR is alredy merged 🙈

PR #1211, opened 2024-01-08
fix(interaction-output): adjust storing behavior and return type of `value` method

JKRhb:
While dealing with the `InteractionOutput` class, I noticed that a small bug (due to the `!= null` check) was introduced that prevents already `parsedValue`s of type `null` from being properly returned. This PR solves the issue by replacing the check with `!== undefined`, so that `null` values are not being ignored anymore.

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -79,7 +79,9 @@ export class InteractionOutput implements WoT.InteractionOutput {
 
     async value<T>(): Promise<T> {
         // the value has been already read?
-        if (this.parsedValue != null) return this.parsedValue as T;
+        if (this.parsedValue !== undefined) {
+            return this.parsedValue as T;
+        }
 
         if (this.dataUsed) {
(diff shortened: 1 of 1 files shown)


JKRhb:
A general related question: Shouldn't the return type of this method rather be `Promise<DataSchemaValue>` instead of a generic `Promise<T>`?

relu91:
> A general related question: Shouldn't the return type of this method rather be `Promise<DataSchemaValue>` instead of a generic `Promise<T>`?

The correct way should be:
```typescript
value<T extends DataSchemaValue>(): Promise<T>
```

This should allow the user to do funny stuff like:
```typescript
 const b = await output.value<string>() // b inferred to be string without casting
```

JKRhb:
Thank you @relu91! I incorporated your feedback in fcfe687169fcf283cc85e267e645bcd52a7946c7 :)

relu91 (approved):
Thanks! Good to go!

danielpeintner (approved):

PR #1212, opened 2024-01-09
test(coap-server): don't shutdown servient unnecessarily

JKRhb:
This PR contains a potential fix for the CoAP server problems observed in #1195. As the servients are not used by the `CoapServer` class, I simply removed the invocations of `shutdown` to fix the issue at hand.

In general, I think we could discuss if passing the servient to the protocol interface implementations is actually necessary or if the passing of protocol-specific callbacks for retrieving credentials could be an alternative (this is the approach I used for dart_wot in the meantime). But I would open a dedicated issue for that later :)

### DIFF
--- a/packages/binding-coap/test/coap-server-test.ts
+++ b/packages/binding-coap/test/coap-server-test.ts
@@ -598,14 +598,13 @@ class CoapServerTest {
         });
 
         await coapServer.stop();
-        await servient.shutdown();
     }
 
     @test async "should reject unsupported methods for meta operations"() {
         const coapServer = new CoapServer();
         const servient = new Servient();
 
-        await coapServer.start(servient);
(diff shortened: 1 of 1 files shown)


relu91 (approved):
Good to go too, should we merge this first or maybe try it after we merged #1195 ?

JKRhb:
> Good to go too, should we merge this first or maybe try it after we merged #1195 ?

Either way would be fine for me :) If we merge #1195 first, then I would rebase against the new `master` state.

danielpeintner (approved):

danielpeintner:
I would merge this PR before https://github.com/eclipse-thingweb/node-wot/pull/1195 so that we have clean runs for all PRs that we merge. I hope this is okay for all..

PR #1214, opened 2024-01-10
fix(coap-server): add missing cov:observe subprotocol

JKRhb:
Trying to interact with a test thing using CoAP, I noticed that the forms provided in the Thing Descriptions do not include the `subprotocol` value `cov:observe` for observable properties and events. This PR tries to fix the issue by adding logic to determine if a `subprotocol` should be added to a form.

The implementation got a bit more complicated than initially expected because I noticed that you probably should not simply add the subprotocol to all property forms, but only those containing the op values `observeproperty` and/or `unobserveproperty`. Furthermore, I got the feeling that you should not use `cov:observe` with `readproperty` or `writeproperty` (which I think is currently not clearly stated in the CoAP Binding Template).

Therefore, during the TD generation process, I split the op values for read/write operations and observe/unobserve operations, and generated separate forms for the two kinds of operations. There might be some more discussion and clarification in the protocol binding document needed to decide whether `cov:observe` should or should not be used with forms containing `readproperty` and/or `writeproperty`. In general, the approach in this PR should work regardless of the eventual decision, though.

Besides the actual implementation, I performed some minor refactoring of the process of assigning the forms to affordances and added a test to cover all changes made.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -32,13 +32,15 @@ import { Server, createServer, registerFormat, IncomingMessage, OutgoingMessage 
 import slugify from "slugify";
 import { Readable } from "stream";
 import { MdnsIntroducer } from "./mdns-introducer";
-import { PropertyElement, DataSchema } from "wot-thing-description-types";
+import { PropertyElement, DataSchema, ActionElement, EventElement } from "wot-thing-description-types";
 import { CoapServerConfig } from "./coap";
 import { DataSchemaValue } from "wot-typescript-definitions";
 
 const { debug, warn, info, error } = createLoggers("binding-coap", "coap-server");
 
 type CoreLinkFormatParameters = Map<string, string[] | number[]>;
+
+type AffordanceElement = PropertyElement | ActionElement | EventElement;
 
 // TODO: Move to core?
 type AugmentedInteractionOptions = WoT.InteractionOptions & { formIndex: number };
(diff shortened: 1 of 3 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -228,21 +231,45 @@ export default class CoapServer implements ProtocolServer {
         return opValues;
     }
 
+    private addFormToAffordance(form: TD.Form, affordance: AffordanceElement): void {
+        const affordanceForms = affordance.forms;
+        if (affordanceForms == null) {
+            affordance.forms ??= [form];
```
relu91, line 237: Am I lost or this null coalescing operator is redundant?

packages/binding-coap/src/util.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/**
+ * Convenience function to filter out the `op` values "subscribeevent" and
+ * "unsubscribeevent" from a string array.
+ *
+ * @param opValues The `op` values to be filtered.
+ * @returns A filtered array that might be empty.
+ */
+export function filterEventOperations(opValues: string[]) {
```
relu91, line 56: Is this function used somewhere? I understand you added for completeness right? but given this is a private util file, maybe we can consider remove it...

relu91 (commented):
I have some questions below before approving the PR.

packages/binding-coap/src/coap-server.ts
JKRhb: True, the `??=` is not actually needed here :) Simplified it in 7109dcb913b4d3a8cb61caa5662751358767128d

relu91 (approved):
Good to go now.

JKRhb:
> Good to go now.

Thank you! I incorporated the latest change into 55e197ceb7d09d04ce4a5b2564dade8e823d8db6, squashing the commits. Hope that was fine.

Issue #1108, opened 2023-10-09
Simpler `examples` workflow using plain JS files and JSDoc comments


JKRhb:
At the moment, the [current workflow](https://github.com/eclipse-thingweb/node-wot/tree/master/packages/examples#workflow) for the examples prescribes using TypeScript for the source files whose resulting compiled JS files are then copied over to the top-level `examples` directory, including a manual removal of some added lines.

To keep both the benefits of TypeScript and the possibility to use plain JavaScript, I think a much simpler and more ergonomic approach would be to directly use JavaScript files and enable type checking via JS Doc comments. By using a `// @ts-check` comment at the top of the file, we would still get the benefit of type checks without the compilation and copying step. This would also give us more control over the resulting JS-based examples.

This suggestion corresponds partly with the discussion started by the recent shift of the Svelte library from TypeScript to plain JavaScript with JSDoc comments (see, e.g., [here](https://devclass.com/2023/05/11/typescript-is-not-worth-it-for-developing-libraries-says-svelte-author-as-team-switches-to-javascript-and-jsdoc/)).

JKRhb:
Just noticed that this would be a potential alternative to #185.

PR #1215, opened 2024-01-12
refactor: use type-checked JS instead of TS for quickstart examples

JKRhb:
This PR contains an initial proposal for a partial implementation of #1108, removing the (sort of redundant) TS files for the quickstart examples and instead adding type checking via a `@ts-check` comment to the respective JS example files. Additional type information is added via JSDoc comments.

This PR can serve as a basis for discussing if this is the direction we want to go with the other examples as well – if so, then I would open PRs for the other examples (or rather: those that should remain in the node-wot repository) as well.


relu91:
I'm ok exploring this, but the main reason why we had this "duplicated" code in the first time was to show people how the example would look in two different languages. Using `@js-check` will only make the JS code type-checked but we are going to lose the TS version of the example.

JKRhb:
(I initially linked the wrong issue in the PR description, I meant to be referring to #1108, which is now fixed.)

JKRhb:
> the main reason why we had this "duplicated" code in the first time was to show people how the example would look in two different languages.

Hmm, I see, I wasn't aware of that. I think the current JS examples might not be that great in that respect, though, since they contain some polyfills (e.g. for enums) and lines like `const core_1 = require("@node-wot/core");` which I don't find that nice. As mentioned in #1108, using JS with type-checks could be a more elegant solution than the current approach, since TypeScript is a superset of JavaScript anyway, so we would not have to maintain the pipeline of copying over the files from the examples package. But this is just a proposal that I wanted to discuss with you, if we should decide against it, that's also fine for me :)

packages/examples/src/quickstart/README.md
```diff
@@ -1,10 +0,0 @@
-# Quick Start Things
-
-This set of Things can be used to build mashup applications for a smart home scenario.
-These Things are:
-
--   Simple coffee machine that can take an order to brew a coffee.
--   Presence sensor that emits an event when a person is detected.
--   Smart clock that runs 60 times faster than real time, where 1 hour happens in 1 minute.
-
-These Things are hosted on plugfest.thingweb.io but can be also self-hosted.
```
danielpeintner, line 10: BTW, maybe we can adjust the link to https://www.thingweb.io/services
@egekorkan ?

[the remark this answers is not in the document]
JKRhb: Oh, a copy of it is actually already there, as it is not changed, it does not appear in the PR, though.

danielpeintner:
I tried it out locally and it works nice... similar behavior in VS code as with TS files 👍

On the one hand I agree with @relu91 that having samples for both TS and JS is beneficial.
On the other hand I also dislike the way we create out of _readable_ TS samples **less** readable (ugly looking?) JS samples.

Mhhh, not sure what we should do...
We definitely do not need having the same samples in both places.

I wonder what is most used TS or JS ?
How could we achieve a better split?

packages/examples/src/quickstart/README.md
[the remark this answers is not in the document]
danielpeintner: Oh I see 🙈

examples/quickstart/simple-coffee-machine.js
```diff
@@ -13,25 +12,39 @@
...
-const servient = new core_1.Servient();
+const servient = new Servient();
 // const staticAddress = "plugfest.thingweb.io";
 const staticAddress = "localhost"; // use this for testing locally
 const httpPort = 8081;
 servient.addServer(
-    new binding_http_1.HttpServer({
-        port: httpPort,
+    new HttpServer({
+        port: 8081,
```
danielpeintner, line 33: Why not use pre-defined const value ?

JKRhb:
> I tried it out locally and it works nice... similar behavior in VS code as with TS files 👍
> 
> On the one hand I agree with @relu91 that having samples for both TS and JS is beneficial. On the other hand I also dislike the way we create out of _readable_ TS samples **less** readable (ugly looking?) JS samples.
> 
> Mhhh, not sure what we should do... We definitely do not need having the same samples in both places.
> 
> I wonder what is most used TS or JS ? How could we achieve a better split?

Hmm, yeah, with Node.js, the situation is also not ideal since JavaScript with the CommonJS syntax is not a true subset of TypeScript (for that we would need to use ES Modules, but their compatibility with the older CommonJS approach is somewhat limited). Otherwise, we could cover both languages at the same time :/

However, as every TypeScript developer *should* be familiar with JavaScript, I would personally say that using only JS examples should cover the largest amount of people while making the examples themselves more accessible.

packages/examples/src/quickstart/README.md
JKRhb: I've adjusted the link accordingly in 5f123ea608346424d4bd055efd7b35c18e22ced3 :)

egekorkan:
I **think** I like this approach, some thoughts:
- I am not sure if outsiders find the ts examples anyways
- I also did not like the funny variable names in the JS examples
- Making sure to compile and do the sort of manual changes when I do contribution to examples is annoying.
- User perspective*: For anything that is a quick demo, I do it JS. If I use node-wot as a dependency of something bigger, I use TS. Easier to use JS files are nicer in this case. When building something bigger, sometimes I have issues with some types and want to see an example of their usage. These examples help but so would tests. This concern would be totally invalid once we have end to end tests, that are almost examples. Those test Things and Consumers would in TS and would address my need.


* I say user since I see myself more of a user of node-wot rather than a core developer.

danielpeintner:
Does it make sense to put this topic on the agenda of one of our next weekly calls?

JKRhb:
> EDIT: Maybe an according label would be also good so that we can easily find topics/issues for the calls...

I now created a label called "Discuss in Committer Meeting" for that :)

egekorkan:
Some discussions from the meeting today:
- Ege: A simple hand-written js example, like counter, is given to the users first. They can run it easily and understand node-wot. That example (its documentation) links to TS and JS examples of other examples we have, which are used for using node-wot in more advanced cases. Those other examples are maintained with a single codebase that we can use for testing etc.
- Jan: We need to check whether we can generate js-doc typed js examples. Like https://github.com/futurGH/ts-to-jsdoc
- We should check with @relu91

relu91:
Working with ES2021 does not seem to generate that ugly JS anymore, what do you think?

danielpeintner:
> Working with ES2021 does not seem to generate that ugly JS anymore, what do you think?

That is great! Once we move on with https://github.com/eclipse-thingweb/node-wot/pull/1340 we can also remove the workflow described here

https://github.com/eclipse-thingweb/node-wot/blob/5afa858b7a27ca664adc14850ce54c995c58acfa/packages/examples/README.md?plain=1#L14-L29

relu91:
Given that now the examples are correctly generated we decided to go in another direction: moving generated examples automatically every time we build the examples project.

PR #1217, opened 2024-01-16
Align `value` function to Scripting API

### LINKED ISSUE #1216: Should `InteractionOutput.value()` require a `DataSchema`?
I recently noticed that the `value` method of the `InteractionOutput` implementation currently is not compliant with the [algorithm](https://w3c.github.io/wot-scripting-api/#the-value-function) defined in the Scripting API specification. Here, step 3 of the algorithm demands that

> [i]f [...] _schema_ or its _type_ are `null` or `undefined`, then reject _promise_ with NotReadableError and stop. 

In our implementations (the same is currently true for dart_wot), however, no error is thrown if a `schema` should not be present:

https://github.com/eclipse-thingweb/node-wot/blob/6901b5a4a65af7c5a0c2509ac8cc3dfa881a3a38/packages/core/src/interaction-output.ts#L105-L116

For me, this raised the question of whether we should adjust the implementations or alter the algorithm in the specification. Adjusting the implementation would enforce the guarantee of receiving a valid `DataSchemaValue`, but it would make handling output data a bit less flexible. What would you prefer here?

relu91:
As discussed in #1216, the current implementation of the value function didn't take into account the validation of the DataSchema of the target affordance. This PR fix #1216 but it also introduces some required changes to take into account the new behavior in other places of the codebase. In particular, it updates the `exploreDirectory` method and `ThingDiscoveryProcess` class moving the fetching of the list of ThingDescirption from the method to the async iterator. The main reason for this change is that I figured out that with a DataSchema the `readProperty` would throw before entering even starting the `ThingDiscoveryProcess` which would be strange and violate the contract of the method `exploreDirectory` (errors for the list should be conveyed as errors in the ThingDesscovery object). Plus it has the benefit of lazy loading the list of TDs and they might be even controlled with `limit` and `offset` later. Handling the case in which the TDD does not have the DataSchema declared is still a TODO (but I can include in this PR). 

Finally, I fixed a minor problem in the binding-file that was preventing a clean test run on my local machine.

### DIFF
--- a/packages/binding-http/test/http-server-test.ts
+++ b/packages/binding-http/test/http-server-test.ts
@@ -160,7 +160,7 @@ class HttpServerTest {
         let test: DataSchemaValue;
         testThing.setPropertyReadHandler("test", (_) => Promise.resolve(test));
         testThing.setPropertyWriteHandler("test", async (value) => {
-            test = await value.value();
+            test = Buffer.from(await value.arrayBuffer()).toString("utf-8");
         });
 
         testThing.setActionHandler("try", async (input: WoT.InteractionOutput) => {
--- a/packages/binding-opcua/test/full-opcua-thing-test.ts
+++ b/packages/binding-opcua/test/full-opcua-thing-test.ts
@@ -371,7 +371,7 @@ describe("Full OPCUA Thing Test", () => {
             const content = await thing.readProperty(propertyName, localOptions);
             if (forceParsing) {
                 // In opcua binding it is possible to return a special response that contains
-                // reacher details than the bare value. However this make the returned value
+                // richer details than the bare value. However this makes the returned value
                 // not complaint with its data schema. Therefore we have to fallback to
                 // custom parsing.
                 const raw = await content.arrayBuffer();
--- a/packages/binding-opcua/test/full-opcua-thing-test.ts
+++ b/packages/binding-opcua/test/full-opcua-thing-test.ts
@@ -49,6 +49,7 @@ const thingDescription: WoT.ThingDescription = {
             observable: true,
             readOnly: true,
             unit: "°C",
+            type: "number",
             "opcua:nodeId": { root: "i=84", path: "/Objects/1:MySensor/2:ParameterSet/1:Temperature" },
             // Don't specifu type here as it could be multi form: type: [ "object", "number" ],
             forms: [
@@ -111,6 +112,7 @@ const thingDescription: WoT.ThingDescription = {
             description: "the temperature set point",
             observable: true,
             unit: "°C",
+            type: "number",
             // dont't
             forms: [
                 {
@@ -358,10 +360,25 @@ describe("Full OPCUA Thing Test", () => {
 
         return { thing, servient };
     }
-    async function doTest(thing: WoT.ConsumedThing, propertyName: string, localOptions: InteractionOptions) {
+    async function doTest(
+        thing: WoT.ConsumedThing,
+        propertyName: string,
+        localOptions: InteractionOptions,
+        forceParsing = false
+    ) {
         debug("------------------------------------------------------");
         try {
             const content = await thing.readProperty(propertyName, localOptions);
+            if (forceParsing) {
+                // In opcua binding it is possible to return a special response that contains
+                // reacher details than the bare value. However this make the returned value
+                // not complaint with its data schema. Therefore we have to fallback to
+                // custom parsing.
+                const raw = await content.arrayBuffer();
+                const json = JSON.parse(Buffer.from(raw).toString("utf-8"));
+                debug(json?.toString());
+                return json;
+            }
             const json = await content.value();
             debug(json?.toString());
             return json;
@@ -395,13 +412,13 @@ describe("Full OPCUA Thing Test", () => {
             const json1 = await doTest(thing, propertyName, { formIndex: 1 });
             expect(json1).to.eql(25);
 
-            const json2 = await doTest(thing, propertyName, { formIndex: 2 });
+            const json2 = await doTest(thing, propertyName, { formIndex: 2 }, true);
             expect(json2).to.eql({ Type: 11, Body: 25 });
 
             expect(thingDescription.properties?.temperature.forms[3].contentType).eql(
                 "application/opcua+json;type=DataValue"
             );
-            const json3 = await doTest(thing, propertyName, { formIndex: 3 });
+            const json3 = await doTest(thing, propertyName, { formIndex: 3 }, true);
             debug(json3?.toString());
             expect((json3 as Record<string, unknown>).Value).to.eql({ Type: 11, Body: 25 });
         } finally {
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -16,7 +16,7 @@ import * as util from "util";
 import * as WoT from "wot-typescript-definitions";
 import { ContentSerdes } from "./content-serdes";
 import { ProtocolHelpers } from "./core";
-import { DataSchemaError, NotSupportedError } from "./errors";
+import { DataSchemaError, NotReadableError, NotSupportedError } from "./errors";
 import { Content } from "./content";
 import Ajv from "ajv";
 import { createLoggers } from "./logger";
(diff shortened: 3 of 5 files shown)


relu91:
Ah it seems that I didn't properly rebase with master, after lunch I'll fix it. Sorry!

packages/core/src/wot-impl.ts
```diff
@@ -22,33 +22,34 @@ import Helpers from "./helpers";
...
         this.done = true;
     }
 
     async *[Symbol.asyncIterator](): AsyncIterator<WoT.ThingDescription> {
-        if (!(this.rawThingDescriptions instanceof Array)) {
-            this.error = new Error("Encountered an invalid output value.");
+        let rawThingDescriptions: WoT.ThingDescription[];
+        try {
+            const thingsPropertyOutput = await this.directory.readProperty("things");
+            rawThingDescriptions = (await thingsPropertyOutput.value()) as WoT.ThingDescription[];
```
JKRhb, line 45: Could we use the generic "version" of `value<T>` here to avoid the casting?

```suggestion
            rawThingDescriptions = await thingsPropertyOutput.value<WoT.ThingDescription[]>());
```

packages/core/src/interaction-output.ts
```diff
@@ -79,43 +79,47 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
         }
 
         // read fully the stream
-        const data = await this.content.toBuffer();
+        const bytes = await this.content.toBuffer();
         this.dataUsed = true;
-        this.buffer = data;
+        this.buffer = bytes;
 
         // call the contentToValue
```
JKRhb, line 109: Side note: This comment is kind of redundant and could also be removed, right?

```suggestion
```

packages/core/src/interaction-output.ts
```diff
@@ -79,43 +79,47 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
+            debug(`schema = ${util.inspect(this.schema, { depth: 10, colors: true })}`);
+            debug(`value: ${json}`);
+            debug(`Errror: ${validate.errors}`);
+            throw new DataSchemaError("Invalid value according to DataSchema", json as WoT.DataSchemaValue);
         }
 
-        this.parsedValue = value;
-        return this.parsedValue as T;
+        this.#value = json;
+        return this.#value as T;
```
JKRhb, line 123: Should we maybe include an additional check to make sure that the return value is really of type `T` here?

JKRhb (commented):
Thank you, @relu91! See a couple of initial comments below.

packages/core/src/wot-impl.ts
relu91: Sadly no, Given the fact that CosumeThing implementation returns `WoT.InteractionOutput` and not our `InteractionOutput`. As I said another time the generic in the value function is mostly experimental... we should think if it makes sense to have it in the public APIs.

packages/core/src/interaction-output.ts
relu91: the validate function does already all the checks, and the `json` variable is type of `T`. Maybe should we re-write it as:
```suggestion
        this.#value = json;
        return json;
```
?

relu91: Incorporated the feedback in 1148fab.

packages/core/src/wot-impl.ts
JKRhb: I see, then we can mark this thread as resolved :) Or should we open an issue for the evaluation of the generic in the value function?

packages/core/src/interaction-output.ts
```diff
@@ -33,7 +33,7 @@ const ajv = new Ajv({ strict: false });
 
 export class InteractionOutput implements WoT.InteractionOutput {
     private content: Content;
-    private parsedValue: unknown;
+    #value: unknown;
```
danielpeintner, line 36: Mhh, my feeling is that using `#` here inconsistent with the other variables. Either we change all or none I would say. Changing the name from `parsedValue`to `value` is fine by me.

packages/binding-http/test/http-server-test.ts
```diff
@@ -160,7 +160,7 @@ class HttpServerTest {
         let test: DataSchemaValue;
         testThing.setPropertyReadHandler("test", (_) => Promise.resolve(test));
         testThing.setPropertyWriteHandler("test", async (value) => {
-            test = await value.value();
+            test = Buffer.from(await value.arrayBuffer()).toString("utf-8");
```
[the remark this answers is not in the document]
relu91: the `test` property does not have the DataSchema defined. Later on in the code we are writing a value with content-type `text` therefore I decided not to add the Data Schema but simply parse the payload raw. We can choose, what is your preferred solution?

packages/core/src/interaction-output.ts
relu91: good point I was not sure how to align all of them. I'll add a refactor commit tomorrow. Yeah, changing the name to `value` is for having a nice 1-1 mapping with the spec.

packages/core/src/wot-impl.ts
relu91: I think the issue should be raised in the Scripting API repo, first (maybe it is already there?) and if it fails we have to evaluate how to do it here. @danielpeintner what do you think?

danielpeintner: I am not aware of a related issue in the Scripting API repo. Raising the aspect makes sense and maybe gives some more feedback from others...

packages/binding-http/test/http-server-test.ts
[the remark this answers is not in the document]
danielpeintner: I see, the *update* of throwing caused the issue. 
Since it is "our" test either way is fine by me 👍

packages/core/src/interaction-output.ts
danielpeintner: In the current state of the class I would favor using `value` as the the variable name without `#`

relu91: We can't because we have already a method called `value`.

danielpeintner: Mhh, then we need to pick another name _or_ would you go with `#value` (I wouldn't 🤷‍♂️)

relu91: Given that the specs says that the cached result is an internal slot called value, the 1-1 mapping with JS is having it to be `#value`. I like the coherence of the algorithm that we wrote, but I found calling it `#value` a bit odd too. It's a tradeoff that we have to decide. Maybe @JKRhb what's your take here?

danielpeintner: Yup, maybe @JKRhb has a good advice.

BTW, qhat about changing the other variables to # notation too. Doing so makes sense to me (even though a bit unrelated to this PR/change).

JKRhb: Maybe we could call it something like `internalValue`? This would correspond with the "internal slots" phrasing. However, I think I would also prefer the `#value` approach – couldn't we simply rename the other private variables as well (e.g., changing `content` to `#content`)?

danielpeintner: > couldn't we simply rename the other private variables as well (e.g., changing `content` to `#content`)?

That's what I meant as well

JKRhb: > > couldn't we simply rename the other private variables as well (e.g., changing `content` to `#content`)?
> 
> That's what I meant as well

Ah, sorry, I should have refreshed the page before posting my comment 😅🙈

relu91: Should be solved now in c10cdfc

packages/binding-opcua/test/full-opcua-thing-test.ts
```diff
@@ -358,10 +360,25 @@ describe("Full OPCUA Thing Test", () => {
...
+        propertyName: string,
+        localOptions: InteractionOptions,
+        forceParsing = false
+    ) {
         debug("------------------------------------------------------");
         try {
             const content = await thing.readProperty(propertyName, localOptions);
+            if (forceParsing) {
+                // In opcua binding it is possible to return a special response that contains
+                // reacher details than the bare value. However this make the returned value
```
danielpeintner, line 374:
```suggestion
                // richer details than the bare value. However this makes the returned value
```

danielpeintner (approved):
LGTM.

Some minor proposals below.

JKRhb (approved):

PR #1219, opened 2024-01-18
fix(octet-stream): correct limits for integer serialization and add tests

derwehr:
I noticed that the limits for integer serialization are wrong and corrected them. Additionally, I added test for these limits.

Changes in `octetstream-codec.ts`
- serializing `255` as unsigned integer with 8 bits failed, succeeds now;
- serializing `127` as signed integer with 8 bits failed, succeeds new;

Changes in `ContentSerdesTest.ts`
- added tests verifying that `255` as `uint8`, `127` and `-128` as `int8` succeed
- added tests verifying that `256` and `-1` as`unit8`, `128` and `-129` as `int8` fail

### DIFF
--- a/packages/core/test/ContentSerdesTest.ts
+++ b/packages/core/test/ContentSerdesTest.ts
@@ -578,6 +578,30 @@ class SerdesOctetTests {
         body = await content.toBuffer();
         expect(body).to.deep.equal(Buffer.from([0x80, 0x44]));
 
+        content = await ContentSerdes.valueToContent(
+            255,
+            { type: "uint8" },
+            `application/octet-stream;byteSeq=${Endianness.LITTLE_ENDIAN};length=1;`
+        );
+        body = await content.toBuffer();
+        expect(body).to.deep.equal(Buffer.from([0xff]));
+
+        content = await ContentSerdes.valueToContent(
+            127,
+            { type: "int8" },
+            `application/octet-stream;signed=true;length=1;`
+        );
+        body = await content.toBuffer();
+        expect(body).to.deep.equal(Buffer.from([0x7f]));
+
+        content = await ContentSerdes.valueToContent(
+            -128,
+            { type: "int8" },
+            `application/octet-stream;signed=true;length=1;`
+        );
+        body = await content.toBuffer();
+        expect(body).to.deep.equal(Buffer.from([0x80]));
+
         content = ContentSerdes.valueToContent(
             2345,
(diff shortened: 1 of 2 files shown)


packages/core/test/ContentSerdesTest.ts
```diff
@@ -742,11 +766,23 @@ class SerdesOctetTests {
     }
 
     @test "value to OctetStream should throw"() {
-        // @ts-ignore new dataschema types are not yet supported in the td type definitions
-        expect(() => ContentSerdes.valueToContent(2345, { type: "int8" }, "application/octet-stream")).to.throw(
+        expect(() => ContentSerdes.valueToContent(256, { type: "uint8" }, "application/octet-stream")).to.throw(
+            Error,
+            "Integer overflow when representing unsigned 256 in 8 bit(s)"
```
danielpeintner, line 771:
```suggestion
            "Integer overflow when representing 256 as uint8"
```

The error messages are confusing, same for the others. Or is it just me?

derwehr: It somewhat is, would you prefer something along:
```
input requires 9 bits; 8 bits were specified
```
or do you have a different proposal?

danielpeintner: I sketched above what I think is more clear. The same applies for the other messages...

"Integer overflow when representing unsigned 256 in 8 bit(s)"
-->
"Integer overflow when representing 256 as uint8"

derwehr: Sorry, somehow I completely missed the suggestion :smile:

The suggest can't be applied to the error message returned by the [test in lines 801 - 807](https://github.com/derwehr/node-wot/blob/1aa6559a92d282b9c3cc03cad28c34de0aadbb6f/packages/core/test/ContentSerdesTest.ts#L807) though, and it would require to add a `type` argument to the `valueToInteger` method.

Therefore, I'd suggest
```
Integer overflow when representing -2345 as a signed integer using 10 bit(s)
```

danielpeintner: That makes sense

danielpeintner (approved):

danielpeintner:
I noticed that there is a style check issue. Can you run `npm run format`...

danielpeintner:
@relu91, @JKRhb, others
any more comments before integrating the PR?

PR #1229, opened 2024-02-08
Events for simple coffee machine

egekorkan:
This is adding events to the simple coffee machine. With @relu91 we will use this in the nodejs conf presentation. There are those return statements added after event emissions since without doing that I was getting an event emission with data `"water""beans"`. Some race condition maybe? It is not an elegant solution but if that is indeed an issue with node-wot (and not my oversight), I would propose to go with this PR and open an issue about it.

### DIFF
--- a/examples/quickstart/simple-coffee-machine.js
+++ b/examples/quickstart/simple-coffee-machine.js
@@ -19,15 +19,18 @@ const core_1 = require("@node-wot/core");
 const binding_http_1 = require("@node-wot/binding-http");
 // create Servient add HTTP binding with port configuration
 const servient = new core_1.Servient();
+// const staticAddress = "plugfest.thingweb.io";
+const staticAddress = "localhost"; // use this for testing locally
+const httpPort = 8081;
 servient.addServer(
     new binding_http_1.HttpServer({
-        port: 8081,
+        port: httpPort,
     })
 );
-core_1.Helpers.setStaticAddress("plugfest.thingweb.io"); // comment this out if you are testing locally
-let waterAmount = 28;
-let beansAmount = 28;
-let milkAmount = 28;
+core_1.Helpers.setStaticAddress(staticAddress);
+let waterAmount = 1000;
+let beansAmount = 1000;
+let milkAmount = 1000;
 // promisify timeout since it does not return a promise
 function timeout(ms) {
     return new Promise((resolve) => setTimeout(resolve, ms));
@@ -73,15 +76,22 @@ servient.start().then((WoT) => {
             refill: {
                 synchronous: true,
                 input: {
-                    type: "string",
-                    enum: ["water", "beans", "milk"],
+                    type: "array",
+                    items: {
+                        type: "string",
+                        enum: ["water", "beans", "milk"],
+                    },
                 },
             },
         },
         events: {
             resourceEmpty: {
                 data: {
-                    enum: ["water", "beans", "milk"],
+                    type: "array",
+                    items: {
+                        type: "string",
+                        enum: ["water", "beans", "milk"],
+                    },
                 },
             },
         },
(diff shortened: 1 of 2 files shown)


examples/quickstart/simple-coffee-machine.js
```diff
@@ -99,6 +106,14 @@ servient.start().then((WoT) => {
                         waterAmount = waterAmount - 10;
                         beansAmount = beansAmount - 10;
                         thing.emitPropertyChange("resources");
+                        if (waterAmount <= 10) {
+                            thing.emitEvent("resourceEmpty", "water");
+                            return undefined;
```
danielpeintner, line 111: @egekorkan do you men this return is needed even though it follows also afterwards?

danielpeintner:
> without doing that I was getting an event emission with data `"water""beans"`. Some race condition maybe?

That would mean that when `emitPropertyChange` is called both if statements are true:
* waterAmount <= 10
* beansAmount <= 10

How does the code prevent that this is not happening?

examples/quickstart/simple-coffee-machine.js
egekorkan: So when I don't do the return after an event emission and the conditions are met to fire two events, the two events are fired at the same time.

danielpeintner: I see. What I wanted to understand is whether you get *one* event with both or 2 events (which would be okay..).
I think what you are saying is that there is just 1 event, right?

egekorkan:
> That would mean that when emitPropertyChange is called both if statements are true:

Just to be precise, it is the emitEvent that is the issue.

I am only sending a GET request from the browser to localhost:8081/coffee-machine/events/resourceEmpty and if both conditions are met, I get `"water""beans"`. Both payloads are somehow merged into one since they are emitted very quickly. I should have gotten only `"water"` since the browser will not send another GET before the other event could be emitted.

examples/quickstart/simple-coffee-machine.js
egekorkan: Yes 1 event from the Client side. Sorry, I also gave the same information below. We had a race condition ourselves here :D

egekorkan: I have changed this part of code so I am resolving this. The actual problem is an issue now

egekorkan:
I had to adapt the code a bit, a bit more than just adding an event. Now there are arrays for refilling and empty event so that there can be more resources empty etc. This avoid the race condition but also makes it more logical (both resources can finish at the same time and you want to fill both at the same time)

examples/quickstart/simple-coffee-machine.js
```diff
@@ -110,7 +139,22 @@ servient.start().then((WoT) => {
...
+                        }
+                        if (milkAmount <= 10) {
+                            resourceEvent.push("milk");
+                        }
+                        if (resourceEvent.length > 0) {
+                            thing.emitEvent("resourceEmpty", resourceEvent);
+                            return undefined;
+                        } else {
+                            return undefined;
+                        }
```
danielpeintner, line 157: Same here

examples/quickstart/simple-coffee-machine.js
```diff
@@ -120,7 +164,19 @@ servient.start().then((WoT) => {
...
+                        if (beansAmount <= 10) {
+                            resourceEvent.push("beans");
+                        }
+                        if (resourceEvent.length > 0) {
+                            thing.emitEvent("resourceEmpty", resourceEvent);
+                            return undefined;
+                        } else {
+                            return undefined;
+                        }
```
danielpeintner, line 179: And here also, just one return statement

egekorkan:
@danielpeintner adapted to your feedback

danielpeintner (approved):

relu91 (approved):
Since we are using the value function is better to cast than to ignore ts error. I fixed it and I think we can now merge.

relu91:
CI still randomly falling, merging anyway.

Issue #1228, opened 2024-02-08
InteractionOutput.value() throws an error for Actions with no output DataSchema


FadySalama:
I encountered this problem while interacting with the [TestThing](http://plugfest.thingweb.io:8083/testthing). When invoking the ``"void-void"`` action, the action results in an ``505: Internal Server Error`` with the message ``No schema defined``. This is thrown inside ``value()``  as seen below:

https://github.com/eclipse-thingweb/node-wot/blob/82f7f7acd1e68f80af9b53e9ef623d9a83334a39/packages/examples/src/testthing/testthing.ts#L240

As no data schema is needed for an action with no output, calling ``value()`` on the interaction output throws an error. However, this behavior seems a little counterintuitive. I would think, intuitively, that an action without an ``output`` ``DataSchema`` would either provide no ``InteractionOutput`` at all or would provide and ``InteractionOutput`` with a ``value()`` method that resolves with ``undefined`` and an ``arrayBuffer()`` method that resolves with an empty array (maybe also ``undefined`` but it makes more sense you would actually get an array, but it's empty because there is no output).

danielpeintner:
We could look whether a schema is available at all

https://github.com/eclipse-thingweb/node-wot/blob/82f7f7acd1e68f80af9b53e9ef623d9a83334a39/packages/core/src/interaction-output.ts#L94

Hence, we should not do anything if `this.schema == null ` and return `undefined` in such a case?

@relu91 @JKRhb any opinion?

danielpeintner:
I created draft PR https://github.com/eclipse-thingweb/node-wot/pull/1230 to show what I mean...

FadySalama:
Throwing an error if ```schema.type == null`` will throw in cases in which the schema is correctly defined using ``"oneof"``

egekorkan:
> Throwing an error if `schema.type == null` will throw in cases in which the schema is correctly defined using `"oneof"`

Related to https://github.com/w3c/wot-scripting-api/pull/534

relu91:
I think this is also related to satellite discussions: https://github.com/w3c/wot-thing-description/issues/1234 and in particular https://github.com/w3c/wot-thing-description/issues/1441. My opinion here is that the implementation is still correct, but we are lacking a way to clearly express the fact that we are expecting an empty payload.

PR #1230, opened 2024-02-09
refactor: report undefined for no expected value()

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1228

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -82,8 +82,7 @@ export class InteractionOutput implements WoT.InteractionOutput {
     async value<T extends WoT.DataSchemaValue>(): Promise<T> {
         // is there any value expected at all?
         if (this.schema == null) {
-            // #value is undefined at this point
-            return this.#value as T;
+            return undefined as unknown as T;
         }
 
         // the value has been already read?
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -62,8 +62,6 @@ export class InteractionOutput implements WoT.InteractionOutput {
     }
 
     async arrayBuffer(): Promise<ArrayBuffer> {
-        // TODO check for this.schema == null also and return empty array?
-
         if (this.#buffer) {
             return this.#buffer;
         }
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -81,7 +81,7 @@ export class InteractionOutput implements WoT.InteractionOutput {
         // is there any value expected at all?
         if (this.schema == null) {
             warn(
-                `No schema defined. Hence undefined is reported for value() function. If you need more control consider using arrayBuffer().`
+                `No schema defined. Hence undefined is reported for value() function. If you are invoking an action with no output that is on purpose, otherwise consider using arrayBuffer().`
             );
             return undefined as unknown as T;
         }
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -82,6 +82,7 @@ export class InteractionOutput implements WoT.InteractionOutput {
     async value<T extends WoT.DataSchemaValue>(): Promise<T> {
         // is there any value expected at all?
         if (this.schema == null) {
+            // #value is undefined at this point
             return this.#value as T;
         }
 
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -21,7 +21,7 @@ import { Content } from "./content";
 import Ajv from "ajv";
 import { createLoggers } from "./logger";
 
-const { debug } = createLoggers("core", "interaction-output");
+const { debug, warn } = createLoggers("core", "interaction-output");
 
 // Problem: strict mode ajv does not accept unknown keywords in schemas
 // however property affordances could contain all sort of fields
@@ -80,6 +80,9 @@ export class InteractionOutput implements WoT.InteractionOutput {
     async value<T extends WoT.DataSchemaValue>(): Promise<T> {
         // is there any value expected at all?
         if (this.schema == null) {
+            warn(
+                `No schema defined. Hence undefined is reported for value() function. If you need more control consider using arrayBuffer().`
+            );
             return undefined as unknown as T;
         }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/interaction-output.ts
```diff
@@ -78,6 +80,11 @@ export class InteractionOutput implements WoT.InteractionOutput {
     }
 
     async value<T extends WoT.DataSchemaValue>(): Promise<T> {
+        // is there any value expected at all?
+        if (this.schema == null) {
+            return this.#value as T;
```
JKRhb, line 85: At this point, `this.#value` would be `undefined`, right? Maybe we could return that value explicitly to make it more obvious? However, this might not make that much sense when it comes to the cast to `T` 🤔

```suggestion
            return undefined as T;
```

danielpeintner: Exactly that was the problem...  hence I went the _easy_ way 😉.

image

Maybe we can just add a comment that it returns `undefined` ?

JKRhb: > Exactly that was the problem... hence I went the easy way 😉.

Oh, I see!

> Maybe we can just add a comment that it returns undefined ?

Sure, maybe something like this? :)

```suggestion
            // #value is undefined at this point
            return this.#value as T;
```

danielpeintner:
We have a failing test

https://github.com/eclipse-thingweb/node-wot/actions/runs/7889804138/job/21530521048#step:9:198

`should throw if the stream was accessed before calling value`

@relu91 mhh not sure if the test does the correct thing in this case

packages/core/src/interaction-output.ts
danielpeintner: @JKRhb I made returning undefined more explicitly, see https://github.com/eclipse-thingweb/node-wot/pull/1230/commits/04d96864bfc41c413d1da2c15e9eeb880d0d2830
It doesn't look nice either 🙈

danielpeintner:
> We have a failing test

I fixed the test by adding a DataSchema which is now required, see https://github.com/eclipse-thingweb/node-wot/pull/1230/commits/051febc72faca2659f277214ed9dc542cf6dac19

packages/core/src/interaction-output.ts
```diff
@@ -62,6 +62,8 @@ export class InteractionOutput implements WoT.InteractionOutput {
     }
 
     async arrayBuffer(): Promise<ArrayBuffer> {
+        // TODO check for this.schema == null also and return empty array?
```
danielpeintner, line 65: Question: Should we return an empty array for `arrayBuffer()` calls also, if there is no schema?

JKRhb: Hmm, looking at the corresponding [algorithm](https://w3c.github.io/wot-scripting-api/#the-arraybuffer-function), I would say that `arrayBuffer()` should not take the schema into account at all (as it would also be the alternative method to use when the `value()` method fails).

relu91: Yes I agree, however, I still think that the issue has to be solved in the specification (TD) first...

danielpeintner: > Yes I agree

Makes, sense, removed comment with https://github.com/eclipse-thingweb/node-wot/pull/1230/commits/0a9a69be9720c76f4f3af473ccfaf1a85eb0c1d2

> however, I still think that the issue has to be solved in the specification (TD) first...

I know, that there is something to fix/decide. Anyhow, I think this is still an improvement and fixes https://github.com/eclipse-thingweb/node-wot/issues/1228. Other/more changes can come later.
I let you decide...

relu91: I might be ok for the sake of "let's fix this quickly" and then find a solution later. But can I ask you to add a warning log in the code and open an issue to keep this tracked?

danielpeintner: I created https://github.com/eclipse-thingweb/node-wot/issues/1243 to keep track of the _broader_ issue.

Anyhow, I am not sure what "warning log" I should/can add in the code. It is not really the fault of the user. It is up to **us** to fix the issues in TD/Scripting spec etc.
Do you have something specific in mind?

relu91: I'd warn the user that he reading a value from a TD that does not have an explicit `schema` defined and we are treating it as if there is no output value.  There might not be what the TD designer wanted.

danielpeintner: In https://github.com/eclipse-thingweb/node-wot/pull/1230/commits/56133b995c0f9048a76a6af4d118793bed84341c I added a warn log. Please let me know whether that is sufficient.

packages/core/src/interaction-output.ts
```diff
@@ -78,6 +78,14 @@ export class InteractionOutput implements WoT.InteractionOutput {
     }
 
     async value<T extends WoT.DataSchemaValue>(): Promise<T> {
+        // is there any value expected at all?
+        if (this.schema == null) {
+            warn(
+                `No schema defined. Hence undefined is reported for value() function. If you need more control consider using arrayBuffer().`
+            );
```
danielpeintner, line 85: The warning might confuse people that call `value()` on actions that do not have an return value. We could also just use `debug`.
@relu91 ?

relu91: What about:
```suggestion
            warn(
                `No schema defined. Hence undefined is reported for value() function. If you are invoking an action with no output that is on purpose otherwise prefer using arrayBuffer().`
            );
```

danielpeintner: Suggestion is fine by me...  I am not sure though whether it should read "prefer using". I think it should still be "consider using". Anyhow, no strong objection 👍

relu91: Ok for "consider using"

danielpeintner: Great. Fixed with https://github.com/eclipse-thingweb/node-wot/pull/1230/commits/be93623330bb2be13751f25095721b0d8e8afa2d.

packages/core/src/interaction-output.ts
danielpeintner: Comment resolved by now

relu91 (approved):
Thanks @danielpeintner and sorry for being a little bit stubborn on this issue :)

JKRhb (approved):

PR #1244, opened 2024-02-21
fix(binding-http): fill security scheme when baseURI defined

relu91:
Found a bug when exposing things with a `baseURI` defined. I added a failing test to verify the correct behavior and avoid regression.

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -276,6 +276,11 @@ export default class HttpServer implements ProtocolServer {
         if (this.getPort() !== -1) {
             debug(`HttpServer on port ${this.getPort()} exposes '${thing.title}' as unique '/${urlPath}'`);
             this.things.set(urlPath, thing);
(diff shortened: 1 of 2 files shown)


danielpeintner (approved):
LGTM

The only _concern_ I had initially was whether it makes any difference to put 
`this.fillSecurityScheme(thing);`

after the `if else`  ... as it was before but I don't think this matters in this case.

relu91:
I reviewed the code and I can confirm that `addEnpoint` does not have a direct dependency on what `fillSecurityScheme` is doing. You can check the code and see that `addEnpoint` does not manipulate anything other than affordances objects inside the `thing`. I would say that it is safe to keep the code as proposed and it will increase readability (e.g. since security schemes are often coming before affordances the coder is less surprised to see them handled first).

JKRhb (approved):

danielpeintner (approved):
Latest changes look good again 👍

Issue #1250, opened 2024-03-08
Response not valid if charset is indicated in header


egekorkan:
In line https://github.com/eclipse-thingweb/node-wot/blob/0b7b108a0de0477dffd5d9d258fe90ca509a86d7/packages/core/src/consumed-thing.ts#L682, there is a check whether the header content type matches the content type in the TD. This is a strict string check but the header can actually contain charset information. For example, JSON is UTF-8 encoded by default so there is no need to indicate `"contentType":"application/json; charset=utf-8` in the TD. However, some HTTP servers like the popular expressjs [put the charset information _to be sure_](https://github.com/expressjs/express/blob/1b51edac7c5f2844e23602164a52643bb625993a/lib/response.js#L174). So an expressjs Thing consumed by node-wot MUST have `"contentType":"application/json; charset=utf-8` otherwise one gets the following error:

> node-wot:binding-http:http-client-impl:debug HttpClient received Content-Type: application/json; charset=utf-8 +1ms
Error: Unexpected type in response

So one side, the WoT specs should be more clear on this possibility but node-wot needs to be more flexible as well. Since this code is in the consumed thing of core, we can only check for inclusion of the form's content type in the response header but not a direct comparison. A more elegant solution would be to move these kinds of checks to each content serdes implementation. This way, JSON content serdes, which knows what are the defaults, can do the checks.

Found the bug via https://github.com/eclipse-thingweb/test-things/pull/16

egekorkan:
The _interesting_ part is that this check is only there for an action whose form contain a response where the payload for input and output have different media types. It is fine if a property read results with a `"application/json; charset=utf-8"`. In this case, the easiest solution would be to remove that check in the action.

PR #1251, opened 2024-03-11
refactor: handle also enhanced contentTypes

danielpeintner:
e.g., expected "application/json" should deal with "application/json; charset=utf-8"

fixes https://github.com/eclipse-thingweb/node-wot/issues/1250

### DIFF
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -745,8 +745,6 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
             formWithoutURITemplates,
             // next
             (content) => {
-                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- tsc get confused when nullables are to listeners lambdas
-                if (!content.type) content.type = form!.contentType ?? "application/json";
                 try {
                     listener(this.handleInteractionOutput(content, form, tp));
                 } catch (e) {
@@ -803,8 +801,6 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         await client.subscribeResource(
             formWithoutURITemplates,
             (content) => {
-                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- tsc get confused when nullables are to listeners lambdas
-                if (!content.type) content.type = form!.contentType ?? "application/json";
                 try {
                     listener(this.handleInteractionOutput(content, form, te.data));
                 } catch (e) {
--- a/packages/core/src/consumed-thing.ts
+++ b/packages/core/src/consumed-thing.ts
@@ -679,8 +679,12 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
 
         // check if returned media type is the same as expected media type (from TD)
         if (form.response != null) {
-            if (content.type !== form.response.contentType) {
-                throw new Error(`Unexpected type in response`);
+            const indexOfSemicolon = content.type.indexOf(";");
+            const contentType = indexOfSemicolon < 0 ? content.type : content.type.substring(0, indexOfSemicolon);
+            if (contentType !== form.response.contentType) {
+                throw new Error(
+                    `Unexpected type '${content.type}' in response. Should be '${form.response.contentType}'`
+                );
             }
         }
         try {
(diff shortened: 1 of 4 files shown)


danielpeintner:
Question: is it enough to look into the contentType before ":"

Note: I assume if we have `response: { contentType: "application/json; charset=utf-8" }` in a TD we expect to get exactly this encoding  ... even though we could run into issues with multiple spaces after ";" etc. Not sure if we want to handle all this..

egekorkan:
How about checking for string inclusion instead?

danielpeintner:
> How about checking for string inclusion instead?

I think to get it right we might want to use a dedicated library like
* https://www.npmjs.com/package/content-type-parser or
* https://www.npmjs.com/package/content-type

I fear that there might be some edge cases we miss otherwise...

danielpeintner:
Note: I updated accordingly
* added the same InteractionOutput handling for readAttribute also
* use dedicated library to parse contentType

packages/core/src/consumed-thing.ts
```diff
@@ -555,7 +557,32 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
...
+        return this.handleInteractionOutput(content, form, tp);
+    }
+
+    private handleInteractionOutput(
+        content: Content,
+        form: TD.Form,
+        outputDataSchema: WoT.DataSchema | undefined
+    ): InteractionOutput {
+        // infer media type from form if not in response metadata
+        if (!content.type) content.type = form.contentType ?? "application/json";
```
JKRhb, line 569: I think this *could* be shortened very slightly, like so:

```suggestion
        content.type ??= form.contentType ?? "application/json";
```

However, should the `type` within `content` be overwritten here? As the `content` is not returned again by this method, maybe we could consider something like this instead:

```suggestion
        const contentType = content.type ?? form.contentType ?? "application/json";
```

JKRhb (commented):
Looks good to me in general :) See one relatively minor comment below:

danielpeintner:
> Do we want to apply this to all interaction styles in this PR?

https://github.com/eclipse-thingweb/node-wot/pull/1251/commits/01e0b84b4630356785191f66d001de316678aef8 adds teh same handling for observeProperty and event

packages/core/src/consumed-thing.ts
```diff
@@ -555,7 +557,33 @@ export default class ConsumedThing extends TD.Thing implements IConsumedThing {
         form = this.handleUriVariables(tp, form, options);
 
         const content = await client.readResource(form);
-        return new InteractionOutput(content, form, tp);
+        try {
+            return this.handleInteractionOutput(content, form, tp);
+        } catch (e) {
+            const error = e instanceof Error ? e : new Error(JSON.stringify(e));
+            throw new Error(`Error while processing property for ${tp.title}. ${error.message}`);
+        }
```
JKRhb, line 565: I am wondering if we could maybe also simplify these try-catch-blocks, as the `handleInteractionOutput` method itself is probably the only place where an error could originate from, right? But this would only be an optimization step.

danielpeintner: I was thinking about the same. In the first commits the `try catch` block was inside `handleInteractionOutput()`. 

The downside was/is that either the error message is not very clear (not stating for which interaction name & type it originates from) OR we need to pass additional arguments like interaction-type (property, action, event) and interaction-name is used to `handleInteractionOutput()` purely for logging.
Personally I didn't like that much and hence decided to do the error handling in each interaction even though I know it duplicates code.

Anyhow, I am open to either way...

JKRhb: > The downside was/is that either the error message is not very clear (not stating for which interaction name & type it originates from) OR we need to pass additional arguments like interaction-type (property, action, event) and interaction-name is used to handleInteractionOutput() purely for logging.
> Personally I didn't like that much and hence decided to do the error handling in each interaction even though I know it duplicates code.

Yeah, I agree, passing these parameters only for logging is definitely not that ideal. So I think we can keep it as is :) Maybe one potential solution for the future could be introducing a common interface for `ConsumedThingProperty`, `ConsumedThingAction`, `ConsumedThinEvent` and then passing the respective interaction affordance to the handler function which could then obtain the necessary information for logging.

relu91 (approved):
In general, I'm a little bit reluctant to add external libraries to the core package unless they are really something needed. However, in this case, I can make an exception given the fact that the package is really just one single file. We can merge!

JKRhb (approved):

PR #1253, opened 2024-03-17
Align ajv usage

### LINKED ISSUE #1245: Aligning ajv usage
Currently, there are two ajv inclusions in package.json. One is in core package and the other one is in opcua binding. They are using the same version, thankfully. However, the ajv-formats is not part of the core package, which results in the following warning message:

```sh
unknown format "date-time" ignored in schema at path "#"
```

This will result in node-wot not being able validate formats passed in the string data schema.

So I propose the following:
- removing ajv and ajv-formats dependency from binding-opcua
- Adding ajv-formats to core package as a dependency (see https://ajv.js.org/packages/ajv-formats.html). 

Until TD removes the format keyword, we have to support it in one way or another in node-wot.

Also related to https://github.com/eclipse-thingweb/node-wot/issues/754

egekorkan:
fixes #1245


egekorkan:
The failures are not related to this PR. Also, they make sense since a TM cannot be fetched from example.com. However, I do not know how it passed so far...

JKRhb (approved):
LGTM :) Should the `package-lock.json` also be updated before merging, though (by running `npm install`)?

egekorkan:
@JKRhb you are right, sorry for the oversight

danielpeintner:
> The failures are not related to this PR.

Mhh, I agree the changes look sane.
Anyhow, It seems strange to me that not a single test run succeeds. I did try several times now... seems suspicious. It always fails for

```
  1) tests to verify the Thing Model Helper
       should fail on unavailable linked ThingModel - https:

      AssertionError: expected promise to be rejected with an error including 'https status code not 200 but 404 for…' but got 'https status code not 200 but 500 for…'
      + expected - actual

      -https status code not 200 but 500 for https://example.com/models/colored-lamp-1.0.0.tm.jsonld
      +https status code not 200 but 404 for https://example.com/models/colored-lamp-1.0.0.tm.jsonld
```


I checked it out locally and everything is fine. Hence I created https://github.com/eclipse-thingweb/node-wot/issues/1254.

egekorkan:
Btw we should NOT merge this since I'm actually not using the formats yet or at least not close the linked issue.

packages/td-tools/test/ThingModelHelperTest.ts
```diff
@@ -489,7 +486,7 @@ class ThingModelHelperTest {
 
         await expect(this.thingModelHelpers.getPartialTDs(thing)).to.be.rejectedWith(
             Error,
-            "https status code not 200 but 404 for https://example.com/models/colored-lamp-1.0.0.tm.jsonld"
+            "https status code not 200"
```
danielpeintner, line 489: This PR includes now changes that belong to another PR

egekorkan: Sorry about that. I forgot to pull the master beforehand. Now only this PR changes are included

relu91 (approved):

danielpeintner (approved):

Issue #1252, opened 2024-03-14
BUG: Error when exposing a Thing because "structuredClone" is not defined


idilsezgin:
Whenever I try to expose a Thing, I get this error
```bash
const tdSchemaCopy = structuredClone(tdSchema);
                             ^

ReferenceError: structuredClone is not defined
    at Function.createExposeThingInitSchema (..\node_modules\@node-wot\core\dist\helpers.js:182:30)
    at Object.<anonymous> (..\TestingTD\node_modules\@node-wot\core\dist\helpers.js:317:49)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (..\TestingTD\node_modules\@node-wot\core\dist\exposed-thing.js:51:35)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
```
I checked and `structuredClone` is not defined in helpers.ts but also in consumed-thing.ts and http-server.ts (See below)
 https://github.com/eclipse-thingweb/node-wot/blob/0b7b108a0de0477dffd5d9d258fe90ca509a86d7/packages/core/src/helpers.ts#L223

https://github.com/eclipse-thingweb/node-wot/blob/0b7b108a0de0477dffd5d9d258fe90ca509a86d7/packages/core/src/consumed-thing.ts#L368

https://github.com/eclipse-thingweb/node-wot/blob/0b7b108a0de0477dffd5d9d258fe90ca509a86d7/packages/binding-http/src/http-server.ts#L323

Confirmed with @FadySalama

danielpeintner:
Mhh, are you using a very old version of node or what is the context you are using it?

In https://github.com/eclipse-thingweb/node-wot/pull/1105 we introduced "structuredClone" and partly reverted it in https://github.com/eclipse-thingweb/node-wot/pull/1152 because we were running into issues.

Proposal to @JKRhb and others
I would create a helper function `Helpers.structuredClone()` that for now uses the old fashion cloning until we are sure the official `structuredClone()` does no longer cause issues. We can  point all places to this helper function...

[...]

PR #1258, opened 2024-03-22
refactor: create helper function structuredClone and use *old* way

danielpeintner:
point all functions to it so that in future we can easily update

fixes https://github.com/eclipse-thingweb/node-wot/issues/1252

### DIFF
--- a/packages/binding-http/src/http-server.ts
+++ b/packages/binding-http/src/http-server.ts
@@ -320,7 +320,7 @@ export default class HttpServer implements ProtocolServer {
             for (const [inUri, toUri] of Object.entries(this.urlRewrite)) {
                 const endsWithToUri: boolean = form.href.endsWith(toUri);
                 if (endsWithToUri) {
-                    const form2 = structuredClone(form);
+                    const form2 = Helpers.structuredClone(form);
                     form2.href = form2.href.substring(0, form.href.lastIndexOf(toUri)) + inUri;
                     forms.push(form2);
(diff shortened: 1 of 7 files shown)


danielpeintner:
Note: in package `td-tools` we use `JSON.parse(JSON.stringify())` also but I don't think the helper functions belongs to `td-tools` package.

https://github.com/eclipse-thingweb/node-wot/blob/1d2e67e04dc7805547aa901ab2b2c550f256e1e7/packages/td-tools/src/td-parser.ts#L222

Hence we can just use it in packages that depend on `core`.

relu91:
I'm not sure that we should add this if we are not 100% sure that this is a bug of the current supported Nodejs versions. From the issue is not clear if the submitter is just using and old version or not.

danielpeintner:
I just feel it is not worth to keep the built-in `structuredClone` if it causes issues. After it was first introduced in https://github.com/eclipse-thingweb/node-wot/pull/1105 we had to partly revert it in https://github.com/eclipse-thingweb/node-wot/pull/1152 because of issues. Now we see new issues popping up.

Hence I think it is an improvement until it is better supported. Moreover, with this change we can easily update the code once we feel comfortable.
Anyhow, if others think we should not make the change it is fine by me also.

relu91 (approved):
As pointed into the commiter call I'm ok merging this PR, it's just refectoring the code to have a single point where to update. We can refactor later if needed.

Issue #1168, opened 2023-11-23
Align Modbus prefix with binding document


danielpeintner:
According to https://github.com/w3c/wot-binding-templates/issues/119#issuecomment-1823043348 the Modbus prefix will become `modv`. Once the change is applied we need to update our code as well since we use `modbus` as prefix for now.

danielpeintner:
The `modv` prefix is in place now, see https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html

Are the changes in node-wot only about changing "modbus:" to "modv:" or is there more to align?
@relu91 do you know?

relu91:
Todo items:
- Support `modv:type` see possible [values](https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html#payloaddatatype). Sadly, I think this is not going to be trivial given the current node-wot architecture and the fact that we handled that conversions with `ContentSerders` and DataSchema.
- Support `modv:mostSignificantByte` `modv:mostSignificantByte`. Same for this but maybe easier because we can use directly Buffer primitives. 

Other than these two, we are aligned.

egekorkan:
Also, it is expecting the `modbus:address` field to be present

relu91:
Fixed in the above linked PR.

PR #1261, opened 2024-04-02
Modbus binding alignment

### LINKED ISSUE #1234: Modbus readme is outdated
The modbus readme is not very friendly like the other ones and it is outdated.

Not friendly: The modbus client instantiation is not examplified (see http example https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-http/README.md#client-example)

Outdated: It is using the very outdated api that uses `read()` instead of `readProperty()`: https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-modbus/README.md#combination-using-the-node-wot-api

egekorkan:
Fixes #1234 and #1168 (when the PR is ready ofc).

### DIFF
--- a/packages/binding-modbus/README.md
+++ b/packages/binding-modbus/README.md
@@ -22,7 +22,7 @@ let servient = new Servient();
 servient.addClientFactory(new ModbusClientFactory());
 
 async function main() {
-    let td = {} // see https://github.com/eclipse-thingweb/test-things/blob/main/things/elevator/modbus/js/modbus-elevator.td.json
+    let td = {}; // see https://github.com/eclipse-thingweb/test-things/blob/main/things/elevator/modbus/js/modbus-elevator.td.json
 
     const WoT = await servient.start();
     const thing = await WoT.consume(td);
--- a/packages/binding-modbus/README.md
+++ b/packages/binding-modbus/README.md
@@ -130,15 +130,15 @@ To help the MODBUS binding to perform combination a user of the API should creat
 
 ```javascript
 console.info("Creating promise vl1n");
-let vl1n = pac3200.properties["Voltage/L1N"].read();
+let vl1n = pac3200Thing.readProperty("Voltage/L1N");
 console.info("Creating promise vl2n");
-let vl2n = pac3200.properties["Voltage/L2N"].read();
+let vl2n = pac3200Thing.readProperty("Voltage/L2N");
 console.info("Creating promise vl3n");
-let vl3n = pac3200.properties["Voltage/L3N"].read();
+let vl3n = pac3200Thing.readProperty("Voltage/L3N");
 
 console.info("Resolving all promises");
 Promise.all([vl1n, vl2n, vl3n])
-    .then((values) => values.forEach((v) => console.info("Voltage = ", v)))
+    .then((values) => values.forEach((v) => console.info("Voltage = ", await v.value())))
     .catch((reason) => console.warn("Failed ", reason));
 ```
 
@@ -152,12 +152,10 @@ Reads the 8th input register of the unit 1
 
 ```json
 {
-    "href": "modbus+tcp://127.0.0.1:60000",
+    "href": "modbus+tcp://127.0.0.1:60000/1/8",
     "contentType": "application/octet-stream;length=2",
     "op": ["readproperty"],
     "modv:function": "readInputRegister",
-    "modv:address": 8,
-    "modv:unitID": 1,
     "modv:timeout": 2000
 }
 ```
@@ -168,12 +166,10 @@ Read and write the 8th holding register of the unit 1
 
 ```json
 {
-    "href": "modbus+tcp://127.0.0.1:60000",
+    "href": "modbus+tcp://127.0.0.1:60000/1/8",
     "contentType": "application/octet-stream;length=2",
     "op": ["readproperty", "writeproperty"],
-    "modv:entity": "HoldingRegister",
-    "modv:address": 8,
-    "modv:unitID": 1
+    "modv:entity": "HoldingRegister"
 }
 ```
 
@@ -183,12 +179,10 @@ Polls the 8th holding register of the unit 1 every second.
 
 ```json
 {
-    "href": "modbus+tcp://127.0.0.1:60000",
+    "href": "modbus+tcp://127.0.0.1:60000/1/8",
     "contentType": "application/octet-stream;length=2",
     "op": ["observeproperty"],
     "modv:entity": "HoldingRegister",
-    "modv:address": 8,
-    "modv:unitID": 1,
     "modv:pollingTime": 1000
 }
 ```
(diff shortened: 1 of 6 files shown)


egekorkan:
Below is a code I have used that does full TD parsing. I will edit the code based on the changes: https://gist.github.com/egekorkan/e2489da6ca5528720ea10f601c4a0166

It acts against https://github.com/eclipse-thingweb/test-things/blob/main/things/elevator/modbus/js/main.js with modified TD (in the code) but unmodified thing code

packages/binding-modbus/src/modbus-client.ts
```diff
@@ -243,31 +247,31 @@ export default class ModbusClient implements ProtocolClient {
         const mode = contentLength > 0 ? "w" : "r";
 
         // Use form values if provided, otherwise use form values (we are more merciful then the spec for retro-compatibility)
```
relu91, line 249:
```suggestion
        // Use URI values if provided, otherwise use form values (we are more merciful then the spec for retro-compatibility)
```
? Or even get rid of this comment since we are not that merciful anymore 😆

packages/binding-modbus/src/modbus-client.ts
```diff
@@ -212,22 +212,26 @@ export default class ModbusClient implements ProtocolClient {
         return endianness;
     }
 
-    private overrideFormFromURLPath(input: ModbusForm) {
+    // This generates a form with url content based on the uri scheme
+    // Ideally, more code should be refactored to use uri only
+    private generateFormFromURLPath(input: ModbusForm) {
```
relu91, line 217: I'm ok with finding a better name for this function. However, `generate` does not really explain the fact that the function actually modifies the form in input with the values taken from the URI. We can clone the form input and make it immutable to better fit the description.

relu91 (approved):
Good to go, just some minor point below. Also a question: the `modv:type` and brothers are not yet implemented right?

packages/binding-modbus/src/modbus-client.ts
egekorkan: Good point. We are overriding the form but more that we are adding terms into it whereas before the URI values would be prioritized (no code change though). How about `addFormElementsFromURLPath`

packages/binding-modbus/src/modbus-client.ts
egekorkan: I have adjusted it completely

egekorkan:
> Also a question: the modv:type and brothers are not yet implemented right?

No not yet. I didn't have time for those but I can do them as a follow-up PR.

packages/binding-modbus/src/modbus-client.ts
danielpeintner: I like the idea of not changing the input form and rather clone and update. Doing so the term `generate` with reporting back the update is fine I think

packages/binding-modbus/README.md
```diff
@@ -3,62 +3,98 @@
...
-This package uses [modbus-serial](https://www.npmjs.com/package/modbus-serial) as a low level client for Modbus TCP.
+This package uses [modbus-serial](https://www.npmjs.com/package/modbus-serial) as a low-level client for Modbus TCP.
 W3C WoT Binding Template for Modbus can be found [here](https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/index.html).
 
 Current Maintainer(s): [@relu91](https://github.com/relu91) [@fillobotto](https://github.com/fillobotto)
 
-## Protocol specifier
+## Client Example
+
+You can use a code like the following to use the binding. This specific code is interacting with one of the Eclipse Thingweb Test Things at <https://github.com/eclipse-thingweb/test-things/tree/main/things/elevator>.
```
danielpeintner, line 13: Note: Pointing to other repos/folders makes it even more likely that it gets out of sync but I do not have a better proposal.

packages/binding-modbus/README.md
```diff
@@ -3,62 +3,98 @@
...
+// Protocols and Servient
+Servient = require("@node-wot/core").Servient;
+ModbusClientFactory = require("@node-wot/binding-modbus").ModbusClientFactory;
+
+// create Servient and add Modbus binding
+let servient = new Servient();
+servient.addClientFactory(new ModbusClientFactory());
+
+async function main() {
+    let td = {}; // see https://github.com/eclipse-thingweb/test-things/blob/main/things/elevator/modbus/js/modbus-elevator.td.json
```
danielpeintner, line 25: Why don't you want to include the TD ?

packages/binding-modbus/README.md
egekorkan: My long-term vision is to have those test things very stable with respect to the binding specs and even to use them for some end-to-end tests. For me, it is similar to pointing to to hosted things we have such as we do at https://github.com/eclipse-thingweb/node-wot/tree/master/packages/binding-http#client-example

packages/binding-modbus/src/modbus.ts
```diff
@@ -42,45 +43,81 @@ export type ModbusFunctionName =
...
+    | "xsd:float"
+    | "xsd:decimal"
+    | "xsd:byte"
+    | "xsd:short"
+    | "xsd:int"
+    | "xsd:long"
+    | "xsd:unsignedbyte"
+    | "xsd:unsignedshort"
+    | "xsd:unsignedint"
+    | "xsd:unsignedlong"
```
danielpeintner, line 121: I think they MUST be capitalized
"xsd:unsignedbyte" --> "xsd:unsignedByte" etc

danielpeintner (commented):
Find some comments below

packages/binding-modbus/README.md
egekorkan: It is changing at the moment due to the changes in this PR. Once this PR is merged and that TD is updated, I can add it inline. Another alternative would be to add a fetch method but that would imply adding the HTTP binding

packages/binding-modbus/src/modbus.ts
egekorkan: In that case, we have a mistake in the binding spec as well 😱 but you are correct:
- Binding: https://w3c.github.io/wot-binding-templates/bindings/protocols/modbus/#payloaddatatype
- XML Schema: https://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#unsignedByte

egekorkan: Done here and in the binding spec: https://github.com/w3c/wot-binding-templates/pull/359

Resolving

danielpeintner (approved):

packages/binding-modbus/src/modbus-client.ts
egekorkan: This is now done. Can we resolve this thread and merge the PR?

relu91 (approved):
Good to go! thank you Ege!!

PR #1266, opened 2024-04-15
refactor: avoid adding schemas to AJV

danielpeintner:
to avoid memory leak

tackles the first easy step of https://github.com/eclipse-thingweb/node-wot/issues/1062


egekorkan:
This will decrease performance but avoid the crashes that happen in the online test things. Agreed to merge in the committer call.

relu91 (approved):

Issue #1268, opened 2024-04-22
allowSelfSigned not working for requestThingDescription


danielpeintner:
Not sure where to inquire for this: 

HttpConfig for https client seems to ignore allowSelfSigned:

```javascript
const { HttpsClientFactory } = require("@node-wot/binding-http");


const servient = new Servient();
servient.addClientFactory(new HttpsClientFactory({ 
        allowSelfSigned : true,
    }));

servient.start().then(async (WoT) => {
    const td = await WoT.requestThingDescription("https://localhost:8083/spectrometer/ocean-optics/USB2000-plus/resources/wot");
```

I get error:

FetchError: request to https://localhost:8083/spectrometer/ocean-optics/USB2000-plus/resources/wot failed, reason: self-signed certificate
    at ClientRequest.<anonymous> (H:\New Work\node-clients\serial-utility\node_modules\node-fetch\lib\index.js:1501:11)
    at ClientRequest.emit (node:events:514:28)
    at TLSSocket.socketErrorListener (node:_http_client:495:9)
    at TLSSocket.emit (node:events:514:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  type: 'system',
  errno: 'DEPTH_ZERO_SELF_SIGNED_CERT',
  code: 'DEPTH_ZERO_SELF_SIGNED_CERT'
}

_Originally posted by @VigneshVSV in https://github.com/eclipse-thingweb/node-wot/issues/177#issuecomment-2067688741_

danielpeintner:
I looked into your problem and noticed that `allowSelfSigned` is not taken into account for `WoT.requestThingDescription(...)`. Instead it kicks in for interaction methods like `thing.readProperty(...)` only

danielpeintner:
I created https://github.com/eclipse-thingweb/node-wot/pull/1269 which should fix the issue... I tested locally. if you can test/check also that would be great. Thanks!

VigneshVSV:
I pulled the latest master branch and built the wot-bundle-min browser add-on. 

I still get the same error.


```javascript
// create servient - servients have both clients and servers - we are interested in the client
const servient = new Wot.Core.Servient();
const clientFactory = new Wot.Http.HttpsClientFactory({ allowSelfSigned : true });
servient.addClientFactory(clientFactory)
[...]

PR #1269, opened 2024-04-22
refactor: use the same agent etc for requestThingDescription as with …

danielpeintner:
…any other interaction

fixes https://github.com/eclipse-thingweb/node-wot/issues/1268


egekorkan:
Looks good to me, not sure if it would break something though

danielpeintner:
> Looks good to me, not sure if it would break something though

I don't think it should break anything since at the moment the access for TDs works only for publicly accessible resources.

relu91 (approved):
Yes, that was definitely an oversight. Thanks!

Issue #854, opened 2022-10-22
Handle form `content-type` client side


relu91:
Hi all. I'm opening this issue as a consequence of the latest discussion during the TD/Protocol binding templates call. During the call, people complained about how node-wot clients (primarily HTTP and COAP) are ignoring the `content-type` field in forms. Sadly I couldn't find the minutes of the call so I'm trying to summarize by heart. 

Looking at https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-http/src/http-client-impl.ts, what I see, is that the content type is taken into account for every write operation (i.e., invokeaction and writeproperty), whereas is ignored in reading operations (node-wot uses what the server gives). Do people feel like that node-wot HTTP client should set the `Accept` header to the  `content-type` value when reading? To me is quite redundant but it does not hurt that much. Do you have a compelling reason for doing so? 

I know that the COAP discussion is more controversial due to mappings between coap-specific fields and WoT  `content-type`, so I would leave it after we have a common take on HTTP. 

pinging relevant people: @egekorkan @mkovatsc @danielpeintner @hidetak @JKRhb. If I recalled something wrong feel free to correct me, I'll update this first post accordingly.

egekorkan:
I'd say that it should set accept header since it can be that there are two identical forms, only differing in contentType.

mkovatsc:
The purpose of the `contentType` field (and `contentCoding` where applicable) is to inform the upper layers what format comes out / goes into this particular interaction using this form. `op` is used to match a form to the interaction intended by the application. All other fields of the form have the purpose to describe how to generate a request that results in exactly the behavior the application expects and the Thing is able to understand, without any negotiations left open -- within a form, all negotiations are decided at design time.

There still is the concept of content negotiation, as it is part of the Web Architecture. It is not happening at protocol binding level, however, yet at the form level: when a Thing provides N different formats, the TD author can include N different forms, each with a different  `contentType`, so that WoT Servients can pick a format for which they have a handler implemented. (This is why the field must not be moved up to the affordance level!) Content format handlers are between bindings and affordances, so they can be used across all bindings. Content format strings are WoT's abstraction mechanism to deal with any data format on the wire; some protocols already use them, for others we need to define them like the URI schemes, both resulting in uniform interfaces for these protocols (modulo the ever-missing HATEOAS part...).

[...]

PR #1283, opened 2024-05-09
Support for default accept header in http client

relu91:
As discussed in #854  this PR adds the default assumptions for the Accept header when interacting with a remote Thing. It also assumes that if the user provides the headers field his choice has preference over the default assumptions.

### DIFF
--- a/packages/binding-http/src/http-client-impl.ts
+++ b/packages/binding-http/src/http-client-impl.ts
@@ -112,7 +112,10 @@ export default class HttpClient implements ProtocolClient {
     }
 
     public async readResource(form: HttpForm): Promise<Content> {
-        const request = await this.generateFetchRequest(form, "GET");
+        // See https://www.w3.org/TR/wot-thing-description11/#contentType-usage
+        // Cases: 1B
+        const headers = form.contentType != null ? [["accept", form.contentType]] : [["accept", ContentSerdes.DEFAULT]];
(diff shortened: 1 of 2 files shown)


egekorkan:
This Thing has behavior implementation to test this:https://github.com/eclipse-thingweb/test-things/tree/main/things/calculator/http/express
 Sadly issues with deployment on thingweb server at the moment.

packages/binding-http/src/http-client-impl.ts
```diff
@@ -112,7 +112,10 @@ export default class HttpClient implements ProtocolClient {
     }
 
     public async readResource(form: HttpForm): Promise<Content> {
-        const request = await this.generateFetchRequest(form, "GET");
+        // See https://www.w3.org/TR/wot-thing-description11/#contentType-usage
+        // Cases: 1B
```
danielpeintner, line 116:
```suggestion
        // Case: 1B
```

danielpeintner (approved):
LGTM.

One minor suggestion

relu91:
Given @egekorkan's comment, I think we should merge this after someone has tried to interact with the mentioned Thing. I don't have time today so if anybody else wants to test these changes please feel free to report it here.

egekorkan (approved):
I have just tested and it works as expected. Thanks!

PR #1285, opened 2024-05-22
fix(coap): wrong error variable name

danielpeintner:
Note: interesting that this is not detected automatically... `error` exists as import also...


JKRhb (approved):

JKRhb:
Thank you for fixing this :) Really a bit strange that this issue has not been uncovered by the checks we have in place :/

relu91:
I think we don't have explicit checks for unused variables in catch statements? 🤔 .

Issue #1294, opened 2024-06-13
observe property deserialized value (may) not (be) completely consistent


VigneshVSV:
The observable change events are pushed through HTTP SSE in may case. The following is the property affordance:

```json
"position": {
      "title": "position",
      "description": "current position, write operation does not cause movement but overwrite register value.",
      "type": "number",
      "forms": [
        {
          "href": "https://localhost:8080/linear-stage/position",
          "op": "readproperty",
          "htv:methodName": "GET",
          "contentType": "application/json"
        },
        {
          "href": "https://localhost:8080/linear-stage/position",
          "op": "writeproperty",
          "htv:methodName": "PUT",
          "contentType": "application/json"
        },
        {
          "href": "https://localhost:8080/linear-stage/position/change-event",
          "subprotocol": "sse",
          "op": "observeproperty",
          "htv:methodName": "GET",
          "contentType": "text/plain"
        }
      ],
      "observable": true
    }
```
Either using `text/plain` or `application/json` as `contentType` , one still only retrieves the entire event object, not the `event.data`

Web browser gives the following output as expected from the event-stream. Server set header is `text/event-stream`:

```
data: 20.36

data: 20.005

data: 19.585

data: 19.07

data: 18.4825

data: 17.83

data: 17.0825

```

But what I get from node-wot (please ignore the stringification) 

```console
movements.tsx:24 Position changed to:
movements.tsx:25 "{\"type\":\"message\",\"data\":\"17.1025\",\"lastEventId\":\"\",\"origin\":\"https://localhost:8080\"}"
movements.tsx:24 Position changed to:
movements.tsx:25 "{\"type\":\"message\",\"data\":\"17.70375\",\"lastEventId\":\"\",\"origin\":\"https://localhost:8080\"}"
movements.tsx:24 Position changed to:
movements.tsx:25 "{\"type\":\"message\",\"data\":\"18.36\",\"lastEventId\":\"\",\"origin\":\"https://localhost:8080\"}"
movements.tsx:24 Position changed to:
movements.tsx:25 "{\"type\":\"message\",\"data\":\"18.96375\",\"lastEventId\":\"\",\"origin\":\"https://localhost:8080\"}"
movements.tsx:24 Position changed to:
movements.tsx:25 "{\"type\":\"message\",\"data\":\"19.48375\",\"lastEventId\":\"\",\"origin\":\"https://localhost:8080\"}"
movements.tsx:24 Position changed to:
[...]

PR #1298, opened 2024-07-02
fix: SSE data reporting

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1294

### DIFF
--- a/packages/binding-http/src/subscription-protocols.ts
+++ b/packages/binding-http/src/subscription-protocols.ts
@@ -117,10 +117,7 @@ export class SSESubscription implements InternalSubscription {
             };
             this.eventSource.onmessage = (event) => {
                 debug(`HttpClient received ${JSON.stringify(event)} from ${this.form.href}`);
-                const output = new Content(
-                    this.form.contentType ?? ContentSerdes.DEFAULT,
-                    Readable.from(JSON.stringify(event))
-                );
+                const output = new Content(this.form.contentType ?? ContentSerdes.DEFAULT, Readable.from(event.data));
                 next(output);
             };
             this.eventSource.onerror = function (event) {
--- a/packages/binding-http/test/http-client-test.ts
+++ b/packages/binding-http/test/http-client-test.ts
@@ -25,6 +25,7 @@ import * as http from "http";
 import { Content, DefaultContent, ContentSerdes, createLoggers, ProtocolServer } from "@node-wot/core";
 
 import { Readable } from "stream";
+import { text } from "node:stream/consumers";
 
 import HttpClient from "../src/http-client";
 import { HttpForm } from "../src/http";
@@ -463,7 +464,9 @@ class HttpClientTest2 {
         };
 
         client
-            .subscribeResource(form, (data) => {
+            .subscribeResource(form, async (data) => {
+                const str = await text(data.body); // SHOULD be "Test event"
+                str.should.eql("Test event");
                 client.unlinkResource(form);
                 server.close();
                 clock.uninstall();
--- a/packages/binding-http/test/http-client-test.ts
+++ b/packages/binding-http/test/http-client-test.ts
@@ -38,6 +38,7 @@ import FakeTimers from "@sinonjs/fake-timers";
 
 // Add spies
 import spies from "chai-spies";
+import { fail } from "assert";
 
 const { debug } = createLoggers("binding-http", "http-client-test");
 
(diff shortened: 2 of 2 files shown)


packages/binding-http/src/subscription-protocols.ts
```diff
@@ -117,10 +117,7 @@ export class SSESubscription implements InternalSubscription {
             };
             this.eventSource.onmessage = (event) => {
                 debug(`HttpClient received ${JSON.stringify(event)} from ${this.form.href}`);
-                const output = new Content(
-                    this.form.contentType ?? ContentSerdes.DEFAULT,
-                    Readable.from(JSON.stringify(event))
-                );
+                const output = new Content(this.form.contentType ?? ContentSerdes.DEFAULT, Readable.from(event.data));
```
danielpeintner, line 120: Why was there `JSON.stringify()` around ?

@relu91

packages/binding-http/test/http-client-test.ts
```diff
@@ -463,7 +464,9 @@ class HttpClientTest2 {
         };
 
         client
-            .subscribeResource(form, (data) => {
+            .subscribeResource(form, async (data) => {
+                const str = await text(data.body); // SHOULD be "Test event"
+                str.should.eql("Test event");
```
danielpeintner, line 469: Question: IF I add a _wrong_ test like

`str.should.eql("Test event XXX");`

I get the following error

`Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called;`

How can I fix that?

danielpeintner: Mhh, what you proposed didn't work for me. I played around but didn't manage to succeed.

Anyhow, I found another clumsy code that manages to detect any future error, see https://github.com/eclipse-thingweb/node-wot/pull/1298/commits/2808a7f2ab67315ef55028007bf16f752dcb75c3

egekorkan:
Some comments:
- Here is a Test Thing you can use to test the code: https://github.com/eclipse-thingweb/test-things/blob/main/things/calculator/http/express/http-simple-calculator-thing.td.jsonld the code is in there as well. There are client tests with node-wot in the same folder but they are commented out until this is resolved
- There are some issue threads in the standards: https://github.com/w3c/wot-binding-templates/issues/330 , https://github.com/w3c/wot-binding-templates/issues/342 , https://github.com/w3c/wot-binding-templates/issues/347

packages/binding-http/test/http-client-test.ts
relu91: I'll give it a try to fix it tomorrow if I don't suceed let's go with your solution.

danielpeintner: I know, the code is not very "nice" but it does the job 🤷‍♂️🙈

relu91: Ok I dag deeper into the issue, I suggest to remove the code added in 2808a7f and update npm run test adding this env variable: NODE_OPTIONS=--unhandled-rejections=strict. As explained in https://github.com/mochajs/mocha/issues/2640#issuecomment-1663388547. It seems that mocha does ignore the new nodejs mechanism of reporting async unhandled exceptions.

danielpeintner: I can confirm that setting the env varaible works. I also _quickly_ checked how this can be done in a cross environment manner (`set NODE_OPTIONS=bla` on Windows vs. `export NODE_OPTIONS=bla` on Linux/Mac). There is no really good solution I think.

There is also [cross-env package](https://www.npmjs.com/package/cross-env) which is in maintenance mode.

Anyhow, all this makes me think this is becoming overly complex and not worth it for this problem.

Any thoughts and/or ideas?

egekorkan:
> > Given that 1 is always the case for sse and we can understand that from subprotocol, it is fine to put json in the form content type.
> 
> This is something that (I think) goes beyond this PR. I suggest to open another issue/PR for that.
> 
> @egekorkan w.r.t. the other comments I am not sure whether they affect "this" PR. There is a test added that checks the consistency of the change. Do you want me to change the PR further?

I think I got confused after some discussions with @relu91 and commented here. My comments do not affect this PR.

packages/binding-http/test/http-client-test.ts
relu91: mocha relays node options to the node runner so you add the option to the command easily:
`mocha --unhandled-rejections=strict --require ts-node/register --extension ts`

relu91: a bit weird cause I did not spot those errors when testing it (only if the SSE test was forcefully caused to fail). But I'm ok to fix it later in another issue.

relu91 (approved):

packages/binding-http/test/http-client-test.ts
danielpeintner: I created https://github.com/eclipse-thingweb/node-wot/issues/1299

VigneshVSV:
works

Issue #1304, opened 2024-07-12
two netconf-codec.ts files


danielpeintner:
> By the way: I just fully realized only now that there are two `netconf-codec.ts` files, one in the `core` package and one in the `binding-netconf` package. I wonder if we should remove one of them (probably the one in the `core` package)?

_Originally posted by @JKRhb in https://github.com/eclipse-thingweb/node-wot/pull/1210#discussion_r1673472530_

File in `core` package should be removed

PR #1305, opened 2024-07-12
refactor: remove obsolete netconf codec from core package

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1304


JKRhb (approved):

danielpeintner:
I think it is simple enough to go ahead since @JKRhb approved also

Issue #1301, opened 2024-07-09
TODOs after major release


danielpeintner:
This is an _umbrella_ issue meant to collect tasks which should be done after our next major release. In general it is meant to clean-up stuff getting rid of old stuff that has been collected over the years.

* [ ] remove td-tools package (see https://github.com/eclipse-thingweb/td-tools/issues/31)
  * tm-helpers, see https://github.com/eclipse-thingweb/node-wot/issues/1296)
  * td-parser.ts
  * resolver-interface.ts
  * td-helpers.ts
* [ ] ...

danielpeintner:
I think we can remove the out-dated code. 
I am wondering whether we can remove it in 0.9.x (new minor version) or do we need to wait for a major version (e.g., 1.0.0) ?

PR #1308, opened 2024-07-21
refactor: remove unused findProtocol and findPort functions

### LINKED ISSUE #1307: Replacing Protocol and Port Detection
With https://www.npmjs.com/package/@thingweb/td-utils published, we can use it here and replace the old code at td-helpers with this.

hasanheroglu:
- Removed unused findProtocol and findPort functions
- Closes #1307

### DIFF
--- a/packages/td-tools/src/td-helpers.ts
+++ b/packages/td-tools/src/td-helpers.ts
@@ -23,7 +23,6 @@
 
 import ThingDescription from "./thing-description";
 
-
 /**
  * @deprecated This function will be removed in the future. Please use '@thingweb/td-utils' package instead.
  */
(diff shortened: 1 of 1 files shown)


danielpeintner:
Those functions are exported and we should NOT remove them. The idea is to remove them once we upgrade the version. 
see https://github.com/eclipse-thingweb/node-wot/issues/1301

Hence I propose to deprecate them again.

Note: I am wondering why the CI fail. Are those functions still used? I will re-run the pipeline.. maybe there is another issue...

danielpeintner (approved):

relu91 (approved):

Issue #1243, opened 2024-02-16
InteractionOutput - no schema/schema.type and the expectations of the value() function


danielpeintner:
PR https://github.com/eclipse-thingweb/node-wot/pull/1230 reports/fixes/adjusts the case for `schema == null` (no expected value) for `value()` function reporting `undefined` .

Anyhow, there is a much broader issue:
* what is a valid schema?
* `schema.type == null` since we use `oneof`
* optional return value
* ...

Relates also to:
* https://github.com/w3c/wot-scripting-api/pull/534
* https://github.com/w3c/wot-thing-description/issues/1441
* https://github.com/w3c/wot-thing-description/issues/1234

egekorkan:
> what is a valid schema?

That is a big question, maybe, but anything that validates the JSON Schema meta schema is valid: https://github.com/json-schema-org/json-schema-spec/blob/draft-handrews-json-schema-validation-00/schema.json

My bigger question is why we are looking inside a data schema, it should be just "passed over". If it is validation, the whole property is a schema, with action input or output etc.

relu91:
> My bigger question is why we are looking inside a data schema, it should be just "passed over". If it is validation, the whole property is a schema, with action input or output etc.

Because we are not using the schema just to validate the payload but also to guide the *decoding* of the payload. Basically, having the schema allows us to convert non-JSON content types to Javascript objects that can be later validated with the schema. 

> what is a valid schema?

Given what is said above I think that for us a valid schema (to be used in the `value` function) can be defined as (more or less):

```js
function validSchema(schema) {
   return schema.type != null || schema.const != null || schema.enum != null || schema.oneOf?.reduce((acc, val)=> acc || validSchema(val), true)
}
```

with `const` or `enum` we can infer the type (number, string, object, or array) and the `oneOf` is just recursive.  I should have included all the possibilities looking at https://www.w3.org/TR/wot-thing-description11/#dataschema . Let me know if I missed something. 

P.s. I'd probably prefer that `const` and `enum` should always explicitly state the type but it could be just a warning.

VigneshVSV:
Hi, I have following use case:

1)  nullable properties of fixed type - 

For example, I have the following property definition
```json

"state": {
      "title": "state",
      "description": "current state machine state if state machine present",
      "const": false,
      "readOnly": true,
      "writeOnly": false,
      "type": "string",
      "oneOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
[...]

PR #1309, opened 2024-07-22
fix(core/interactionOutput): support for no-type based root schemas

relu91:
As discussed in #1243 and solved in https://github.com/w3c/wot-scripting-api/pull/534, we now allow more root schema types to be accepted in the value function. The validation still happens but it supports: `const`, `enum`, and `oneOf`. Those terms can now be used instead of `type`. The behavior of using these terms **at the same time** is delegated to Ajv.validate function.

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -104,7 +104,12 @@ export class InteractionOutput implements WoT.InteractionOutput {
             throw new NotReadableError("No form defined");
         }
 
-        if (this.schema.type == null) {
+        if (
+            this.schema.const == null &&
+            this.schema.enum == null &&
+            this.schema.oneOf == null &&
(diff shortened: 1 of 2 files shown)


danielpeintner (approved):
Looks good.

Question: How can we proceed with the _other_ [schema combinations](https://json-schema.org/understanding-json-schema/reference/combining) like `allOf`, `anyOf` and `not` ?

relu91:
> Question: How can we proceed with the other [schema combinations](https://json-schema.org/understanding-json-schema/reference/combining) like allOf, anyOf and not ?

I had the same question, but I preferred to stick with what was in the Scripting API spec. I think we can easily fix it there and then add them in node-wot later.

relu91:
given @danielpeintner 's approval, I'm merging this.

PR #1310, opened 2024-07-23
docs: add deprecation message to td-tools

danielpeintner:
What about https://github.com/eclipse-thingweb/node-wot/blob/master/packages/td-tools/src/thing-description.ts

It will continue to live in `core`, right?

### DIFF
--- a/packages/td-tools/src/td-helpers.ts
+++ b/packages/td-tools/src/td-helpers.ts
@@ -43,6 +43,9 @@ export function findPort(td: ThingDescription): number {
     return parseInt(returnString);
 }
 
+/**
+ * @deprecated This function will be removed in the future.
+ */
 export function setContextLanguage(thing: ThingDescription, language: string, forceOverride: boolean): void {
     // forceOverride == false -> set @language if no @language set
     // forceOverride == true  -> set/override @language in any case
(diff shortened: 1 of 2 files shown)


relu91 (approved):
> What about https://github.com/eclipse-thingweb/node-wot/blob/master/packages/td-tools/src/thing-description.ts
>
>It will continue to live in core, right?

Yes, but I would deprecate it and move it (or refactor) in one single PR.

danielpeintner:
> Yes, but I would deprecate it and move it (or refactor) in one single PR.

Shall I deprecate each export like

https://github.com/eclipse-thingweb/node-wot/blob/15c5af795dfb4ed63b003bf9084c6882b57e8b8c/packages/td-tools/src/thing-description.ts#L26

https://github.com/eclipse-thingweb/node-wot/blob/15c5af795dfb4ed63b003bf9084c6882b57e8b8c/packages/td-tools/src/thing-description.ts#L62

etc in this PR. IF we do so we get warnings everywhere.

OR I would do it once we completely remove the entire `td-tools` package. That's what I would do.
I hope that matches with you...

relu91:
What I meant was to move on with this PR and in other PR tackle the refactoring of duplicating thing-descriptio.ts to core package and deprecate the code in td-tools package. If we follow this approach we should not get any warning (once all the imports will be updated to use core instead of td-tools).

Issue #1311, opened 2024-07-23
PR to tackle the refactoring of duplicating thing-description.ts to core


danielpeintner:
>  What I meant was to move on with this PR and in other PR tackle the refactoring of duplicating thing-descriptio.ts to core package and deprecate the code in td-tools package. If we follow this approach we should not get any warning (once all the imports will be updated to use core instead of td-tools).

_Originally posted by @relu91 in https://github.com/eclipse-thingweb/node-wot/issues/1310#issuecomment-2245480678_

PR #1312, opened 2024-07-24
Remove dependencies from td-tool package 

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1311

### DIFF
--- a/packages/core/src/thing-description.ts
+++ b/packages/core/src/thing-description.ts
@@ -1,0 +1,251 @@
+/********************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the W3C Software Notice and
+ * Document License (2015-05-13) which is available at
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+
+// global W3C WoT Scripting API definitions
+import * as WoT from "wot-typescript-definitions";
+import * as TDT from "wot-thing-description-types";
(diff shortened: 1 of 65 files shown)


danielpeintner:
The PR is now ready for review. With the changes we can be removing the `td-tools` package one needed (there is no dependency anymore).

danielpeintner:
Anyone willing to review?

There is no *real* technical change.... it moves  _necessary_ functionality from `td-tools` to `core` and removes in all packages the dependency to `td-tools`

egekorkan (approved):
Looks fine and the changes seem mostly mechanical and copy-paste

packages/core/src/thing-description.ts
```diff
@@ -0,0 +1,258 @@
...
+export const DEFAULT_CONTEXT_V11 = "https://www.w3.org/2022/wot/td/v1.1";
+export const DEFAULT_CONTEXT_LANGUAGE = "en";
+export const DEFAULT_THING_TYPE = "Thing";
+
+type DeepPartial<T> = T extends Record<string, unknown>
+    ? {
+          [P in keyof T]?: T[P] extends Array<infer I> ? Array<DeepPartial<I>> : DeepPartial<T[P]>;
+      }
+    : T;
+export type ThingModel = DeepPartial<Thing>;
```
relu91, line 30: Spotted a possible collision with the concept of `ThingModel` from `@thingweb/thing-model`. Is it the same concept? or something else?

relu91 (approved):
Other than the question below the changes are ok!

packages/core/src/thing-description.ts
danielpeintner: Mhh, what you mean by that. I took it over from

https://github.com/eclipse-thingweb/node-wot/blob/274bf4aada923429e9164cfe8a82cd387ac6e6ea/packages/td-tools/src/td-tools.ts#L23-L28

Do you think this should not be needed?

danielpeintner: With https://github.com/eclipse-thingweb/node-wot/pull/1312/commits/8c1e0598f4c45cb42e60b292013db3cf2e31516a I removed our own definition of `ThingModel`. Instead I used the one from "wot-thing-model-types"

I think this should resolve your concerns...

PR #1315, opened 2024-08-12
feat(binding-coap): remove cov:observe subprotocol

JKRhb:
This PR incorporates the recent changes applied to the CoAP Binding Template document in https://github.com/w3c/wot-binding-templates/pull/353, removing the `cov:observe` subprotocol from the TDs that are generated when adding a CoAP server to a servient.

In order to better cover the consumer behavior described in the CoAP binding template (falling back to polling if the server does not support CoAP observe), some more work is needed that might also need to be applied to the underlying CoAP libraries. I will try to cover these aspects soon in a follow-up PR.

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -35,7 +35,7 @@ import { MdnsIntroducer } from "./mdns-introducer";
 import { PropertyElement, DataSchema, ActionElement, EventElement } from "wot-thing-description-types";
 import { CoapServerConfig } from "./coap";
 import { DataSchemaValue } from "wot-typescript-definitions";
-import { filterPropertyObserveOperations, getPropertyOpValues } from "./util";
+import { getPropertyOpValues } from "./util";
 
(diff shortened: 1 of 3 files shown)


danielpeintner (approved):
Looks good and consistent to me 👍

danielpeintner:
> some more work is needed that might also need to be applied to the underlying CoAP libraries. I will try to cover these aspects soon in a follow-up PR.

Do you want to list/collect the future work in an issue?

JKRhb:
Thank you for the feedback, @danielpeintner :)

> > some more work is needed that might also need to be applied to the underlying CoAP libraries. I will try to cover these aspects soon in a follow-up PR.
> 
> Do you want to list/collect the future work in an issue?

I've opened https://github.com/eclipse-thingweb/node-wot/issues/1316 to keep track of the additions needed to support the new behavior :)

JKRhb:
Since the PR got approved and #1316 is created, I would go ahead and merge this PR :)

Issue #1317, opened 2024-08-20
core package is missing dependencies in v0.8.15


hasanheroglu:
[serdes.ts](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/serdes.ts) uses two dependencies `is-absolute-url` and `url-toolkit` which are not included in core's [package.json](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/package.json). Since td-tools package installs those two dependencies it may seem working in the monorepo but when it is installed alone it causes an error.

danielpeintner:
Mhh, the problem was surely introduced in https://github.com/eclipse-thingweb/node-wot/pull/1312 and I will provide a fix soon.

I wonder whether there is a way to detect such an issue ... any ideas ?

hasanheroglu:
We can separate CI runs for each package. We install only package-specific and root dependencies (devDependencies of the root package.json), run the test, and see if it passes.

relu91:
> We can separate CI runs for each package. We install only package-specific and root dependencies (devDependencies of the root package.json), run the test, and see if it passes. I can work on this if it makes sense for you.

I'm curious to see a concrete PR on this, but I'm afraid it will get quite complicated to maintain and defeat the advantages of having a mono repo.

danielpeintner:
FYI, v0.8.16 has been published.

PR #1318, opened 2024-08-21
fix: add missing packages for serdes.ts

danielpeintner:
Note: I used the exact same version for `is-absolute-url` and `url-toolkit` as in https://github.com/eclipse-thingweb/node-wot/blob/f63b72951fa9190ced8126e2f6b4cfe045b148bf/packages/td-tools/package.json#L20-L22 to avoid issues.

fixes https://github.com/eclipse-thingweb/node-wot/issues/1317

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12290,7 +12290,9 @@
                 "cbor": "^8.1.0",
                 "content-type": "^1.0.5",
                 "debug": "^4.3.4",
+                "is-absolute-url": "3.0.3",
                 "uritemplate": "0.3.4",
+                "url-toolkit": "2.1.6",
                 "uuid": "^7.0.3",
                 "web-streams-polyfill": "^3.0.1"
             },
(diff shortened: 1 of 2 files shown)


danielpeintner:
I also wonder if this justifies a quick new release, because the library is not really usable without the fix, right ...

@hasanheroglu ?

JKRhb (approved):

JKRhb:
> I also wonder if this justifies a quick new release, because the library is not really usable without the fix, right ...

I think that would make sense :)

hasanheroglu:
> I also wonder if this justifies a quick new release, because the library is not really usable without the fix, right ...
> 
> @hasanheroglu ?

As a workaround, one can install the missing dependencies separately, but yes the library is broken.

danielpeintner:
@relu91 are you fine with releasing an update ASAP ?

relu91 (approved):
Definitely ok for releasing

PR #1321, opened 2024-08-22
detect phantom dependencies and fix

### LINKED ISSUE #1319: Detect issue with dependencies in monorepo
> I wonder whether there is a way to detect such an issue ... any ideas ?

We can separate CI runs for each package. We install only package-specific and root dependencies (devDependencies of the root package.json), run the test, and see if it passes. I can work on this if it makes sense for you.

_Originally posted by @hasanheroglu in https://github.com/eclipse-thingweb/node-wot/issues/1317#issuecomment-2301483122_

hasanheroglu:
Closes #1319.

As @relu91 mentioned, separating CI runs gets complicated and doesn't match the logic of a monorepo. I found a better and simpler solution: adding `import/no-extraneous-dependencies` to eslint configuration.

This lint config deals with "phantom dependencies" and gives an error if they exist.


JKRhb:
Thank you, @hasanheroglu, that looks very promising :) I am wondering, though: Does this PR make the comon devDependencies in the top-level `package.json` obsolete? If so, then that section could probably be removed.

Consdering the conflicts, you may need to rebase your branch against `master` to resolve them.

hasanheroglu:
> Thank you, @hasanheroglu, that looks very promising :) I am wondering, though: Does this PR make the comon devDependencies in the top-level `package.json` obsolete? If so, then that section could probably be removed.

I just realized I didn't add devDependencies for types in the packages. I can add them and then remove the obsolete dependencies from the root.

relu91:
Great, that's an ingenuous move 👍🏻 well done.  The only problem that I see is that having dev deps in the root package.json help with the consistency of the dev tools among all packages... however, I always feel that is not a big deal since every package could in theory be developed differently than the others (with different dev tools). @danielpeintner what do you think?

hasanheroglu:
> Great, that's an ingenuous move 👍🏻 well done. The only problem that I see is that having dev deps in the root package.json help with the consistency of the dev tools among all packages... however, I always feel that is not a big deal since every package could in theory be developed differently than the others (with different dev tools). @danielpeintner what do you think?

Actually, we can configure the lint rule to apply only to product dependencies. Therefore, we don't need to change any devDependencies, and it shouldn't cause any issues for the package users too.

danielpeintner:
> however, I always feel that is not a big deal since every package could in theory be developed differently than the others (with different dev tools). @danielpeintner what do you think?

I still think it would be better to keep the versions consistent. Apart from the `package.json` in root I think the following script should check the consistency also, right?

https://github.com/eclipse-thingweb/node-wot/blob/99b83034c7949f394753faf5be2252b127d01834/package.json#L31

We do run it also on the CI workflow.

https://github.com/eclipse-thingweb/node-wot/blob/99b83034c7949f394753faf5be2252b127d01834/.github/workflows/ci.yaml#L90

This makes me think that we do not really need `package.json` in the root, right?

relu91:
> This makes me think that we do not really need package.json in the root, right?

It is still required for generic monorepo configuration and lifecycle scripts, but where to list dev dependencies is completely up to us.

relu91 (approved):
Anyhow I forgot to approve the PR.

hasanheroglu:
So I added eslint configs, that extend the root config override "no-extraneous-dependencies" rule, to test directories. Therefore, the rule only applies to the product dependencies and we can share common devDependencies in the root package.json.

danielpeintner (approved):

relu91:
thanks!!!

PR #1327, opened 2024-10-18
chore: Simplify the workspaces list

lu-zero:
Incidentally unbreaks `deno install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "workspaces": [
         "./packages/td-tools",
         "./packages/core",
-        "./packages/binding-!()",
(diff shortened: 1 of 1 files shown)


danielpeintner:
The change looks fine but the Eclipse ECA check fails...

danielpeintner:
@lu-zero can you look into the ECA check failure...

If it is too cumbersome, and you don't mind, someone on our side could provide a _similar_ PR...

lu-zero:
It should be fine now, sorry for the delay.

relu91 (approved):

PR #1329, opened 2024-11-06
Deleting moved test things

egekorkan:
Now these are in test-things repo. Can be deleted here. I will fix the documentation in test-things before making it ready.


danielpeintner:
Note: with this PR we need to update https://github.com/eclipse-thingweb/node-wot#online-things also...

Issue #1331, opened 2024-11-08
WoT consume returns an implementation instead of an interface


RobWin:
Hi,
is this correct in the WoT impl ?

The `WoT.produce` function returns an interface `WoT.ExposedThing? 
https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/wot-impl.ts#L125

But `WoT.consume` function returns an implementation class `ConsumedThing` instead of the interface `WoT.ConsumedThing`
Is this correct or a bug?

https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/src/wot-impl.ts#L104)

danielpeintner:
I am still setting up my machine after a *forced* reinstall but just looking at the code on GitHub I would say it is fine as is.

An implementation can return an instance/implementation of an interface. `ConsumedThing`  `implements IConsumedThing` (FYI: the `IConsumedThing` is just named like this but is infact ConsumedThing  from `"wot-typescript-definitions"`)

Having said that, it might be better to change

`async consume(td: WoT.ThingDescription): Promise<ConsumedThing>`
to
`async consume(td: WoT.ThingDescription): Promise<WoT.ConsumedThing>`

for consistency reasons which should be fine also.. I will try after my toolchain is in place again.

RobWin:
From a user perspective, if I retrieve a ConsumedThing implementation I have access to methods/functions which only the runtime should have access to.  
As a user I should only have access to the functions which are defined in the Scripting API, or not?

RobWin:
Ok, now i understand.
I missed the line
`import { ConsumedThing as IConsumedThing, InteractionInput, Subscription } from "wot-typescript-definitions";`

But how does this code know it should return the interface and not the implementation? 

`const newThing: ConsumedThing = new ConsumedThing(this.srv, thing as ThingModel);`

Sry, I'm not an expert Typescript developer :(

danielpeintner:
I will provide a PR that uses the interface (to be consistent).

danielpeintner:
@RobWin I think https://github.com/eclipse-thingweb/node-wot/pull/1332/files should be what you expect to see, right?

RobWin:
Yes, perfect

PR #1332, opened 2024-11-08
refactor: Use interface for WoT.consume()

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1331


relu91 (approved):
Technically this is a change of the interface... therefore we should publish a minor next. The changes are fine, I appreciate the fact that we are now consistent. Weird that the CI broke even if you didn't introduce any new dep.

Issue #1334, opened 2024-11-28
Expanded Node ID for OPCUA


danielpeintner:
During a PlugFest we discovered that the opcua binding did not manage to properly resolve _expanded_ node ids

We tried to use

`"opcua:nodeId": "nsu=xxxxxx;s=\"LED\".\"State\""`

(Note: xxxx was the actual namespace)

instead of something like this

`"opcua:nodeId": "ns=3;s=\"LED\".\"State\""`

Anyhow, it did not work, and we were not exactly sure whether it is not properly implemented or we are doing something wrong.

@erossignon maybe you have some idea?

erossignon:
I can take this on board. The feature exists in node-opcua and need to be called appropriatly in the binding.

danielpeintner:
Resolved by https://github.com/eclipse-thingweb/node-wot/pull/1335

PR #1335, opened 2024-11-29
opcua: nsu in nodeId (fixing #1334)

erossignon:


sebastiankb:
@erossignon many thanks! 

beside of the "ns=..." nodeId, TD should needs also to allow "nsu=..." as node ID. Still we need to clearify, if the nodeID can be part of the href or is dedicated ua:nodeId term is needed.

danielpeintner:
The idea is that we use `href` instead of the own term `opcua:nodeId` (we use `href` in other bindings too).

For example, we now have

```JSON
{
    "base": "opc.tcp://192.168.120.237:4840/",
    "properties": {
        "foo": {
            "forms": [
                {
                    "opcua:nodeId": "nsu=http://example.org/SpecialNamespace/;s=\"LED\".\"State\""
                }
            ]
        }
    }
}
```

which would become the following

```JSON
{
    "base": "opc.tcp://192.168.120.237:4840/",
    "properties": {
        "foo": {
            "forms": [
                {
                    "href": "nsu=http://example.org/SpecialNamespace/;s=\"LED\".\"State\""
                }
            ]
        }
    }
}
```

The problem I see are href parsers since the absolute URL would become the following

`opc.tcp://192.168.120.237:4840/nsu=http://example.org/SpecialNamespace/;s=\"LED\".\"State\"`

A more Web style friendly format would be to use query parameters

`opc.tcp://192.168.120.237:4840/?nsu=http://example.org/SpecialNamespace/&s=\"LED\".\"State\"`

Note1: the difference is the `?` after the host part and that arguments get concatenated via `&` and not `;`

Note2: it has the downside that one needs to reconstruct the format string at OPC UA side.

Question: I wonder whether people share the concern that if we purely use the OPC UA format string href parsers will _destroy/corrupt_  the information.

Note3: we also need to be careful with some specific characters like `#` that get interpreted as fragment identifier. That could be the case in the `nsu` part. Try to parse the following string
`opc.tcp://192.168.120.237/nsu=http://example.org/#foo;s=SpecialVar`
on https://www.freeformatter.com/url-parser-query-string-splitter.html

egekorkan:
Going with URL encoding was also the preliminary decision taken for MQTT at https://github.com/w3c/wot-binding-templates/issues/292#issuecomment-1670848467

danielpeintner:
> Going with URL encoding was also the preliminary decision taken for MQTT at [w3c/wot-binding-templates#292 (comment)](https://github.com/w3c/wot-binding-templates/issues/292#issuecomment-1670848467)

The problem I see with doing so is that for humans, it is not easy to read the "encoded" string.
(Not sure if this is an actual problem)

Anyhow, It seems we need a general decision that can be taken across all/other bindings...

egekorkan:
We had a chat about this together with @sebastiankb @danielpeintner @wiresio and @Kaz040. Here our points, that need to be discussed with the OPC UA community as well.

- The TDs should be still human-readable: Encoding the entire URI breaks this. This can turn off the ones who are used to OPC UA.
- It should be easy to copy-paste node ids from a tool like UA Expert to TDs 
- We have evaluated some aspects:
  - Using single quotes inside the double quotes: It is valid JSON, but the UA server can still have double quotes that need to be converted when copy-pasted. Surrounding a string with single quotes is not valid JSON.
  - Double quote seems to be the most annoying part since even when we do not use `href` it is a special character for JSON. We are not sure how common it is but in Siemens S7, it is quite common.
  - Escaping the quotes would be tolerable for someone copy-pasting. It can be done by humans easily as opposed to URL-encoding.
  - Best would be to use query parameters if people can get used to it.
  - If we want to do encoding, we can simply encode the part after `nsu`

sebastiankb:
In the OPC UA WG we discussed 2 relevant options to provide the nodeId information:

----------

**Option I:**
The value of `href` can be used with the following patterns:
```
opc.tcp:// <address>:<port>/?nsu=<namespace>&s=urlEncoded(<stringId>) 
                                            &b=urlEncoded(<base64string>) 
                                            &i=<integer> 
                                            &g=<guid>
```
Note: The UA server endpoint information `opc.tcp:// <address>:<port>` can be specified in the `base` instead.

Pro: 
- compliant to RFC3986
- compliant to WoT datapoint addressing assumption
- only for sting-based IDs there is an URL encoding requested

Con:
- nodeID cannot be used `as is` (e.g., `nsu=http://example.org/SpecialNamespace/;s=\"LED\".\"State\"`); must be separated into the query parameters  and URL encoding is needed for string-based IDs
 
 **Option II:**

1. The value of `href` is empty or only used for UA server endpoint information (e.g., if `base` is not used).
2. A separate term is defined `ua:nodeID` that takes the UA nodeID as value, e.g.:
 ```
"href":"",
"ua:nodeID":"nsu=http://example.org/SpecialNamespace/;s=\"LED\".\"State\""
```
Pro: 
- compliant to RFC3986
- nodeID cannot be used `as is` (e.g., no URL encoding is needed) 

Con:
- not compliant to WoT datapoint addressing assumption

----------

What do you think?

**Comment by sebastiankb** [comment]
Many thanks for your feedback!

>Why is the " escaped and then encoded again

This example coming from a real sample implementation which uses nested `"` in the string-based nodeID.

> namespace can need url encoding too since it can contain # too

OK, good point. It was hoped that the values of queries would be excluded...

>@sebastiankb why are there multiple query parameters and why not everything is encoded?

Actually, the answer is already given [above](https://github.com/eclipse-thingweb/node-wot/pull/1335#issuecomment-2527964513). 
The idea was to keep the readability of nodeIDs in TDs as high as possible and to avoid the noise of URL encoding as much as possible. But if we also need to encode the namespace, then I see no advantage in defining separate query parameters. In that context, the proposal from @erossignon above makes most sense: 

`opc.tcp:// <address>:<port>/urlEncoded(<nodeID>)`

It seems that a more readable approach is not possible...

**Comment by barnstee** [comment]
> There also seems to be a nice alternative where the UA nodeID remains as is (=readability) and keeps the WoT consistency:
> 
> `opc.tcp:// <address>:<port>/#<nodeID>`
> 
> The fragment identifier # takes any character, an URL encoding is not needed.
> 
> (btw: this approach was also [discussed](https://github.com/w3c/wot-binding-templates/issues/292#issuecomment-1641920726) for MQTT in the past)

This is perfect!

**Comment by relu91** [comment]
> There also seems to be a nice alternative where the UA nodeID remains as is (=readability) and keeps the WoT consistency:
> 
> `opc.tcp:// <address>:<port>/#<nodeID>`
> 
> The fragment identifier # takes any character, an URL encoding is not needed.
> 
> (btw: this approach was also [discussed](https://github.com/w3c/wot-binding-templates/issues/292#issuecomment-1641920726) for MQTT in the past)

It might be a clean solution from the readability point of view, but it still breaks the URI semantics explained in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5) (see my [old comment](https://github.com/w3c/wot-binding-templates/issues/292#issuecomment-1643496844) in the thread cited by @sebastiankb).

We can still ofc bend a little bit the rules... but it doesn't convince me 100%.

**Comment by sebastiankb** [comment]
I have just found an interesting piece of information from the OPC UA specification about [NodeIds](https://reference.opcfoundation.org/Core/Part6/v105/docs/5.1.12)

>The URI portion of [NodeIds](https://reference.opcfoundation.org/search/409?t=NodeIds) are escaped with URI percent encoding as defined in [RFC 3986](https://reference.opcfoundation.org/Core/Part6/v105/docs/?r=RFC3986). Semicolons are added to the list of reserved characters of all URI schemes.

I'm not sure I understand the examples presented, e.g.

>  nsu=tag:acme.com,2023:schemas:data#off%3B;b=M/RbKBsRVkePCePcx24oRA== 

Why only `#` character is encoded here?

I will discuss this in the next UA WG meeting. @relu91 do you have time to join us on Tuesday at 4 pm CET?

**Comment by sebastiankb** [comment]
At yesterday's working group meeting, we achieved a breakthrough with the following solution:

```text
opc.tcp://<address>:<port>/?id=<nodeId>

Where is:

    {address} OPC UA server (IP) address
    {port} OPC UA server port number
    {nodeId} OPC UA NodeId with the following expectations:
            1) any hash character (#) must be URL encoded (%23)
            2) any ampersand character (&) must be URL encoded (%26)  
    
    Examples:
      - "href":"opc.tcp://192.168.120.237:4840/?id=ns=10;i=12345"   
      - "href":"opc.tcp://192.168.120.237:4840/?id=nsu=http://widgets.com/schemas/hello;s=水 World"  
      - "href":"/?id=nsu=http://example.com/hello%23;s=temperature"                        
 
```

egekorkan:
W3C WoT TD TF call today:
- We should add something saying that "Encoder SHOULD be special and only encode `#` and `&`" but this is not a MUST since nothing would break if someone encodes everything, other than humans trying to read it.
- For the implementers, we should tell them that they can just decode it using any url decoder.
- Following the both points above: Most compression algorithms specify the decoder and the encoder can be implemented as wanted (smarter or not).
- We should probably escape `%` as well. We are not aware of any node id containing that but it is theoratically possible.
- The spec text should reference both "URL Encoding/Decoding" and "Percent Encoding/Decoding". The latter is apparently the correct wording. See https://en.wikipedia.org/wiki/Percent-encoding

danielpeintner:
Hi @erossignon 
With https://github.com/eclipse-thingweb/node-wot/pull/1343 we upgraded to TypeScript 5 and moved to `"module": "Node16"`.
Can you merge in these updates into your PR? After thatwe can merge your contribution 👍. Thanks!

packages/examples/tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "outDir": "dist",
-        "module": "commonjs",
+        "module": "Node16",
```
danielpeintner, line 5: @relu91 you kept `"commonjs"` because of scrambled output, right? Do you mind changing it?

packages/binding-opcua/tsconfig.json
```diff
@@ -2,7 +2,10 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDir": "src"
+        "rootDir": "src",
+        "skipLibCheck": true,
+        "module": "Node16",
+        "moduleResolution": "Node16"
```
danielpeintner, line 8: I don't think those 2 statements are necessary since the root `tsconfig.json` defines them already like this.

packages/binding-opcua/tsconfig.json
```diff
@@ -2,7 +2,10 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "outDir": "dist",
-        "rootDir": "src"
+        "rootDir": "src",
+        "skipLibCheck": true,
```
danielpeintner, line 6: Do we run into issues if we do the checking?

package.json
```diff
@@ -51,7 +52,7 @@
         "@types/chai-as-promised": "^7.1.4",
         "@types/chai-spies": "^1.0.4",
         "@types/mocha": "^9.0.0",
-        "@types/node": "^18.19.71",
+        "@types/node": "22.13.1",
```
danielpeintner, line 55: Is this change necessary? @relu91 ?

danielpeintner (commented):
I think we are almost there. I have some questions and propose some changes. See below.
Thank you @erossignon 👍

packages/examples/tsconfig.json
relu91: Yes, this should be kept, any reason for changing it @erossignon ?

package.json
relu91: we still support Node 18, so the maximum upgrade that we can do here is to move the minor. See https://github.com/eclipse-thingweb/node-wot/blob/master/.github/workflows/ci.yaml#L22

egekorkan:
@sebastiankb @danielpeintner could you document the decision that was taken in this PR or point to the decision?

danielpeintner:
> @sebastiankb @danielpeintner could you document the decision that was taken in this PR or point to the decision?

We have to differentiate 2 things here

1. This PR extends the _existing_ node-wot implementation with support for expanded node ids
2. The discussions/comments in the PR were (mis-)used to discuss the future of the OPC UA URI scheme etc. The PR does not yet implement it. This aspect and the outcome should be moved to the binding repo or so. Essentially this [comment](https://github.com/eclipse-thingweb/node-wot/pull/1335#issuecomment-2593027121). @egekorkan do you want to do that?

danielpeintner:
@erossignon do you think you can resolve the last pending issues. We are very close 👍

packages/binding-opcua/tsconfig.json
erossignon: "skipLibCheck": true required only for binding-opcua

package.json
erossignon: This should not cause any problem and will put you ahead of the curven but we can downgrade,

danielpeintner: I noticed that the CI runs on Node 18.x and 20.x were fine. @relu91 any opinion?

relu91: > I noticed that the CI runs on Node 18.x and 20.x were fine. @relu91 any opinion?

They might run fine now, but we would notice only late if Node 22.x introduces any breaking changes in the API. This is something very rare but it can happen. Since we don't need any new feature of Node 22 (yet), I'd prefer to stick with 18.x. Plus the scope of this PR is something else and this change is not strictly related.

danielpeintner (approved):

PR #1336, opened 2024-12-02
Move to esbuild instead of browserify

relu91:
This PR should eventually completely move us from the old browserify build to a more modern toolchain based on esbuild and web/test-runner. There are still things to fix:
- [x] Rebase on main
- [x] Create a new workflow only for browser tests (we don't need the matrix for browser testing)
- [x] Configure web/test-runner to run on CI with the playwright or similar but use developer browser locally (?). 
- [x] fix code style errors in web test runner


relu91:
I opted out browser tests by removing the `test` script in the browser bundle. So now the CI will only run node tests and in the next commit, I will create a workflow just for browser tests. The only incovient is that now developers should run `npm run test:all` instead of `npm run test`. `test:all` is automatically executed once you push too.

danielpeintner:
> The only incovient is that now developers should run `npm run test:all` instead of `npm run test`. `test:all` is automatically executed once you push too.

We don't mention the test commands at all (at the moment). Since the CI will run the check, developers will notice problems latest once pushing. Hence, I don't think we _need_ to mention `test:all` (but we can of course).

relu91:
Sorry there were complications with prettier (which I've updated to keep the new runner tests in check). I had to also re-run the format command since it seems that the new prettier has different rules. Sorry for making reviewing this harder... please let me know if there something wrong that you spot.

Regarding how the browser tests are run, they can run using `npm run test:browser` but only if the developer really want to test there. The new workflow should be activated once we merge this PR and should guarantee that even if the developer forgets we get always green light from browser too.

.github/workflows/browser-test.yml
```diff
@@ -0,0 +1,32 @@
+name: Browser Tests
+
+on:
+    workflow_dispatch:
+    push:
+        branches: [main]
```
danielpeintner, line 6:
```suggestion
        branches: [master]
```

.github/workflows/browser-test.yml
```diff
@@ -0,0 +1,32 @@
...
+    pull_request:
+        branches: [main]
```
danielpeintner, line 8:
```suggestion
        branches: [master]
```

danielpeintner (changes requested):
I didn't spot any obvious issue apart from "master" vs "main".

After fixing that, I am fine with merging...

Thanks a lot @relu91 !!

relu91:
> I didn't spot any obvious issue apart from "master" vs "main".
> 
> After fixing that, I am fine with merging...
> 
> Thanks a lot @relu91 !!

You are right! that's why they didn't run!!! nice catch. pushed the fix.

relu91:
Now we hit a bug with optional dependencies: 
> Cannot find module @rollup/rollup-linux-x64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828).

This [error](https://github.com/eclipse-thingweb/node-wot/actions/runs/12317389224/job/34379768120?pr=1336#step:6:17) is suggesting to re-generate the `package-lock.json`

danielpeintner (approved):
I think now everything seems to work 👍

PR #1339, opened 2024-12-13
Run prettier on main

relu91:
While working on #1336 I noticed that some files got "out of style" conventions. This PR fixes the problem so that we can get a clean CI again.


relu91:
Weird we got errors even if I run prettier locally an fixed the same files. I'll try to rebase from master once again in a new PR.

Issue #1267, opened 2024-04-19
Upgrading to ES2021


egekorkan:
As agreed last week, we will upgrade to Node.js ES2021 and if it doesn't work, we can go for older versions. We will bump minor version, i.e. going to 0.9.0.

We will merge #1266 and #1258 before doing that.

relu91:
Tried to upgrade to ES2021 and as @danielpeintner predicted something broke. It seems that `browserify` can't handle modern js syntax out of the box. Instead of keeping personalizing browserify I would prefer to upgrade to more modern solutions and therefore this issue is blocked by: https://github.com/eclipse-thingweb/node-wot/issues/387

egekorkan:
We can try something like https://parceljs.org that allows customization (we can use readable stream 4) and is new and fast.

PR #1340, opened 2024-12-13
Move to es2021

relu91:
As explained in #1267 we decided to move to a more modern js target. After #1336 everything goes pretty smooth, I'll double check next week if there is anything more to do (or please just comment below if you know more).

### DIFF
--- a/examples/scripts/countdown.js
+++ b/examples/scripts/countdown.js
@@ -80,7 +80,7 @@ WoT.produce({
                 const listToDelete = [];
                 for (const id of countdowns.keys()) {
                     const as = countdowns.get(id);
-                    if ((as === null || as === void 0 ? void 0 : as.output) !== undefined) {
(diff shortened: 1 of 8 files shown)


danielpeintner (approved):
It looks everything seems to work 👍

Maybe we can remove the statement here

https://github.com/eclipse-thingweb/node-wot/blob/5ab4236e24a8087d5b87252951ec39d32fe0f3f8/packages/examples/tsconfig.json#L6

as part of this PR

relu91:
Should I also update the examples if needed?

danielpeintner (approved):
Looks good 👍

In a follow-up, we might want to get rid of the whole workflow 

https://github.com/eclipse-thingweb/node-wot/blob/5afa858b7a27ca664adc14850ce54c995c58acfa/packages/examples/README.md?plain=1#L14-L29

PR #1341, opened 2024-12-16
docs: list different set of tests

danielpeintner:
After https://github.com/eclipse-thingweb/node-wot/pull/1336 it might make sense to list the possible set of tests.

@relu91 do you think this is the right place?


relu91 (approved):
Yes I think it works 👍🏻 thanks!

Issue #1344, opened 2025-01-29
readProperty documentation fault


BinkertR:
The [API documentation](https://github.com/eclipse-thingweb/node-wot/blob/master/API.md) of how to read a property is currently faulty.

current description is:
```
let read1 = await thing.readProperty("count");
console.info("count value is", read1);
```
but it is necessary to read the value like:
```
let read1 = await thing.readProperty("count");
let value = await read1.value()
console.info("count value is", value);
```

egekorkan:
I would keep this document and update it. This will be also rendered on the new website

relu91:
Yes, let's discuss the concrete steps in the next committer meeting. Meanwhile, we can merge #1345

relu91:
@danielpeintner @egekorkan shall we close this? Feel free to re-open it if you need it.

PR #1345, opened 2025-01-29
fix API documentation for readProperty

BinkertR:
fix the API description of how to read a property value. 
As described in #1344

### DIFF
--- a/API.md
+++ b/API.md
@@ -278,7 +278,8 @@ Here we use the await functionality of Node.js.
 
 ```javascript
 let read1 = await thing.readProperty("count");
-console.info("count value is", read1);
+let value = await read1.value();
+console.info("count value is", value);
 ```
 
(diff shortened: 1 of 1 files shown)


relu91 (approved):
I would accept this small fix even if most of the content needs to be moved and fixed it is still correct and valuable.

danielpeintner:
It is fine by me also even if more work is needed.

@BinkertR can you look into the Eclipse ECA issue.. you need to register and accept the Eclipse rules *once*
At the moment, your GitHub user isn't known to Eclipse
https://api.eclipse.org/git/eca/status/gh/eclipse-thingweb/node-wot/1345

danielpeintner:
Strange, I cannot restart the CI pipeline either, but the change looks fine --> merging.

Thanks @BinkertR

PR #1346, opened 2025-01-29
Prepare for next release with new minor version 0.9.0

danielpeintner:
Minor version change needed due to
https://github.com/eclipse-thingweb/node-wot/pull/1332

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,32 +52,13 @@
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.25.7",
+            "version": "7.26.2",
+            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.26.2.tgz",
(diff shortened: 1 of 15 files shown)


danielpeintner:
FYI: Locally, I got *once* an error that I haven't been seeing before 

grafik

For the CI it seems to work fine... maybe just an outlier

JKRhb (approved):

JKRhb:
Thank you for preparing the new release, @danielpeintner! I just ran the tests locally (under Linux with Node 22) and there they passed :) Does increasing the timeout maybe have an influence on whether the test is failing?

danielpeintner:
> Thank you for preparing the new release, @danielpeintner! I just ran the tests locally (under Linux with Node 22) and there they passed :) Does increasing the timeout maybe have an influence on whether the test is failing?

Mhh, not sure if we should do anything. Re-running the tests didn't cause anymore the issue locally. Maybe it was just something *temporarily* broken on my machine, but I thought I share it anyway

relu91:
I think the problem is that the failing test is leveraging on the fact that the TM is not available under example.com . It is a pretty weak assumption and it depends on the fact of running tets with an active internet connection. we could design it better ... but we can still live with it.

relu91 (approved):

Issue #1347, opened 2025-02-07
Outdated documentation about generating examples


relu91:
> Working with ES2021 does not seem to generate that ugly JS anymore, what do you think?

That is great! Once we move on with https://github.com/eclipse-thingweb/node-wot/pull/1340 we can also remove the workflow described here

https://github.com/eclipse-thingweb/node-wot/blob/5afa858b7a27ca664adc14850ce54c995c58acfa/packages/examples/README.md?plain=1#L14-L29

_Originally posted by @danielpeintner in https://github.com/eclipse-thingweb/node-wot/issues/1215#issuecomment-2547834527_

PR #1348, opened 2025-02-10
refactor: update documentation about generating JS examples

danielpeintner:
Update the JS output accordingly also

fixes https://github.com/eclipse-thingweb/node-wot/issues/1347

### DIFF
--- a/examples/quickstart/presence-sensor.js
+++ b/examples/quickstart/presence-sensor.js
@@ -13,6 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
+Object.defineProperty(exports, "__esModule", { value: true });
 // This is an example Thing script which is a simple presence detector
 // It fires an event when it detects a person (mocked as every 5 second)
 const core_1 = require("@node-wot/core");
--- a/examples/scripts/countdown.js
+++ b/examples/scripts/countdown.js
@@ -1,5 +1,6 @@
+"use strict";
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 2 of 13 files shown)


danielpeintner:
Mhh, the failing "coding style check" seems to have something to do with my local setup. See also https://github.com/eclipse-thingweb/node-wot/issues/1349

examples/quickstart/presence-sensor.js
```diff
@@ -13,6 +13,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
  ********************************************************************************/
+Object.defineProperty(exports, "__esModule", { value: true });
```
relu91, line 16: This should not be there (according to the tests that I did in the past), I'll double-check locally to understand why it generates this line now.

examples/scripts/countdown.js
```diff
@@ -1,5 +1,6 @@
+"use strict";
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
```
relu91, line 3: These license changes are ok? should we maybe update the ones coming from TS ? are they good?

relu91 (commented):
I had similar problems in previous PRs, it might be a misalignment with the version of prettier used by CI and the one used by our local stack. 

About the PR itself, I found some issues below

examples/scripts/countdown.js
danielpeintner: I noticed that also, but I think we should not bother too much. I think updating according to TS files is fine...

danielpeintner:
> I had similar problems in previous PRs, it might be a misalignment with the version of prettier used by CI and the one used by our local stack.

What I don't understand is how does the prettier GH action gets the information which prettier options are to be used. Does it just pick up  .prettierrc.json from the repository? 

https://github.com/eclipse-thingweb/node-wot/blob/4d083fcbb35e73dfcd2e15eddcbfc3987d793dca/.github/workflows/ci.yaml#L102-L109

danielpeintner:
Based on https://github.com/eclipse-thingweb/node-wot/pull/1335#discussion_r1942425485 and the change to `Node16` the following line is still there

`Object.defineProperty(exports, "__esModule", { value: true });`

I convert this PR to draft given that
* we need to figure out whether there is a way to get rid (automatically) of this unwanted line
* we (I ?) have issues with prettier

PR #1362, opened 2025-03-13
style: npm run format

danielpeintner:
Sync with latest prettier version/settings

fixes https://github.com/eclipse-thingweb/node-wot/issues/1349

PR #1363, opened 2025-03-13
refactor: remove td-tools package

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1301


danielpeintner:
The code style issues should be solved once https://github.com/eclipse-thingweb/node-wot/pull/1362 is merged.

Anyhow, I noticed one more thing about the browser example that no longer work.
e.g., examples/browser/index.html

grafik

relu91 (approved):

relu91 (approved):
I checked a little bit the versions and I found out that we already published a deprecation message in 0.8.x (see https://www.npmjs.com/package/@node-wot/td-tools/v/0.8.16). So I think that given the current versioning approach that we are using ( minor increase ~= major increase), we should have already removed it. Therefore I would go ahead with this PR 👍🏻

relu91:
btw, should we also deprecate the package in npm? I think so.
image

danielpeintner:
> btw, should we also deprecate the package in npm?

Yes, I did set it as deprecated now.

relu91 (approved):

Issue #790, opened 2022-06-29
Use streams in write property operation on browsers


relu91:
I've created this issue mostly to monitor the two issues that I linked below. 
> After plumbing into the inner workings of broserify and the Web platform, I've found that sadly stream upload is not supported for browsers. This means that we can't use a stream to write properties or invoke actions.  Solving this issue is going to be tricky but I hope to find a patch that will only be active on browser environments. 

For the ones that are curious. Here are the relevant issues:
https://github.com/whatwg/fetch/issues/1438 
https://bugs.chromium.org/p/chromium/issues/detail?id=688906

_Originally posted by @relu91 in https://github.com/eclipse/thingweb.node-wot/issues/785#issuecomment-1169744324_

samubura:
I'm having issues using the browser bundle to invoke actions
the error I get is:

    TypeError: stream_1.Readable.from is not a function
    valueToContent https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@0.8.1/dist/wot-bundle.min.js:49981
    invokeAction https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@0.8.1/dist/wot-bundle.min.js:49762
    __awaiter https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@0.8.1/dist/wot-bundle.min.js:49326
    __awaiter https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@0.8.1/dist/wot-bundle.min.js:49322
    invokeAction https://cdn.jsdelivr.net/npm/@node-wot/browser-bundle@0.8.1/dist/wot-bundle.min.js:49747`
    
is this related to this issue?

danielpeintner:
The issue was resolved by https://github.com/eclipse/thingweb.node-wot/pull/809.

This was done on 27 July and no more release has been published since than. I hope we can publish a *new* release soon. In the meantime you may want to use `wot-bundle.min.js` from a master build.

relu91:
The new release is out, @samubura can you confirm that you don't experience that issue anymore?

samubura:
Yes! I can confirm it is working as expected now.

relu91:
The tracked issue is now resolved, but the landscape is not really consistent. Check this [report](https://staging.wpt.fyi/results/fetch/api/basic/request-upload.any.html?label=experimental&label=master&aligned)

relu91:
[...]

PR #1369, opened 2025-04-12
Handle long streams as inputs in HTTP binding

### LINKED ISSUE #1366: InvokeAction with stream (NodeJS.ReadableStream) as body sends only first chunk
I use a NodeJS.ReadableStream as body in the invokeAction method to send the response body of a previous fetch request as input for the action call. On the server side only the first chunk is received and the request hangs and is never ended. The code looks like the following: 

`
import fetch from 'node-fetch'
import Servient, { ProtocolHelpers } from '@node-wot/core';
import { HttpClientFactory } from '@node-wot/binding-http';

//Fetching the resource from an endpoint as stream
const response = await fetch('https://someurl/resourcendpoint', {method: 'GET'})
//response.body is of NodeJS.ReadableStream
const result = await response.body 

//Create servient
const thingDescriptionId = "someUniqueThingDescription";
const servient = new Servient();
servient.addClientFactory(new HttpClientFactory());

//Thing Description
const thingDescription = {
   "id": "someUniqueThingDescription",
   "title": "myDescription",
   "@context": [
        "https://www.w3.org/2019/wot/td/v1",
        {                                                                                                                      
              "schema": "https://schema.org#"
        }                                                                                                                  
   ],
    "@type": "Thing",  
   "securityDefinitions": {
    "no_sec": {
      "scheme": "nosec"
    }
  },
   "actions": {
       "postAction": {
           "forms": [
                {
                    "op": "invokeaction",
                    "href": "http://127.0.0.1:30000/postAction",
                    "contentType": "application/octet-stream",
                    "response": {
                     },
                     "htv:headers": [
                          {
                               "htv:fieldName": "Transfer-Encoding",
                               "htv:fieldValue": "chunked"
                          }
                     ]
                }
            ],
            "output": {
                "type": "object"
            },
            "input": {
                "type": "string"
            }
        }
    }
}

//Consume the thing Description
const consumedThing = await WoT.consume(thingDescription);
//Invoke the action with the stream as input
const actionResult = await consumedThing.invokeAction('postAction', ProtocolHelpers.toWotStream(result))
`

On server side we see that the first chunk is received but nothing more. I also recognized that when the stream is previously read for example with readStreamFully and then given to the toWotStream function the request works. It would be nice, if i can just use the result stream of the first fetch request as input for the invokeAction call. Or do I have to handle the stream differently? Can you help me with that?

relu91:
This simple PR fixes #1366 by forcefully passing the request body to the node-fetch operation. I did some experiments. Basically, previously, the content parameter in the private fetch function was redundant, as it was never actually passed. That surprised me, Then I looked in the generateFetchRequest and found that we were already handling the body in that function. Removing it didn't solve the problem, but then I noticed that if I forcefully pass the body directly in the fetch RequestInit argument, then the stream is consumed correctly. I reproduced the error (the need to pass the body to request init) without using node-wot at all so I assumed that there should be something wrong with node-fetch (digging into its code, there is probably something when it clones the ReadableStream). Therefore, I left a comment for future review. Headers and other parameters seem not to be impacted. 

While I was at it, I upgraded node-fetch to the last compatible version (3.x works only with ESM).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9918,48 +9918,6 @@
                 "node-aead-crypto": "^2.0.0"
             }
         },
-        "node_modules/node-fetch": {
-            "version": "2.7.0",
-            "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.7.0.tgz",
-            "integrity": "sha512-c4FRfUm/dbcWZ7U+1Wq0AwCyFL+3nt2bEw05wfxSz+DWpWsitgmSgYmy2dQdWyKC1694ELPqMs/YzUSNozLt8A==",
-            "license": "MIT",
-            "dependencies": {
-                "whatwg-url": "^5.0.0"
-            },
-            "engines": {
-                "node": "4.x || >=6.0.0"
-            },
-            "peerDependencies": {
-                "encoding": "^0.1.0"
-            },
-            "peerDependenciesMeta": {
-                "encoding": {
-                    "optional": true
-                }
-            }
-        },
-        "node_modules/node-fetch/node_modules/tr46": {
-            "version": "0.0.3",
-            "resolved": "https://registry.npmjs.org/tr46/-/tr46-0.0.3.tgz",
-            "integrity": "sha512-N3WMsuqV66lT30CrXNbEjx4GEwlow3v6rr4mCcv6prnfwhS01rkgyFdjPNBYd9br7LpXV1+Emh01fHnq2Gdgrw==",
-            "license": "MIT"
-        },
-        "node_modules/node-fetch/node_modules/webidl-conversions": {
-            "version": "3.0.1",
-            "resolved": "https://registry.npmjs.org/webidl-conversions/-/webidl-conversions-3.0.1.tgz",
-            "integrity": "sha512-2JAn3z8AR6rjK8Sm8orRC0h/bcl/DqL7tRPdGZ4I1CjdF+EaMLmYxBHyXuKL849eucPFhvBoxMsflfOb8kxaeQ==",
-            "license": "BSD-2-Clause"
-        },
-        "node_modules/node-fetch/node_modules/whatwg-url": {
-            "version": "5.0.0",
-            "resolved": "https://registry.npmjs.org/whatwg-url/-/whatwg-url-5.0.0.tgz",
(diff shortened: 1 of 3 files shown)


danielpeintner (approved):
LGTM

The "bug" is tracked by https://github.com/eclipse-thingweb/node-wot/issues/1366 also

danielpeintner:
Two more points:
1. The initial failing tests were a CI issue with codecov but the browser tests are still failing
2. Maybe @RRebekka can check out the fix also

relu91:
> The initial failing tests were a CI issue with codecov but the browser tests are still failing

Yes, it's a good thing that we introduced them 😆 . I've already a fix, I might publish it this afternoon cause I wanted to learn more about why it is not working... it is related to https://github.com/eclipse-thingweb/node-wot/issues/790 .

relu91:
I did some tests and I concluded that it is safer to keep the current approach -> translate the payload to a buffer and then send it. As far as I discovered things moved a bit from #790 , but the support is not still widely available. Moreover, in my tests, I found that Chromium (inside playwright) requires HTTP/2 (or even 3) if you want to send a body streaming in a request. If we want to support this, I'm afraid we will need to try it on many test configurations (browsers, and with or without a http/2 server). 

I'd keep an eye out for further use cases, but I think that the current implementation is not that bad, even if it is less performant than a full streaming approach.

danielpeintner (approved):
LGTM

The only thing I don't really like too much is the overuse of fetch methods and the confusion that comes along with it ;-)
We have the actual fetch(...) and then we have HttpClient.fetch(...).

_fetch(...) is fine since it is another name 

Maybe we can think about renaming the private HttpClient.fetch() but it was the case already before... so maybe it is fine to leave it as is...

relu91:
> LGTM
> 
> The only thing I don't really like too much is the overuse of fetch methods and the confusion that comes along with it ;-)
> We have the actual fetch(...) and then we have HttpClient.fetch(...).
> 
> _fetch(...) is fine since it is another name 
> 
> Maybe we can think about renaming the private HttpClient.fetch() but it was the case already before... so maybe it is fine to leave it as is...


Agreed, any suggestions for a possible naming refactoring?? `doOperation`?

danielpeintner (approved):
LGTM

PR #1375, opened 2025-05-06
refactor: generate unique names/urls locally

danielpeintner:
Yet another approach... (alternative for https://github.com/eclipse-thingweb/node-wot/pull/1352)

Advantages I see:
* each binding can choose the way it likes to create urls if there are thing name clashes. This is anyway a bit of a problematic case since no one can rely on the strategy since it depends on the order a thing has been added.
* no need for a helpers method
* no need to pass (and create) a set around with the existing names so far

fixes https://github.com/eclipse-thingweb/node-wot/issues/1351

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -160,8 +160,14 @@ export default class CoapServer implements ProtocolServer {
     private createThingUrlPath(thing: ExposedThing) {
         let urlPath = slugify(thing.title, { lower: true });
 
-        while (this.things.has(urlPath)) {
-            urlPath += "_";
+        // avoid URL clashes
+        if (this.things.has(urlPath)) {
+            let uniqueUrlPath;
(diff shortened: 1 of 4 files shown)


danielpeintner:
@relu91 doesn't like adding "`_`" since it may become hard to see how man underscores have been used.
I tend to agree with that statement and I will rework the PR.

relu91 (approved):
Ok for going with this direction, hoping then, in the future, we will have a more consistent way to generate URLs of afforndances! Thank you @danielpeintner !!!

JKRhb (approved):
I am also fine with the direction :) Maybe we can refactor the code at some point in the future and move the "sluggification" to a central location to ensure a consistent naming scheme, but for now, this looks good to me :)

Edit: Or rather: Integrate the name clash avoidance into a common sluggification function.

danielpeintner:
Thank you for your comments!

Issue #1380, opened 2025-05-14
Update TS dependencies from Scripting (March 2025)


danielpeintner:
wot-typescript-definitions@0.8.0-SNAPSHOT.30
wot-thing-description-types@1.1.0-12-March-2025
?? wot-thing-model-types@1.1.0-12-March-2025 ??

danielpeintner:
Note: @thingweb/thing-model needs to be updated first (is used in `cli` and `core`).

PR #1381, opened 2025-05-15
chore: update Scripting TS dependencies (March 2025)

danielpeintner:
Note: @thingweb/thing-model needs to be updates first to get a *clean* package-lock.json

fixes https://github.com/eclipse-thingweb/node-wot/issues/1380

### DIFF
--- a/examples/templates/exposed-thing/package.json
+++ b/examples/templates/exposed-thing/package.json
@@ -11,7 +11,7 @@
     "license": "MIT",
     "devDependencies": {
         "typescript": "4.7.4",
-        "wot-typescript-definitions": "0.8.0-SNAPSHOT.26",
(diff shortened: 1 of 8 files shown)


danielpeintner:
Waiting for https://github.com/eclipse-thingweb/td-tools/pull/62 to be merged and published

danielpeintner:
General note: the new dependency structure (td-tools externalized) makes it somewhat more complicated to update the TS definitions from the Scripting API. We need to first update td-tools and later node-wot otherwise we are running into colliding definitions 😒

relu91:
> General note: the new dependency structure (td-tools externalized) makes it somewhat more complicated to update the TS definitions from the Scripting API. We need to first update td-tools and later node-wot otherwise we are running into colliding definitions 😒

let's discuss briefly in today's committer call

relu91 (approved):

PR #1383, opened 2025-05-16
Remove unsused dependencies

danielpeintner:
I used `npm-check` to identify unsued dependencies. Unfortunately, it gives lots of false positives, and it ends up with trial and error 🤷‍♂️

Anyhow, I managed to get rid of some ..

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15483,13 +15483,11 @@
                 "find-my-way": "^8.2.2",
                 "node-fetch": "^2.7.0",
                 "query-string": "^7.1.1",
-                "rxjs": "5.5.11",
                 "slugify": "^1.4.5"
             },
             "devDependencies": {
                 "@node-oauth/express-oauth-server": "^3.0.1",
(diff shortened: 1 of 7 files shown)


danielpeintner:
The test runs are green but eslint suggests adding some (indirect?) dependencies back.

relu91:
> The test runs are green but eslint suggests adding some (indirect?) dependencies back.

I think they are not indirect... they are just not covered by tests (?) For example in [helpers ](https://github.com/eclipse-thingweb/node-wot/blob/a457f120ffdf8eadc23821642b81db6671321dc7/packages/core/src/helpers.ts#L258) we need the dep to check if it is really a ThingModel...

danielpeintner:
Do you think it is worth removing those unused dependencies in this PR?

danielpeintner:
I added @jspm/core@2.1.0 back in browser-bundle

The remaining changes are minimal. I let others decide whether it is still worth to merge...

relu91 (approved):

egekorkan (approved):

PR #1387, opened 2025-05-22
feat: enhance browser examples, added tab menu and auto consume and improvise ui

manudev0004:
Fixed the redirect issue, Added tab menu for example TD, dropdown to choose from examples. Auto consume feature and Enhanced UI slightly


relu91:
I would at least fix the layout, then yes, we can improve it in a follow-up PR.

egekorkan:
We talked in the committer meeting. @manudev0004 it would be good to fix the tabs:

- Removing duplication (no drop-down needed)
- Custom should be the last tab
- The description below each tab can be simply `Consumed TD at {URL}`, and then we can use the `description` field of the TD to fill the field. If there is an error with fetching, the error should be displayed here (like in Counter).

The rest of the layout issues can be fixed in another PR :)

danielpeintner:
I checked the update and I think we are close.

I noticed some more things.

1.) Error clearAllInteractions 

In the case of the counter, which fails at the moment, I also get some errors that we shouldn't see like
> clearAllInteractions is not defined

The same happen if I try to load an invalid URL in custom tab.

grafik


2.) Hitting "Enter" on custom tab textfield after inserting URL could call "Consume"

Apart from the above, I think we are fine now and can tackle the other issues. I created https://github.com/eclipse-thingweb/node-wot/issues/1389 for that purpose.

examples/browser/index.js
```diff
@@ -163,9 +188,168 @@ function removeSchemaEditor() {
...
+        descriptionElement.style.color = "";
+    } else {
+        descriptionElement.textContent = "No description available";
+        descriptionElement.style.color = "";
+    }
+}
+
+// Clear all interactions and editor
+function clearAllInteractions() {
+    console.log("clearAllInteractions function called");
```
relu91: Remove this leftover.

relu91 (changes requested):
We are almost there. Currently, when the fetch fails, the error message is displayed in the TD description. I think this is counterintuitive. Please reintroduce the error banner and hide the TD Description field if the fetch fails.

examples/browser/index.js
danielpeintner: FYI: @relu91 refers to the console log only...

danielpeintner (commented):
Besides the comments from @relu91 it looks good to me.

relu91:
@danielpeintner, please explicitly approve.

danielpeintner (approved):
LGTM 👍

danielpeintner:
A nice follow-up work that come out of this PR is https://github.com/eclipse-thingweb/node-wot/issues/1389
@manudev0004

Issue #1395, opened 2025-06-12
AJV dependency multiplication in opcua binding


egekorkan:
AJV is specified for the second time in the binding opc ua code: https://github.com/eclipse-thingweb/node-wot/blob/master/packages/core/package.json

In http it is not and there should be no reason to add it again in opcua

danielpeintner:
Question: I agree it is not necessary. On the other hand, it should not hurt as long as we use the exact same version, right?
Note: we have a check for that...

danielpeintner:
@egekorkan I think it is actually needed, see https://github.com/eclipse-thingweb/node-wot/actions/runs/15610649514/job/43970495245?pr=1396

PR #1396, opened 2025-06-12
chore: remove duplicated AJV dependency from binding-opcua and cli

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1395

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15554,8 +15554,6 @@
             "license": "EPL-2.0 OR W3C-20150513",
             "dependencies": {
                 "@node-wot/core": "0.9.2",
-                "ajv": "^8.11.0",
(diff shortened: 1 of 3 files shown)


danielpeintner:
@egekorkan The CI fails with the according error... 
Hence, I think adding AJV multiple times is fine

egekorkan:
@danielpeintner this is quite weird and should not be necessary no? If there is no way around it, of course it is fine to keep track of it

danielpeintner:
I think it doesn't hurt having the dependency in opcua and cli also. Hence, I think we should close the issue and the PR...

relu91:
ok for closing @egekorkan?

PR #1397, opened 2025-06-13
Add skeleton for query and cancel action

danielpeintner:
Updating "wot-typescript-definitions" to version "0.8.0-SNAPSHOT.31"

see also https://github.com/w3c/wot-scripting-api/pull/561

### DIFF
--- a/packages/core/test/InteractionOutputTest.ts
+++ b/packages/core/test/InteractionOutputTest.ts
@@ -18,7 +18,7 @@ import { suite, test } from "@testdeck/mocha";
 import promised from "chai-as-promised";
 import { expect, use } from "chai";
 import { Readable } from "stream";
-import { InteractionOutput } from "../src/interaction-output";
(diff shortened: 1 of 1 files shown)


danielpeintner:
Note: we have an option `ignoreValidation` that at the moment it is used for async actions only. Hence, we could remove that option for `InteractionOutput` (and keep it for `ActionInteractionOutput` only). However, I think we should keep it as is.

relu91:
I'm not sure whether to keep it as it was designed as an escape hatch to support async actions... Now that we have introduced a new kind of interaction output, we should implement that business logic (e.g., ignore validation) inside it.

danielpeintner:
I still think that having the possibility to tweak the `InteractionOutput` so that I can decide whether to validate data or not makes sense.
Besides async actions, we discussed other aspect such as performance and such also.

Anyhow, I don't have a very strong feeling either way. I tend to keep it as said. If you ask me to remove it, I can do it.

Issue #1400, opened 2025-06-24
ContentType seems to confuse OPC UA


danielpeintner:
The following code tries to read an OPC UA property

```
const read1 = await thing.readProperty("1:Speed");
console.log("speed value is: ", await read1.value());
```

with the following TD

```
              "properties": {
                  "1:Speed": {
                      "title": "Speed",
                      "type": "number",
                      "forms": [
                          {
                              "href": "/?id=ns=1;s=MotorSpeed"
                          }
                      ],
                      "observable": true
                  }
             }
```

works fine while adding a  `"contentType": "application/octet-stream"` which looks like this breaks it.

```
              "properties": {
                  "1:Speed": {
                      "title": "Speed",
                      "type": "number",
                      "forms": [
                          {
                              "href": "/?id=ns=1;s=MotorSpeed",
                              "contentType": "application/octet-stream"
                          }
                      ],
                      "observable": true
                  }
             }
```

fails with 

```
TypeError: dataType.toLowerCase is not a function
at OctetstreamCodec.valueToBytes 
```

This seems to indicate that OPC UA data is processed by OctetstreamCodec also (again?)

danielpeintner:
I _think_ the problem is, that the [node-opcua](https://github.com/node-opcua/node-opcua) library we are using converts the data on the wire directly to a [DataValue](https://github.com/node-opcua/node-opcua/blob/53bc91f8992b50b9ddb938e279f78ac1026ec13f/packages/node-opcua-data-value/source/datavalue.ts#L259) class by calling `read(...)`.

https://github.com/eclipse-thingweb/node-wot/blob/3325e129b5fe7fb9fdb663c55d1a2101dccd9b11/packages/binding-opcua/src/opcua-protocol-client.ts#L305-L313

Maybe it is possible to get the raw data, but I am not sure.
Hence, I wonder whether a simplified fix is also reasonable. What do I mean by that, we remove the `contentType` for any OPC UA form so that no other ContentSerdes codec can kick in?

@sebastiankb @egekorkan do you think that is okay?
@erossignon do you have another proposal?

danielpeintner:
In https://github.com/eclipse-thingweb/node-wot/pull/1402 I show what I intend to do...

danielpeintner:
> The developer should not need to do that

BTW, the developer doesn't need to do anything. We take care internally...

danielpeintner:
@erossignon do you plan to work on the content type issues?

[...]

PR #1402, opened 2025-06-25
Avoid external contentType processing for OPC UA

danielpeintner:
provides a workaround for https://github.com/eclipse-thingweb/node-wot/issues/1400, but does not really fix the issue.

### DIFF
--- a/packages/binding-opcua/src/codec.ts
+++ b/packages/binding-opcua/src/codec.ts
@@ -258,7 +258,7 @@ export const theOpcuaJSONCodec = new OpcuaJSONCodec();
 
 export class OpcuaBinaryCodec implements ContentCodec {
     getMediaType(): string {
-        return "application/opcua+octet-stream"; // see Ege
+        return "application/opcua+octet-stream";
     }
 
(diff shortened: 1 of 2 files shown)


packages/binding-opcua/src/codec.ts
```diff
@@ -258,7 +258,7 @@ export const theOpcuaJSONCodec = new OpcuaJSONCodec();
 
 export class OpcuaBinaryCodec implements ContentCodec {
     getMediaType(): string {
-        return "application/opcua+octet-stream"; // see Ege
+        return "application/opcua+octet-stream";
```
danielpeintner, line 261: This is just a left-over I would like to remove

danielpeintner:
We quickly discussed some possible solutions in the committer call and we were wondering whether the `contentType`s used so far 
* `application/opcua+json`
* `application/opcua+octet-stream`

are actually right. Should it be with or without `opcua+` prefix

@sebastiankb ?

erossignon:
assigning to me, an other library is required to convert node-opcua internal objects to OPCUA+JSON compliant one

Issue #1403, opened 2025-06-25
Wrong TD URL returns invalid TD error


egekorkan:
I mistakenly gave a wrong url for the requestThingDescription method (`http://mything.com` instead of `http://mything.com/td`) and got the following error:

```
node_modules\@node-wot\core\dist\validation.js:14
    return new Error(errors);
           ^
 
Error: must be object
    at getLastValidationErrors (code\node_modules\@node-wot\core\dist\validation.js:14:12)
    at WoTImpl.requestThingDescription (code\node_modules\@node-wot\core\dist\wot-impl.js:70:56)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async myFunction (code\dist\main.js:69:16)
 
Node.js v22.15.1
```

So I thought that my TD is invalid and lost some time trying to understand. I think another error should be thrown in this case.

PR #1404, opened 2025-06-25
Improve TD validation error in requestThingDescription

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1403

### DIFF
--- a/packages/binding-http/src/http-client-impl.ts
+++ b/packages/binding-http/src/http-client-impl.ts
@@ -433,6 +433,8 @@ export default class HttpClient implements ProtocolClient {
         if (HttpClient.isOAuthTokenExpired(result, this.credential)) {
             this.credential = await (this.credential as OAuthCredential).refreshToken();
             return await this._fetch(await this.credential.sign(request));
+        } else if (result.status === 401) {
+            throw new Error("Client error: Unauthorized");
         } else if (result.status >= 400) {
             // report generic error
             throw new Error(`request to ${request.url} failed, reason: HTTP status result ${result.status} `);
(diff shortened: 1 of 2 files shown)


relu91:
> Unfortunately, it is not possible to wrap an error in TS/JS.

You should be able to define custom errors. 

```ts
export class ValidationError extends Error {

  public constructor(input: ??) {
    //... 
  }
}
```

or did I miss your point?

danielpeintner:
What I meant was wrapping this error in another error to get a stack trace like in Java

relu91:
> The issue is that you get a result .. in this case, a JSON array as response...

As I said above, if this is the case, the answer @egekorkan described in the issue is indeed correct.

relu91:
> I seem to be on the wrong track somehow!? 

No, you are correct! I was saying that since we are receiving a valid json (and not a 404 or other error codes like a timeout) the error that we are throwing is correct, we can improve as you are doing here, but I would not even bother (but we can move on if you like).

danielpeintner:
I will take a closer look at the possible error cases and look also into `client.requestThingDescription(url)` and check whether a HTTP status >= 400 throws an error also

danielpeintner:
Commit https://github.com/eclipse-thingweb/node-wot/pull/1404/commits/9afaaf283748270cab7ae4e0d6222574e2fdfae3 should add the information needed if an HTTP error results...

danielpeintner:
I think this brings up a new issue 

https://github.com/eclipse-thingweb/node-wot/actions/runs/15978405985/job/45066738871?pr=1404#step:9:310

I think it should be handled by the following part already but it doesn't seem to be the case

https://github.com/eclipse-thingweb/node-wot/blob/699e6b6cc640c60a1ca338050d496f59af7646f7/packages/binding-http/src/http-client-impl.ts#L433

relu91 (approved):
LGTM, as discussed in call. Let's merge it.

PR #1405, opened 2025-06-26
Update exposer.js

IvoVargas:
Correct situation when the consumer invokeAction("sayOk") returns "undefined".


IvoVargas:
@relu91 done 👍

PR #1377, opened 2025-05-07
chore: add Node.js v24 to CI workflow

danielpeintner:
resolves https://github.com/eclipse-thingweb/node-wot/issues/1373

PR #1412, opened 2025-08-19
Add possibility to register ContentCodec with scheme

### LINKED ISSUE #1409: Allow different Codecs for the same contentType
At the moment, it is possible to "register" **one** `Codec` per `contentType` via `ContentSerdes`.

e.g., `ContentSerdes.get().addCodec(new MyCodec("application/myType"))`

There are situations where this is not enough. For example, one might want to register a generic codec for `application/octet-stream` and yet another for the _same_ `application/octet-stream` which is used for binding-opcua only.

We should extend the API in a way to allow for that use-case.

danielpeintner:
TODO calls for contentToValue and valueToContent need to provide scheme

fixes https://github.com/eclipse-thingweb/node-wot/issues/1409


danielpeintner:
Note: The failing tests are in areas that haven't been touched and the error

` Exception during run: SyntaxError[ @/home/runner/work/node-wot/node-wot/packages/binding-websockets/test/ws-tests.ts ]: Invalid or unexpected token`

reminds me to what we have been seeing for Node.js 24.
see https://github.com/eclipse-thingweb/node-wot/pull/1377#issuecomment-3127185692

Anyhow, the failures are coming from Node.js 22

erossignon:
@danielpeintner yes, let's move forward, sorry for the long silence. OPC UA needs this for #1400: core serializes a write before the binding is called, so an OPC UA form with `application/octet-stream` always gets the generic codec.

I rebased your commit onto master and added the missing part as a second commit:

- `ConsumedThing`, `ExposedThing`, `InteractionOutput` and `ProtocolListenerRegistry` pass the scheme of the form href.
- One lookup for every path. It also fixes two gaps: a media type registered only for a scheme was never found on read, and `isSupported()` ignored the scheme.
- Scheme-specific codecs are not listed as offered.

I can't push to your branch, so it is in #1572, with your commit and authorship kept. Close this one in favour of it, or take the second commit into your branch, as you prefer.

One open point: with MQTT over WebSocket (#1475), a `ws://` href doesn't say which binding owns it. Fine for OPC UA (`opc.tcp`), but the key may later need the subprotocol too.

danielpeintner:
Closing in favor of https://github.com/eclipse-thingweb/node-wot/pull/1572

PR #1413, opened 2025-08-21
Fix decode HTTP content-encoding

FreuMi:
Hi everyone!

**Problem:**
When consuming endpoints that return compressed payloads, the HTTP binding currently forwards the compressed bytes directly to the codecs. This causes parsing/validation errors since the codecs expect plain data.

For example, this public datetime API:
```
https://aisenseapi.com/services/v1/datetime
```
responds with:
```
Content-Type: application/json
Content-Encoding: gzip, br
```

With the current HTTP binding, the JSON codec receives compressed bytes and fails with errors such as:
```
Invalid value according to DataSchema
```

Here is the TD I used to reproduce the issue:
```json
{
  "@context": "https://www.w3.org/2022/wot/td/v1.1",
  "title": "dateTimeAPI",
  "securityDefinitions": { "nosec_sc": { "scheme": "nosec" } },
  "security": ["nosec_sc"],
  "properties": {
    "currentDateTime": {
      "description": "Get current datetime",
      "readOnly": true,
      "type": "object",
      "properties": {
        "datetime": { "type": "string", "format": "date-time" }
      },
      "forms": [
        {
          "href": "https://aisenseapi.com/services/v1/datetime",
        }
      ]
    }
  }
}
```

**Fix:**
The HTTP client now checks the `content-encoding` response header and transparently decompresses the payload before passing it to the codecs. I added a helper in `http-client-impl.ts` to handle this logic when a response is compressed.

I thought this fix could be useful for others too.

### DIFF
--- a/packages/binding-http/src/http-client-impl.ts
+++ b/packages/binding-http/src/http-client-impl.ts
@@ -52,6 +52,7 @@ import {
 } from "./credential";
 import { LongPollingSubscription, SSESubscription, InternalSubscription } from "./subscription-protocols";
 import { Readable } from "stream";
+import { createGunzip, createBrotliDecompress, createInflate, createInflateRaw } from "zlib";
 
 const { debug, warn, error } = createLoggers("binding-http", "http-client-impl");
 
@@ -133,7 +134,9 @@ export default class HttpClient implements ProtocolClient {
 
         // in browsers node-fetch uses the native fetch, which returns a ReadableStream
         // not complaint with node. Therefore we have to force the conversion here.
-        const body = ProtocolHelpers.toNodeStream(result.body as Readable);
+        let body = ProtocolHelpers.toNodeStream(result.body as Readable);
+        const enc = result.headers.get("content-encoding") ?? undefined;
+        body = HttpClient.decodeByContentEncoding(body, enc);
         return new Content(result.headers.get("content-type") ?? ContentSerdes.DEFAULT, body);
     }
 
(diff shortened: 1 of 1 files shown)


relu91:
I reviewed the issue. Here are my findings:
- It should definitely be handled by node-fetch, but we have a quite old dependency
- Even with the latest node-fetch, there is a bug where they basically don't handle multiple content-encodings (even though it is allowed by the RFC) 
- The now stable `fetch` node implementation works correctly 

Therefore, I'd keep the issue, but I would fix it by upgrading the HTTP client implementation with the new Node.js native fetch api (which is also cross-platform). However, for node 20, we should set the flag `--experimental-fetch` since it was marked as experimental in that release. Given that node 20 is [sunsetting](https://nodejs.org/en/about/previous-releases) (is going to be around till April 2026), I think this might be a good comprise.

FreuMi:
Thanks @relu91 for the clarifications!
I agree it’s better to upgrade in the future rather than adding a custom workaround, since that would only duplicate logic that should already be handled by the HTTP client.

If you want, I can close this PR and open an issue instead.

danielpeintner:
> @danielpeintner are you ok to close this PR and open an issue?

If I understand the proposal correctly, we wait for May 2026 (Node.js no longer maintained) and switch to native fetch
OR
we change already now to native fetch and use the experimental flag?

Either way is fine by me.
Closing this PR and creating the issue is definitely fine!

relu91:
> If you want, I can close this PR and open an issue instead.

@FreuMi if you don't mind, can you create the issue?

Issue #1401, opened 2025-06-24
Security for OPC UA


danielpeintner:
According to https://github.com/node-opcua/node-opcua the library supports several "Security Policies" and "Authentication"

![Image](https://github.com/user-attachments/assets/5d7b77dd-9105-4ab6-b526-682554562376)

In node-wot the code doesn't do anything

https://github.com/eclipse-thingweb/node-wot/blob/3325e129b5fe7fb9fdb663c55d1a2101dccd9b11/packages/binding-opcua/src/opcua-protocol-client.ts#L474-L477

erossignon:
I'll keep it opent to cover:


[ ] documentation 
[ ] proposing our extension to the OPC Foundation Working group  uav:channel-security and uav:authentication as at this time the spec is relatively loose on security options.

erossignon:
@danielpeintner — to answer your question directly: **not resolved**, and the reason is new information rather than missing code.

The OPC Foundation has since published **OPC 10101 "OPC UA for WoT Binding" v1.00** (2026-01-08): https://reference.opcfoundation.org/specs/OPC-10101

That retires the second checkbox above. I had proposed taking `uav:channel-security` / `uav:authentication` to the working group on the grounds that the spec was loose on security. It is not loose any more — the WG published its own naming, and ours does not match it.

Comparing our implementation against §6.3.3 and §6.5:

| OPC-10101 v1.00 | our implementation | |
|---|---|---|
| `"scheme": "uav:channelsec"` | `"uav:channel-security"` | differs |
| `"scheme": "uav:authentication"` | `"uav:authentication"` | **matches** |
| `uav:securityMode`: `None` / `Sign` / `SignAndEncrypt` | `messageMode`: `none` / `sign` / `sign_encrypt` | name and values |
| `uav:securityPolicy` | `policy` | name |
| `uav:userIdentityToken`: `Anonymous` / `UserName` / `Certificate` / `IssuedToken` | `tokenType`: `anonymous` / `username` / `certificate` | name, casing, and no `IssuedToken` |
| `uav:issueToken` (reference to an OAuth2 scheme) | not implemented | missing |

So one scheme name is already right and the other is not; the property names inside both are the ones that diverge.

### Some OPC UA background, since the two schemes are not interchangeable

OPC UA has **two independent security layers**, and the spec models them as two separate WoT schemes deliberately:

- **Channel security** (`uav:channelsec`) — protects the *connection*. Two settings:
[...]

PR #1415, opened 2025-09-02
feat(binding-opcua): Add channel security support  #1401

erossignon:
Implements OPC UA channel security by allowing the configuration of security mode and policy and authentication

This change introduces a certificate manager to handle client-side certificates.

Key changes:
- Added new SecurityScheme

```code
     OPCUASecureSecuritySchemeBase (abstract)
         -> OPCUAChannelSecurityScheme
	       -> OPCUAUnsecureChannelScheme
	       -> OPCUASecureSecurityScheme
	 -> OPCUASecureSecurityScheme 
               -> OPCUACertificateAuthenticationScheme 
               -> OPCUAUserNameAuthenticationScheme
```

- Implemented a shared `OPCUACertificateManager` for PKI.

  PKI folder for OPCUAPrococolClient is set to env-path("binding-opcua-wot").PKI

    which resolve to
        on window:
	  C:\Users\<User>\AppData\Roaming\binding-opcua-node-wot\Config\PKI
	on linux:
	  ~/.config/binding-opcua-node-wot/PKI on linux'
	on macOs:
	  ~/Library/Application Support/binding-opcua-node-wot/PKI

- Added tests for secure communication.


packages/core/src/thing-description.ts
```diff
@@ -181,6 +184,82 @@ export interface OAuth2SecurityScheme extends SecurityScheme, TDT.OAuth2Security
...
+}
+
+export type OPCUAChannelSecurityScheme = OPCUASecureSecurityScheme | OPCUAUnsecureChannelScheme;
+export interface OPCUACAuthenticationSchemeBase extends OPCUASecuritySchemeBase {
+    scheme: "opcua-authentication";
+    tokenType: "username" | "certificate" | "anonymous";
+}
+
+export interface OPCUACUserNameAuthenticationScheme extends OPCUACAuthenticationSchemeBase {
+    scheme: "opcua-authentication";
```
danielpeintner: Please take a look at https://www.w3.org/TR/wot-thing-description11/#example-49

We should use the `uav` prefix (I think that was the prefix agreed on for OPCUA)

`opcua-authentication`
becoming
`uav:authentication`

The same goes for other schemes

erossignon: Thanks, done

relu91:
PR looks good, but I have these two questions/suggestions:
- ClientManager logic can be factored out in a single ts file/class
- Are we all ok to expose and add OPCUA specific typings and test to the core package (which should be agnostic from the underlying protocol)?

packages/binding-opcua/src/opcua-protocol-client.ts
```diff
@@ -464,16 +536,95 @@ export class OPCUAProtocolClient implements ProtocolClient {
...
+                        // pick the first one for now
+                        // later we might use credentials to select the most appropriate one
+                        success = this.setSecurity([combo.oneOf[0]], credentials);
+                    } else {
+                        success = false;
+                    }
+                    break;
+                }
+                default:
+                    // not for us , ignored
```
danielpeintner, line 552: We _could_ log other schemes that start with "uav:".
They should be handled by the code, but currently aren't ...

Not sure if it is worth doing.
Hence, more a question than a proposal...

erossignon:
> * Are we all ok to expose and add OPCUA specific typings and test to the core package (which should be agnostic from the underlying protocol)?

Can you clarify what you have in mind ?

erossignon:
> Created https://github.com/eclipse-thingweb/node-wot/issues/1419

I can take it while I am here.

danielpeintner (approved):
LGTM

relu91:
Maybe I'm wrong, but I think you missed this:
> ClientManager logic can be factored out in a single ts file/class

relu91:
Ah, now I noticed a little bit of inconsistency in the file naming scheme, sometimes using _ sometime - . We can fix it later with a refector PR.

Issue #1416, opened 2025-09-02
ComboSecurityScheme missing and required for OPCUA Security


erossignon:
In thing-description.ts there is a todo for ComboSecurityScheme

https://github.com/eclipse-thingweb/node-wot/blob/8cd451b30f7b2f7bc223e895a56d1d52c7809b8f/packages/core/src/thing-description.ts#L138

It occur that ComboSecurityScheme is required to create OPCUA Security Scheme

An OPCUA Secure connection has two levels:

   - a SecureChannel      (encryption mode) 
   - a Authentication     ( userName password or x509 certificate)

Therefore it make sense to define a secure OPCUA  this way:

```js
const thing = {

   securityDefinitions: {
 
        "c:sign-encrypt_basic256Sha256": <OPCUAChannelSecurityScheme>{
            scheme: "opcua-channel-security",
            messageMode: "sign_encrypt",
            policy: "Basic256Sha256", // deprecated
        },

        "a:username-password": <OPCUACUserNameAuthenticationScheme>{
            scheme: "opcua-authentication",
            userName: "joe",
            password: "password_for_joe",
        },
        "a:x509-certificate": <OPCUACertificateAuthenticationScheme>{
            scheme: "opcua-authentication",
            tokenType: "certificate",
            certificate: "....",
            privateKey: "....",
        },
        
        "sd1": {
            scheme: "combo",
            allOf: ["c:sign-encrypt_basic256Sha256", "a:x509-certificate"],
        },
        "sd2": {
            scheme: "combo",
            allOf: ["c:sign-encrypt_basic256Sha256", "a:xusername-password""],
        },
    }
};
```

see #1401 and PR #1415

danielpeintner:
Resolved by https://github.com/eclipse-thingweb/node-wot/pull/1417

PR #1417, opened 2025-09-02
feat(core): add support for combo security #1416

erossignon:


packages/core/src/consumed-thing.ts
```diff
@@ -444,14 +445,29 @@ export default class ConsumedThing extends Thing implements IConsumedThing {
...
-            if (ws != null) {
-                scs.push(ws);
+        const visitSchemes = (security: Array<string>) => {
+            for (const s of security) {
+                if (visited.has(s)) {
+                    continue;
+                }
+                visited.add(s);
+
+                const ws = this.securityDefinitions[s + ""]; // String vs. string (fix wot-typescript-definitions?)
```
relu91: I think this issue is solved. You can use:

```suggestion
                const ws = this.securityDefinitions[s]; 
```

packages/core/src/consumed-thing.ts
```diff
@@ -444,14 +445,29 @@ export default class ConsumedThing extends Thing implements IConsumedThing {
...
+                // also push nosec in case of proxy
+                if (ws != null) {
+                    if (ws.scheme === "combo") {
+                        const combo = ws as ComboSecurityScheme;
+                        visitSchemes(combo.allOf as string[]);
```
relu91: Here, the Typescript compiler gets confused because both union types in ComboSecurityScheme allow other keywords. This causes the type of allOf to be `unknown`. We can't do that much here, but we should take into account that **allOf** might be undefined (because we are actually dealing with a ComboSecurityScheme that uses oneOf). In case of oneOf, I'm not sure how the behaviour of this method should look like... therefore I suggest to throw as un-unsupported security scheme.

relu91:
> https://w3c.github.io/wot-thing-description/ontology/wotsec#ComboSecurityScheme

Sorry, but I don't get your point. Are you referring to this statement?

> The oneOf combination is equivalent to using different security schemes on forms that are otherwise identical. In this sense a oneOf scheme is not an essential feature but it does avoid redundancy in such cases.

If that's so, I'm ok to return all the schemes referred in the oneOf field.

erossignon:
I believe we could also handle the case of oneOf , by passing a structure of resolved scheme downward. At the end of the day, the binding need to be made aware of the oneOf security scheme and choose one of them. 
Let me propose something

packages/core/src/thing-description.ts
```diff
@@ -180,6 +181,17 @@ export interface PSKSecurityScheme extends SecurityScheme, TDT.PskSecurityScheme
 export interface OAuth2SecurityScheme extends SecurityScheme, TDT.OAuth2SecurityScheme {
     scheme: "oauth2";
 }
+export interface OneOfSecurityScheme extends SecurityScheme {
+    scheme: "combo";
+    oneOf: SecurityScheme[];
+    allOf: never;
```
danielpeintner, line 187: Is this useful/needed?

packages/core/src/thing-description.ts
```diff
@@ -180,6 +181,17 @@ export interface PSKSecurityScheme extends SecurityScheme, TDT.PskSecurityScheme
...
+}
+export interface AllOfSecurityScheme extends SecurityScheme {
+    scheme: "combo";
+    allOf: SecurityScheme[];
+    oneOf: never;
```
danielpeintner, line 192: Same here, I don't think we should do that. Why should AllOf bother with oneOf

packages/core/src/thing-description.ts
erossignon: Yes, because OneOfSecurityScheme derives from SecurityScheme  it could have 
additonnal an [key:string]: any property.

```
export interface SecurityScheme {
    scheme: string;
    description?: TDT.Description;
    descriptions?: TDT.Descriptions;
    proxy?: TDT.AnyUri;
    [k: string]: unknown;
}
``` 

However, we know from the TD specs that a "combo" can only have allOf or oneOf, meaning that when allOf is specify we known that oneOf will never be there, and vice-versa.
Without `never` we would allow an `allOf` to be added.

```
export interface AllOfSecurityScheme extends SecurityScheme {
    scheme: "combo";
    allOf: SecurityScheme[];
    oneOf: never;
}
```
which mean that it is forbidden to add a oneOf property when allOf is specified, however any other propoerty is allowed ( [k: string]: unknown; in SecurityScheme ^)

This also inform the IDE IntelliSense about the correct state of the object.

packages/core/src/thing-description.ts
erossignon: see above, this is symetrical

packages/core/src/consumed-thing.ts
```diff
@@ -444,15 +447,50 @@ export default class ConsumedThing extends Thing implements IConsumedThing {
...
+                    };
+                }
+                if (combo.oneOf instanceof Array) {
+                    const oneOf = visitSchemes(combo.oneOf as string[]);
+                    return <OneOfSecurityScheme>{
+                        scheme: "combo",
+                        oneOf,
+                    };
+                }
+                return undefined; // not supported , but handled gracefully
```
relu91: @danielpeintner, what do you think here? This is sort of a bad erroneous condition. It means that initial validation of the consumed thing didn't spot an unsupported security scheme, and we are returning an undefined that will result in a final sparse `scs` array (with undefined elements). 

I believe it should be better to throw ...

egekorkan:
> I am just adding this lingk https://w3c.github.io/wot-thing-description/ontology/wotsec#ComboSecurityScheme so we can easily retrieve the section that talks about ComboSecurityScheme in our discussion

I would always point to published standards as that URL is more volatile: https://www.w3.org/TR/2023/REC-wot-thing-description11-20231205/#combosecurityscheme

packages/core/src/consumed-thing.ts
danielpeintner: Agree with @relu91 that throwing in such a case is the better solution

packages/core/test/ClientTest.ts
```diff
@@ -805,4 +805,199 @@ class WoTClientTest {
...
+        const comboScheme = pc.securitySchemes[0] as AllOfSecurityScheme;
+
+        expect(comboScheme.allOf[0].scheme).equals("opcua-channel-security");
+        expect(comboScheme.allOf[1].scheme).equals("opcua-authentication");
+    }
+
+    @test "ensure no infinite loop with recursive combo security"() {
+        const ct = new ConsumedThing(WoTClientTest.servient);
+        ct.securityDefinitions = {
+            // a badly designed combo that goes into infinite loop
```
danielpeintner, line 928: Aborting any possible infinite loop is fine 👍

I wonder how node-wot should behave in such a case. I would tend to say we should throw in such a case.. at least it should not be possible to create a such a thing ...
@relu91 ?

packages/core/src/consumed-thing.ts
erossignon: Instead of returning undefined, we could returning null. This would cause the scheme to be simply ignored. 
In both cases this might cause unpredicable results, expecitally in the credentials settings ( could be ignored or by-passed) so I am i favor of throwing an exception too. ( defensive rather than accommodating )

packages/core/test/ClientTest.ts
relu91: Do we have a valid configuration that contains a possible loop? @egekorkan . I believe that having a loop is definitely an error in the design of the Thing Description. Being merciful or not in this case depends a little bit on us... If we are not throwing, we should at least emit a warning. Throwing it is ok, but if we want to keep it as it is, I'm okay.

egekorkan: The spec is definitely open on that. I would not even allow a reference to a combo scheme from a combo scheme, thus only 1 layer of reference is allowed

erossignon: I see a scenario where 2 levels of combo could be interesting.

In opcua, the first combo level will be to combine ChannelSecurity And Authentication. 

In ChannelSecurity  we could have a Combo that specify  SignEncrypt+Basic256Sha256 OR SignEncrypt+Aes128_Sha256_RsaOaep Letting the client choose one of the two possible ChannelSecurity  depending on what the server actually supports.



    {
       OPCUASecurity: { 
          type: "combo" , 
          allOf: [ "ChannelSecurity", "Authentication"]  
       },
       ChannelSecurity: { 
          type: "combo", 
          oneOf: [ 
             "SignEncrypt+Basic256Sha256", 
             "SignEncrypt+Aes128_Sha256_RsaOaep"
          ]
       },
       "SignEncrypt+Basic256Sha256": { ... },
       "SignEncrypt+Aes128_Sha256_RsaOaep": { ... }
    }

danielpeintner: I can see the use-case from @erossignon

> @egekorkan 
> thus only 1 layer of reference is allowed

I think the TD spec needs some clarification if we really want that limitation.
Hence, I created https://github.com/w3c/wot-thing-description/issues/2134 (I hope that's fine)

Coming back to the node-wot implementation, I am also fine with the current solution in this PR since it needs to be clarified in TD first.

danielpeintner: After some talks with @relu91 and @egekorkan I think we can leave the code as is.
I opened https://github.com/w3c/wot-thing-description/issues/2134, after that decision we can check whether some implementation details are needed.

PR #1418, opened 2025-09-02
fix esm parsing issue in test with node 22.00 by upgrading mocha, and fix "rxjs" package version consistency issue

erossignon:
This will fix recurring failing tests in CI/CD with newest revision of nodeJS.

Two changes, in two separate commits
    - update mocha to latest version
    - fix rxjs": "^5.5.11" consitency issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
                 "eslint-plugin-unused-imports": "^3.0.0",
                 "eslint-plugin-workspaces": "^0.9.0",
                 "husky": "^7.0.4",
-                "mocha": "11.7.2",
+                "mocha": "^11.7.2",
                 "prettier": "3.3.3",
                 "pretty-quick": "^4.0.0",
                 "sinon": "11.1.1",
@@ -52,8 +52,6 @@
         },
         "node_modules/@babel/code-frame": {
             "version": "7.27.1",
-            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
-            "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -70,7 +70,7 @@
         "eslint-plugin-unused-imports": "^3.0.0",
         "eslint-plugin-workspaces": "^0.9.0",
         "husky": "^7.0.4",
-        "mocha": "^11.1.0",
+        "mocha": "11.7.2",
```
danielpeintner: Very minor... why we don't use `"^11.7.2"` to allow updates ?

danielpeintner (approved):
LGTM, and thank you for finding the issue 👍

I will try later whether it fixes also the issues we are seeing in https://github.com/eclipse-thingweb/node-wot/pull/1377

relu91 (approved):
Other than @danielpeintner's suggestion, we can go.

package.json
erossignon: OK done, I also fixed a minor error raised by the "npm check:versions" steps

packages/binding-http/package.json
```diff
@@ -39,7 +39,7 @@
         "find-my-way": "^8.2.2",
         "node-fetch": "^2.7.0",
         "query-string": "^7.1.1",
-        "rxjs": "^5.5.11",
+        "rxjs": "5.5.11",
```
danielpeintner: I think you changed something unrelated now...

packages/binding-websockets/package.json
```diff
@@ -17,7 +17,7 @@
     "dependencies": {
         "@node-wot/binding-http": "0.9.2",
         "@node-wot/core": "0.9.2",
-        "rxjs": "^5.5.11",
+        "rxjs": "5.5.11",
```
danielpeintner: Here as well, nothing to do with Mocha.
Can you please revert these changes

packages/binding-http/package.json
erossignon: The build failed "npm check:versions" stage, complaining that sometime 

we have  "rxjs": "^5.5.11", and sometime  "rxjs": "5.5.11",

The change is need to get the CI/CD passing. it is in a separate commit.

However, as you suggested I have used rxjs": "^5.5.11" instead now.
( this requires many package.json update)

I have updated the title of this PR to reflect the 2 changes, needed to get the CI/CD passing.

packages/binding-websockets/package.json
erossignon: See above, this is required to have the "check consistencyt check" passing

danielpeintner:
Okay let's go with it, even though I am not sure why the inconsistency didn't show up before.

Issue #1419, opened 2025-09-19
SecurityType not used anywhere -> remove?


danielpeintner:
> p.s. I noticed that `SecurityType` is not used anywhere in our codebase I'd deprecate it and remove it in 1.0.0 @danielpeintner

_Originally posted by @relu91 in https://github.com/eclipse-thingweb/node-wot/issues/1415#issuecomment-3311165881_

PR #1420, opened 2025-09-19
chore(thing-description): deprecate unused SecurityType #1419

erossignon:


erossignon:
> sorry I wasn't clear, since we are not yet to 1.0.0 we have first signal the users that we are going to remove this type. Therefore we should simply add `@deprecated` for now.

Thanks for the clarification, I have redone the commit wioth a ``//@deprecated message` instead.

relu91 (approved):

danielpeintner (approved):

Issue #1421, opened 2025-09-22
AutoSecurityScheme required for OPCUA binding


erossignon:
Wot specifies:

> 4.1.2 AutoSecurityScheme
> IRI: https://www.w3.org/2019/wot/security#AutoSecurityScheme
> An automatic authentication security configuration identified by the term auto (i.e., "scheme": "auto"). This scheme indicates that the security parameters are going to be negotiated by the underlying protocols at runtime, subject to the respective specifications for the protocol (e.g. [[RFC8288](https://www.w3.org/2019/wot/security#bib-rfc8288)] for Basic Authentication when using HTTP).

OPC Wot Binding RC specifies:

<img width="1469" height="362" alt="Image" src="https://github.com/user-attachments/assets/1e8c664e-3787-42b5-94e7-6da21f873aa5" />

Therefore we need to implement the generic AutoSecurityScheme in core, so it can be used in binding-opcua.

danielpeintner:
Resolved by https://github.com/eclipse-thingweb/node-wot/pull/1425

PR #1422, opened 2025-09-22
feat(core): add support for AutoSecurityScheme #1421

erossignon:


packages/core/src/thing-description.ts
```diff
@@ -193,6 +193,9 @@ export interface AllOfSecurityScheme extends SecurityScheme {
 }
 export type ComboSecurityScheme = AllOfSecurityScheme | OneOfSecurityScheme;
 
+export interface AutoSecurityScheme extends SecurityScheme, TDT.AutoSecurityScheme {
```
danielpeintner: VERY minor: I would have put `AutoSecurityScheme` right after `NoSecurityScheme` so that the order matches with `SecurityType`.

Since `SecurityType` gets removed anyway I REALLY don't mind

danielpeintner (approved):

packages/core/src/thing-description.ts
erossignon: so resolving the discussion

erossignon: OK, I changed it by ordering this way:

```
    | NoSecurityScheme
    | BasicSecurityScheme
    | AutoSecurityScheme
```

which seems to me the most logical

danielpeintner: Everyone seems to have another ordering in mind ;-)

See below the TD ordering

grafik

Anyhow, I think it doesn't matter and is fine as is.

Issue #1423, opened 2025-09-22
Inconsistencies in naming convention inside opc-ua packaged


relu91:
Ah, now I noticed a little bit of inconsistency in the file naming scheme, sometimes using _ sometime - . We can fix it later with a refector PR.

_Originally posted by @relu91 in https://github.com/eclipse-thingweb/node-wot/issues/1415#issuecomment-3319675827_

erossignon:
also in cli, and core packages

danielpeintner:
Resolved by https://github.com/eclipse-thingweb/node-wot/pull/1424 and others

PR #1424, opened 2025-09-23
fixing source filename to follow kebab-case convention and other eslint issues #1423

erossignon:


erossignon:
- First commit migrates to eslint 9.

The main difference is the introduction of eslint.config.mjs instead of eslintrc.json.

- A few eslint packages have been outdated like eslint-node replaced with eslint-n.

- standard typescript-eslint recommanded issues new errors that I have turned to warnings for the time being, We can decide to enforce them back one at a time, and fix the code at a later date. 

- I have added the kebab-nameing rules, it works well and detect issues in cli, core, and binding-opcua. 

- the following commits fixes kebab naming in cli, core, and binding-opcua

- finally, the latest commit enfore the kebab-nameing rule to be an errro.

egekorkan (approved):
Thank you for the changes and I am fine with them

erossignon:
> . I have just a question, why use react naming convention when is already provided by typescript-eslint?


the react-naming-convention plug-in was suggested to us by @danielpeintner.

I have been scrutinizing  https://typescript-eslint.io/rules/ and cannot find an equivalent for filename enforcement.

erossignon:
> What's wrong with this? https://typescript-eslint.io/rules/naming-convention/


The rule naming-convention is about naming things inside the code ( like variables, clases, members). We are looking at naming source filename.

packages/binding-opcua/src/certificate-manager-singleton.ts
```diff
@@ -47,7 +47,9 @@ export class CertificateManagerSingleton {
         if (CertificateManagerSingleton._certificateManager) {
             CertificateManagerSingleton._certificateManager.referenceCounter--;
             // dispose is degined to free resources if referenceCounter==0;
-            CertificateManagerSingleton._certificateManager.dispose();
+            CertificateManagerSingleton._certificateManager.dispose().catch((err) => {
+                console.error("Error disposing CertificateManager", err);
```
danielpeintner: 2 points
* I think we should use the logger not the console
* I wonder whether in some case we should rather throw

relu91: Also, can't this be changed to be async? catching like that is an anti-pattern.

erossignon: Good catch, I'll fix it, ( console: I shouldn't have relied on co-pilot so blindly 🫤 )

erossignon:
> I think we should use the logger not the console

Good point, let's activate the "no-console-rule": "warn" too ...

packages/binding-http/test/http-server-oauth-tests.ts
```diff
@@ -26,13 +26,6 @@ should();
 class OAuthServerTests {
     private server!: HttpServer;
     async before() {
-        // eslint-disable-next-line @typescript-eslint/no-empty-function
```
danielpeintner, line 29: @relu91 the code was there to have a clean test log, right. Do we still need/want it?

erossignon: I checked that the test log is remaining clear, without it . 
For me it was not needed anymore

erossignon:
> Ah right! Then I am more confident to add a vendor-neutral ESLint plugin like this: https://github.com/DukeLuo/eslint-plugin-check-file, if for @danielpeintner is ok.


I now switched to eslint-plugin-check-file in ad4244c

eslint.config.mjs
```diff
@@ -0,0 +1,160 @@
...
+            "check-file": checkFile,
+        },
+
+        rules: {
+            // *************** Ensure that copyright notice is present ***************
+            "notice/notice": [
+                "error",
+                {
+                    mustMatch: "Copyright \\(c\\) [0-9]{0,4} Contributors to the Eclipse Foundation",
+                    templateFile: __dirname + "/license.template.txt",
```
danielpeintner, line 65: We don't have this file?

eslint.config.mjs
```diff
@@ -0,0 +1,160 @@
...
+                    onNonMatchingHeader: "replace",
+                },
+            ],
+
+            // *************** NodeJS specific rules - relaxing default setting of n  ***************cs
```
danielpeintner: "cs" after **** ?

packages/binding-http/test/http-server-oauth-tests.ts
relu91: Yes, probably they were left there as a leftover of the period where we had console log instead of `debug`

relu91 (approved):
I would go ahead, as this PR is getting too big. However, I have a question: What is the newly introduced licence.template.txt? 

Finally, I would turn off the no-console rule for the examples globally instead of manually adding it to each file (To be addressed possibly in the future).

danielpeintner:
> I would go ahead, as this PR is getting too big.

I agree and we got 3 approvals also 👍

PR #1425, opened 2025-09-23
Feature: add auto secure scheme support in binding-opcua

erossignon:


packages/binding-opcua/src/find-most-secure-channel.ts
```diff
@@ -0,0 +1,119 @@
...
+        case null:
+        case "":
+        case "http://opcfoundation.org/UA/SecurityPolicy#None":
+            return 0;
+        case "http://opcfoundation.org/UA/SecurityPolicy#Basic128":
+            return 1 + encryptWeight;
+        case "http://opcfoundation.org/UA/SecurityPolicy#Basic192":
+            return 2 + encryptWeight;
+        case "http://opcfoundation.org/UA/SecurityPolicy#Basic192Rsa15":
+            return 2 + encryptWeight;
```
danielpeintner: If you don't want to increase the priority I think it is clearer to use the *same* return statement with multiple case statements like you did for `null` and `""`

packages/binding-opcua/src/find-most-secure-channel.ts
```diff
@@ -0,0 +1,119 @@
...
+        case "http://opcfoundation.org/UA/SecurityPolicy#Basic256":
+            return 3 + encryptWeight;
+        case "http://opcfoundation.org/UA/SecurityPolicy#Basic256Rsa15":
+            return 3 + encryptWeight;
```
danielpeintner: Same here. Please collapse

packages/binding-opcua/src/find-most-secure-channel.ts
erossignon: That's a typo ! thanks for spotting

packages/binding-opcua/src/find-most-secure-channel.ts
erossignon: Same typo !

danielpeintner (approved):

Issue #1430, opened 2025-09-30
ESLint Warnings


danielpeintner:
With https://github.com/eclipse-thingweb/node-wot/pull/1424 we introduced some *new* eslint warnings.

At the moment, `npm run lint` shows a huge list of warnings. I suggest turning them off and look at them one by one (so that we can fix one by one)
see https://github.com/eclipse-thingweb/node-wot/blob/master/eslint.config.mjs

* [ ] `n/no-missing-import` (see https://github.com/eclipse-thingweb/node-wot/issues/1428)
* [ ] `n/no-extraneous-import` (see https://github.com/eclipse-thingweb/node-wot/issues/1437)
* [x] `n/no-unpublished-import` (see https://github.com/eclipse-thingweb/node-wot/pull/1438)
* [ ] `n/no-unsupported-features/node-builtins` (see https://github.com/eclipse-thingweb/node-wot/issues/1439)
* [ ] `n/no-deprecated-api` (see https://github.com/eclipse-thingweb/node-wot/pull/1436)
* [x] `n/no-process-exit` (see https://github.com/eclipse-thingweb/node-wot/pull/1505)
* [x] `no-unused-private-class-members` (see https://github.com/eclipse-thingweb/node-wot/pull/1434)
* [x] `prefer-const` (see https://github.com/eclipse-thingweb/node-wot/pull/1432)
* [x] `unused-imports/no-unused-vars` (see https://github.com/eclipse-thingweb/node-wot/pull/1435)
* [ ] `extraneous-dependencies/no-extraneous-dependencies` (see https://github.com/eclipse-thingweb/node-wot/issues/1442)
* [ ] `@typescript-eslint/no-floating-promises` (see https://github.com/eclipse-thingweb/node-wot/pull/1446)
* [x] `@typescript-eslint/no-require-imports` (see https://github.com/eclipse-thingweb/node-wot/pull/1443)
* [x] `@typescript-eslint/no-empty-object-type` (see https://github.com/eclipse-thingweb/node-wot/pull/1441)
* [x] `@typescript-eslint/prefer-nullish-coalescing` (see https://github.com/eclipse-thingweb/node-wot/pull/1440)

PR #1431, opened 2025-09-30
Avoid eslint warnings

danielpeintner:
By
* turning off some warnings
* fix warnings like unused eslint-disable directive by removing useless command 


egekorkan (approved):
I see the direction and I am fine with it. If most of the ignores are removed by the 1.0 release, it is fine :)

PR #1433, opened 2025-10-07
refactor: turn off warning/error in CLI

danielpeintner:


danielpeintner:
According to https://nodejs.org/api/process.html#processexitcode it is not best practice to use `process.exit(1)` and one could/should use `process.exitCode = 1`.
Anyhow, in the CLI case I think we should exit and hence I ignored those 3 issues in the CLI code.

Any thoughts?

PR #1436, opened 2025-10-07
refactor: avoid deprecated api

danielpeintner:
'url.parse' was deprecated since v11.0.0. Use 'url.URL' constructor instead  n/no-deprecated-api

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,11 @@ jobs:
               run: |
                   sudo echo "127.0.0.1" `hostname` | sudo tee -a /etc/hosts
 
+            - name: Add missing ipv6 hostname to /etc/hosts
+              if: matrix.os == 'macos-latest'
+              run: |
+                  sudo echo "::1" `hostname` | sudo tee -a /etc/hosts
+
             # Enables the output of log messages from the Node.js debug module
             # for workflow runs that are being retried with "debug logging"
             # enabled.


packages/binding-coap/src/coap-client.ts
```diff
@@ -219,7 +219,10 @@ export default class CoapClient implements ProtocolClient {
             hostname: requestUri.hostname ?? "",
             port: requestUri.port != null ? parseInt(requestUri.port, 10) : 5683,
             pathname: requestUri.pathname ?? "",
-            query: requestUri.query ?? "",
+            query:
+                requestUri.search && requestUri.search.length > 0 && requestUri.search[0] === "?"
+                    ? requestUri.search.substring(1)
+                    : "",
```
danielpeintner, line 225: It is a bit clumsy since `requestUri.search` has a leading `?` while the _old_ `requestUri.query` did not have it

`?v=6xJ`
vs
`v=6xJ`

@JKRhb please have a look

packages/binding-coap/test/coap-server-test.ts
```diff
@@ -219,7 +219,7 @@ class CoapServerTest {
 
         await coapServer.expose(testThing);
 
-        const uri = `coap://[::1]:${coapServer.getPort()}/test/`;
+        const uri = `coap://[::]:${coapServer.getPort()}/test/`;
```
danielpeintner: Server started with ::

danielpeintner: No matter how I change it, it fails on Linux/Mac 🤷‍♂️

packages/binding-coap/src/coap-client.ts
JKRhb: I think I ran into similar problems a lot time ago, also in `node-coap` :/ Unfortunately, the [URL API](https://github.com/whatwg/url) is really awful IIRC, both when it comes to query parameters and handling IPv6 addresses. In some aspects, it is also not compliant with [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) if I am not mistaken, which is one reason I didn't transition to the URL API back in the day.

If I hopefully find the time, I will take a closer look at this PR later this week

JKRhb: I just saw that the goals of the ["URL standard"](https://url.spec.whatwg.org/) include the following:

> Align RFC 3986 and RFC 3987 with contemporary implementations and obsolete the RFCs in the process. 

That contributes to quite a frustrating experience when working with "proper" URIs using this API. Therefore, we might need to consider using a different implementation here instead (maybe [`fast-uri`](https://github.com/fastify/fast-uri) by fastify). Although I am not very happy about that, as I would usually prefer to use the standardized API (if it was actually usable :frowning_face:).

danielpeintner: I mean, if we fail to use the *new* libraries, we should stick with what we have...

PR #1441, opened 2025-10-07
refactor: activate eslint @typescript-eslint/no-empty-object-type

danielpeintner:
disable also one line check

### DIFF
--- a/packages/core/src/protocol-helpers.ts
+++ b/packages/core/src/protocol-helpers.ts
@@ -25,7 +25,7 @@ export interface IManagedStream {
     nodeStream: Readable;
     wotStream: ReadableStream;
 }
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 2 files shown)


packages/core/src/protocol-helpers.ts
```diff
@@ -25,7 +25,7 @@ export interface IManagedStream {
     nodeStream: Readable;
     wotStream: ReadableStream;
 }
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-empty-object-type
 function ManagedStream<TBase extends new (...args: any[]) => {}>(Base: TBase) {
```
relu91: This should work without ignoring eslint rule. 
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
function ManagedStream<TBase extends new (...args: any[]) => object>(Base: TBase) {
```

relu91 (approved):

PR #1443, opened 2025-10-07
refactor: activate eslint @typescript-eslint/no-require-imports

danielpeintner:
some need to stay because of usage/limitations


packages/core/src/consumed-thing.ts
```diff
@@ -39,7 +39,7 @@ import ContentType from "content-type";
 
 import ContentManager from "./content-serdes";
 
-import UriTemplate = require("uritemplate");
+import * as UriTemplate from "uritemplate";
```
erossignon: there should be no need for `* as`

PR #1449, opened 2025-10-15
chore(binding-opcua) fix application/octet-stream encoding/decoding #…

### LINKED ISSUE #1400: ContentType seems to confuse OPC UA

erossignon:
…1400

### DIFF
--- a/packages/binding-opcua/src/opcua-protocol-client.ts
+++ b/packages/binding-opcua/src/opcua-protocol-client.ts
@@ -16,7 +16,6 @@
 import { Subscription } from "rxjs/Subscription";
 import { promisify } from "util";
 import { Readable } from "stream";
-import { URL } from "url";
 import {
     ProtocolClient,
     Content,
@@ -157,6 +156,20 @@ function _variantToJSON(variant: Variant, contentType: string) {
     }
 }
 
+const dataTypeToSchema = new Map<DataType, string>([
+    [DataType.Boolean, "boolean"],
+    [DataType.SByte, "int8"],
+    [DataType.Byte, "uint8"],
+    [DataType.Int16, "int16"],
+    [DataType.UInt16, "uint16"],
+    [DataType.Int32, "int32"],
+    [DataType.UInt32, "uint32"],
+    [DataType.Int64, "int64"],
+    [DataType.UInt64, "uint64"],
+    [DataType.Float, "number"],
+    [DataType.Double, "number"],
+    [DataType.String, "string"],
+]);
 export class OPCUAProtocolClient implements ProtocolClient {
     private _connections: Map<string, OPCUAConnectionEx> = new Map<string, OPCUAConnectionEx>();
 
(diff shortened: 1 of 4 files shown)


packages/binding-opcua/src/opcua-protocol-client.ts
```diff
@@ -600,6 +613,63 @@ export class OPCUAProtocolClient implements ProtocolClient {
             const variantInJson = opcuaJsonEncodeVariant(dataValue.value, false);
             const content = contentSerDes.valueToContent(variantInJson, schemaDataValue, contentType);
             return content;
+        } else if (contentType === "application/octet-stream") {
```
danielpeintner, line 616: I think the general problem with the current code is that it right away decodes the data on the wire. It should first get the data `raw` and afterwards the ContentSerdes decodes the data.

danielpeintner: Take a look at  CoAP for example where the readResource call returns the raw data

https://github.com/eclipse-thingweb/node-wot/blob/8dc6ed8d7a7fd24ef97d703c018a5ec1cd76481a/packages/binding-coap/src/coap-client.ts#L59-L75

Afterwards the configured ContentSerdes deals with the actual content and whether the "byte stream" is JSON, CBOR or anything else..

The same should happen with OPC UA....

erossignon:
> I tried to test the PR with a simple example and I think it does not work as expected. Maybe I am just not understanding it properly.
> 
> As mentioned in #1400 OPC-UA seems to work fine until I add `contentType`.
> 
> Note: FYI, not adding any contentType at all would mean JSON which is not the case for OPC UA I believe.
> 
> If I add `"contentType": "application/octet-stream"` in a `form` the codec tries to do work... which has been done already.
> 
> Also, it seems that the according code hasn't changed in that regard.
> 
> Try to add `"contentType": "application/octet-stream"` in a simple TD and you will get something like.
> 
> ```
> throw new Error(`application/octet-stream only supports primitive types (boolean, integer, number, string) - got ${form.type}`);
>                           ^
> 
> Error: application/octet-stream only supports primitive types (boolean, integer, number, string) - got undefined
> ```

Could you provide an example of things that reproduce what you're doing. That would be easier to grasp . My understanding should be clearly available in the new unit test added. May be I missed your specific case.

packages/binding-opcua/src/opcua-protocol-client.ts
erossignon: > the current code is that it right away decodes the data 
In fact this code is where the encoding takes palce;( not decoding)

danielpeintner:
@erossignon not sure what we do with this PR. I think we head into another direction ...

erossignon:
Yes, I am working on a proposal first. This one is getting unclear to follow anyway 
Stay tuned

erossignon:
@danielpeintner: agreed, let's close this one. Here is where it stands, so the next person does not have to re-derive it.

**What this PR does fix.** On master today, putting `"contentType": "application/octet-stream"` on an OPC UA form fails with `dataType.toLowerCase is not a function`. That is core's generic octet-stream codec trying to process OPC UA content it does not understand. This PR stops that from happening, and that part works.

**What it does not fix, which is the part you ran into.** You reported `only supports primitive types ... - got undefined`. I have now measured it rather than guessed, and your report was correct.

The binding does `switch (form.type)` — it looks for `type` *inside* the form object. But `type` belongs on the affordance, outside `forms`:

```jsonc
"temperature": {
  "forms": [ { "href": "/", "opcua:nodeId": "..." } ],
  "type": "number"          // <-- here, not inside the form
}
```

So `form.type` is `undefined`. The TD author did nothing wrong; the code is looking in the wrong half of the structure.

I tested every combination of where `type` is declared, reading one Double over a real OPC UA connection:

| `contentType` | `type` declared on | result |
|---|---|---|
| omitted | affordance | works, returns JSON |
| `application/octet-stream` | affordance | fails — `got undefined` |
| `application/octet-stream` | form | fails — `No schema type defined` |
| `application/octet-stream` | **both** | **works** |
| `application/octet-stream` | neither | fails — `got undefined` |

So there is exactly one working configuration, and it requires declaring `type` twice. Nobody would write that, and it is not documented. Your test was the normal case, and the normal case is broken.

It is also not fixable by simply reading `type` from the affordance instead: the third row shows that declaring it on the form alone fails *earlier*, because the outer schema validation wants it too. Both layers want a type, from different places.

**Why close rather than rework.** Two reasons.

1. This PR predates the codec split that landed in #1451, so it would need substantial rework against `src/codecs/` regardless.
2. More importantly, the remaining problem is a design question, not a patch: should OPC UA content go through WoT's generic content negotiation at all? You raised the same question with #1402. That deserves to be settled on the issue before anyone writes code, rather than inside a pull request that is now 173 commits behind.

I am writing that up on #1400 with the full measurements. The codec-collision fix from this PR will come back as part of it — it is correct, it just is not sufficient on its own.

Thanks for testing it back in October and for saying plainly that it did not work. That turned out to be the useful part.

PR #1451, opened 2025-10-27
Opcua binding refactor

erossignon:
# code clean up

remove redundant `findBasicType` Code( as already provided by node-opcua with better implementation)

# fix issue: Connection Sharing Bug in OPC UA Client

**Root Cause**
 
A recent change in how the href field is constructed introduced a bug in connection sharing. Previously, the href and opcua:nodeid were separate fields, like this:

```json
{
  "href": "opc.tcp://localhost:4840",
  "opcua:nodeid": "ns=1;s=Temperature"
}
```

Now, the href includes the node ID as a query parameter:

``` json 
{
  "href": "opc.tcp://localhost:4840/?id=ns=1;s=Temperature"
}
```

note: both syntaxes are still supported

Because the href is now unique for each node, the OPCAProtocolClient treats each request as requiring a new connection. This leads to session exhaustion on the server, as connections are not being reused.

**Why this happens**

The `href` was previously used as a key to identify and reuse existing connections. With the new scheme, every `href`is unique, so the client creates a new connection for each operation.

**the fix**
To restore connection sharing, the `href` is first normalized by removing the query parameters before using it as a connection key. This ensures that connections to the same endpoint (e.g., opc.tcp://localhost:4840) are reused, regardless of the node ID in the query.

In the meantime, a refactor and simplification of the connection sharing mecanism has been undertaken

# Refactor codec:

A commit refactors the codec logic in the OPCUA binding into multiple files.
The main `codec.ts` file has been split into:
- `opcua-binary-codec.ts`
- `opcua-data-schemas.ts`
- `opcua-json-codec.ts

### DIFF
--- a/packages/binding-opcua/src/codecs/opcua-json-codec.ts
+++ b/packages/binding-opcua/src/codecs/opcua-json-codec.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.


packages/binding-opcua/src/codecs/opcua-binary-codec.ts
```diff
@@ -0,0 +1,53 @@
+/********************************************************************************
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
```
danielpeintner:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

packages/binding-opcua/src/codecs/opcua-data-schemas.ts
```diff
@@ -0,0 +1,128 @@
+/********************************************************************************
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
```
danielpeintner:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

packages/binding-opcua/src/codecs/opcua-json-codec.ts
```diff
@@ -0,0 +1,155 @@
+/********************************************************************************
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
```
danielpeintner:
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

packages/binding-opcua/src/codecs/opcua-binary-codec.ts
erossignon: The code existed already in 2022 but was in a other file, that why I intentionnaly kept 2022. 
I can change to 2025 if need be?

packages/binding-opcua/src/codecs/opcua-binary-codec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { ContentCodec, DataSchema } from "@node-wot/core";
+import { BinaryStream } from "node-opcua-binary-stream";
+import { DataValue } from "node-opcua-data-value";
+
+import { opcuaJsonEncodeDataValue, opcuaJsonDecodeDataValue, DataValueJSON } from "node-opcua-json";
+
+export class OpcuaBinaryCodec implements ContentCodec {
+    getMediaType(): string {
+        return "application/opcua+octet-stream"; // see Ege
```
danielpeintner, line 24:
```suggestion
        return "application/opcua+octet-stream";
```

danielpeintner: @erossignon can we remove the unnecessary comment? After that, we can merge the PR ...

PR #1452, opened 2025-10-27
Improved opcua example

erossignon:


packages/examples/package.json
```diff
@@ -17,7 +17,18 @@
...
+        "demo:coap:client": "node dist/bindings/coap/example-client.js",
+        "demo:http:server": "node dist/bindings/http/example-server.js",
+        "demo:http:server-secure": "node dist/bindings/http/example-server-secure.js",
+        "demo:http:client": "node dist/bindings/http/example-client.js",
+        "demo:opcua:1": "node dist/bindings/opcua/opcua-demo1.js",
+        "demo:opcua:2": "node dist/bindings/opcua/opcua-demo2.js",
+        "demo:opcua:coffee-machine": "node dist/bindings/opcua/opcua-coffee-machine-demo.js",
+        "quickstart:smart-clock": "node dist/quickstart/smart-clock.js",
+        "quickstart:simple-coffee-machine": "node dist/quickstart/simple-coffee-machine.js",
+        "quickstart:presence-sensor": "node dist/quickstart/presence-sensor.js"
```
danielpeintner: I wonder "what" are the best names and locations ?

Personally, I would not use "demo" and use "example" instead since we are in the `examples` package.
OR, since we are already in examples package the folder they are located in

"demo:coap:server": "node dist/bindings/coap/example-server.js",
-->
"binding:coap:server": "node dist/bindings/coap/example-server.js",

Matches with quickstart also.

What do others think?

packages/examples/src/bindings/README.md
```diff
@@ -0,0 +1,42 @@
...
+await thing.readProperty(...);
+await thing.subscribeEvent(...);
+
+```
+
+Finally, we can shutdown the servient:
+
+```typescript
+await servient.shutdown();
+```
```
danielpeintner: I guess either we add descriptions for all or move the OPCUA explanation in the OPCUA subfolder

danielpeintner (commented):
Thanks. That's a good addition!

See some of my comments below

packages/examples/package.json
erossignon: Good point. let me apply your later suggestion => binding:coap:server etc ...

packages/examples/src/bindings/README.md
erossignon: OK , I moved the OPCUA section into opcua/README.md

packages/examples/src/bindings/opcua/opcua-coffee-machine-thing-description.ts
```diff
@@ -42,68 +53,176 @@ export const thingDescription: WoT.ThingDescription = {
             ],
         },
         waterTankLevel: {
-            // type: "number",
             observable: true,
             readOnly: true,
             forms: [
                 {
-                    href: endpointUrl,
+                    href: "/",
```
egekorkan, line 60: I thought that the example is with the new binding? This href format is not valid anymore

packages/examples/src/bindings/opcua/opcua-coffee-machine-demo.ts
```diff
@@ -25,37 +25,192 @@ const pause = async (ms: number) => new Promise((resolve) => setTimeout(resolve,
...
+            // Standby
+            await pause(1000);
+            state = lastCurrentState;
+        }
+    };
+    const writeLine = (...args: unknown[]) => {
+        process.stdout.write(util.format(...args) + "                         \n");
+    };
+    const displayOnlineStatus = () => {
+        process.stdout.write("\x1B[1;1H"); // move cursor to top left
```
egekorkan, line 66: I guess this is fine but a bit weird to see

packages/examples/src/bindings/opcua/opcua-coffee-machine-thing-description.ts
erossignon: I am not sure what is the "new binding". 

The href="/" because we use the  `base: endpointUrl`, at the top, therefore wot reconstruct the actual href as being base+href. 
So there is no point duplicating the href ein the form ( all elements referes to the same OPCUA server any way, in this example).

danielpeintner:
@erossignon can you take a look at the points @egekorkan raised... thanks!

packages/examples/src/bindings/opcua/opcua-coffee-machine-demo.ts
```diff
@@ -25,37 +25,192 @@ const pause = async (ms: number) => new Promise((resolve) => setTimeout(resolve,
...
                 console.error("Error observing coffeeBeanLevel property:", err);
             });
+        thing
+            .observeProperty("temperature", async (data) => {
+                lastTemperature = (await data.value()) as number;
+                displayOnlineStatus();
+            })
+            .catch((err) => {
+                console.error("Error observing temperature property:", err);
+            });
```
relu91: Since we are at it, can't we just remove all these catch calls and simply log the error inside the catch block of this try? I know that we might lose some granularity (probably not if the error we throw inside the affordance subscription handler is evocative enough)

relu91:
Ok given that @danielpeintner already approved, we can merge this.

PR #1455, opened 2025-10-31
docs: revise README

mkovatsc:
Trying to streamline the information given in the README for "new" users (including people who have not used it in a long time).

I noticed that it is complicated with some of the duplication, e.g., https://thingweb.io/docs/node-wot/api/ or https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-coap/README.md

We also already saw that some of the references online Things are not reachable, e.g., coap://plugfest.thingweb.io/testthing (neither HTTP nor CoAP).

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ You can find more (non-)installation options in the specific [package README](./
 
 ### As a CLI tool
 
-You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool README](./packages/cli/README.md) to find out more.
+You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool README](https://github.com/eclipse-thingweb/node-wot/tree/master/packages/cli) to find out more.
 
 #### As global tool
 


README.md
```diff
@@ -81,27 +77,52 @@ Platforms specific prerequisites:
...
+```
+npm i @node-wot/browser-bundle
+```
+
+You can find more (non-)installation options in the specific [package README](./packages/browser-bundle/README.md).
 
 ### As a CLI tool
 
-You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool's Readme](<[url](https://github.com/eclipse-thingweb/node-wot/tree/master/packages/cli)>) to find out more.
+You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool README](./packages/cli/README.md) to find out more.
```
danielpeintner: Note: I recall we were using the absolute URLs for a good reason - to show links properly if the MD document is shown somewhere else. I would tend to keep it that way...

danielpeintner: Anyhow, I see we don't use it consistently.

relu91: I think the problem arises in npm, there relative links probably don't work. Another reason why we should move most of the doc to a proper documentation site.

danielpeintner: Yes, I also think we should keep the full URLs

relu91: Let's merge this for now, with this change. 
```suggestion
You can alternatively use node-wot via its command line interface (CLI). Please visit the [CLI tool README](https://github.com/eclipse-thingweb/node-wot/tree/master/packages/cli) to find out more.
```

relu91 (approved):

PR #1456, opened 2025-11-28
Update CLI implementation with modern tooling and conventions

relu91:
I know that this PR will be long to review, but it has been sitting on my pc for a while, and it is time to push the changes. The primary goal was to get rid of the VM2 module, which is now deprecated and audited as a **critical** security risk. I used the opportunity to completely revise the implementation and make it:
1. Hopefully easier to maintain
2. More useful and self-explanatory

In summary, this PR delivers a significant refactor and several quality-of-life improvements to the cli module, primarily focusing on simplifying configuration, execution, and improving code organization.

### Key Changes and Improvements

* **Configuration Unification:** Environment variables are now consistently treated as configuration parameters, simplifying how settings are managed and loaded for the CLI execution. 
   * **⚠️ Breaking Change:** Now you should prefix every env with NODE_WOT_ and the cli automatically map that variable to a property in the configuration schema (e.g.,` WOT_SERVIENT_SERVIENT_CLIENTONLY` -> `servient.clientOnly`)
* **Execution Simplification:** Simplified the CLI script execution flow; now we simply require the files, avoiding dependence from virtual machines or complicated compilation steps. Virtualization can always be achieved with other means.
    * **⚠️ Breaking Change:** This drops support for running remote scripts from the Default Servient Thing, a feature that was rarely used and controversial due to security concerns. 
* **Major Refactoring and Testing:**
    * Completed a major refactor of configuration building, configuration sync with the JSON schema, and logging organization.
    * Improved test coverage for the new configuration and parsing functionalities.
* **MQTT Fix :** Allows for optional URI in the MQTT Server configuration.
* **Audit fix:** Now, when you install node-wot, you should not see any audit warnings.

### Open Point for Discussion

Given the extensive changes and the focus on the primary user experience:

* **Should we rename the CLI tool?** The current name, `wot-servient`, is potentially surprising for regular users as it differs from the package name. I propose renaming it to **`node-wot`** to align with the project name and make it feel less surprising.


packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -77,15 +77,14 @@ export default class MqttBrokerServer implements ProtocolServer {
...
-        this.config.uri = this.config.uri ?? this.defaults.uri;
-
         // if there is a MQTT protocol indicator missing, add this
-        if (config.uri.indexOf("://") === -1) {
+        if (config.uri?.indexOf("://") === -1) {
             config.uri = this.scheme + "://" + config.uri;
         }
 
         this.brokerURI = config.uri;
+        this.config = config ?? this.defaults;
```
JKRhb: Can `config` be `null` or `undefined` at this point?

packages/binding-mqtt/src/mqtt-broker-server.ts
```diff
@@ -446,6 +445,10 @@ export default class MqttBrokerServer implements ProtocolServer {
 
     private async startBroker() {
         return new Promise<void>((resolve, reject) => {
+            if (this.brokerURI == null) {
+                throw new Error("Unexpected configuration state broker was started even if brokerURI is null");
```
danielpeintner:
```suggestion
                throw new Error("Unexpected configuration state. Broker was started but brokerURI is null");
```

packages/cli/src/parsers/config-file-parser.ts
```diff
@@ -0,0 +1,31 @@
...
+    try {
+        const open = filename;
+        const data = readFileSync(open, "utf-8");
+        JSON.parse(data);
+        return filename;
+    } catch (err) {
+        if (err instanceof InvalidArgumentError) {
+            throw err;
+        }
+        throw new InvalidArgumentError(`\nError reading config file: ${err}`);
```
danielpeintner:
```suggestion
        throw new InvalidArgumentError(`Error reading config file: ${err}`);
```

I don't think we should start with a new line... any reason for that?

packages/cli/src/parsers/ip-parser.ts
```diff
@@ -0,0 +1,23 @@
...
+ ********************************************************************************/
+import { InvalidArgumentError } from "commander";
+
+export function parseIp(value: string, previous: string) {
+    if (!/^([a-z]*|[\d.]*)(:[0-9]{2,5})?$/.test(value)) {
+        throw new InvalidArgumentError("Invalid host:port combo");
+    }
+
+    return value;
+}
```
danielpeintner: Several comments:
* Parameter previous is never used
* Does it work with ipv6
* do we need the code at all? To me, if a wrong IP is given it should/will fail somewhere else and it is not our job to test valid IPs..

packages/cli/.gitignore
```diff
@@ -0,0 +1,2 @@
+src/generated
+test/resources
```
danielpeintner, line 2: Shouldn't we use the global .gitignore?

Moreover, we add test/resources which in turn says .gitkeep with no files? Is the empty folder needed!?

packages/cli/package.json
```diff
@@ -28,25 +28,27 @@
...
         "commander": "^9.1.0",
+        "debug": "^4.4.0",
         "dotenv": "^16.4.7",
-        "lodash": "^4.17.21",
-        "vm2": "3.9.18"
+        "lodash": "^4.17.21"
     },
     "scripts": {
-        "build": "tsc -b",
+        "build:json": "node import-json.js",
```
danielpeintner: May I suggest renaming the file to something like 
import-json-configuration.js ... or even leave json out.

Initially, I wasn't sure what kind of json gets imported!?


"build:json" --> I don't think the name conveys what is done...

danielpeintner (commented):
Note: I did not fully check nor run it locally, but I would like to make some comments:
* THANK YOU for the effort. I think it is very helpful to get rid of VM2 
* I am surprised that no file has been removed (can it be?)
* some more comments, see below

packages/cli/src/parsers/ip-parser.ts
relu91: We can remove it, but this code is mainly used to parse and validate the inputs for `--inspect` and `--inspect-brk` and it might be useful to fail early rather than brak the nodejs debugger. Nonetheless, I agree with you that the code is not that great as it doesn't cover ipv6... we can remove it for now.

packages/cli/.gitignore
relu91: I think it is better to ignore them in this subfolder as they are specific files for this package. About `.gitkeep` I used another (more solid) approach we can remove it once the tests pass.

packages/cli/package.json
relu91: The script has two purposes:
- import package.json so that we can use the version field safely
- import configuration schema and generate its corresponding ts file

any better naming suggestions? `build:transform` ?  and the file name `import-json-to-ts` ?

danielpeintner: The above naming sounds much better to me 👍

egekorkan:
> Should we rename the CLI tool? The current name, wot-servient, is potentially surprising for regular users as it differs from the package name. I propose renaming it to node-wot to align with the project name and make it feel less surprising.

I like this :) I could not review the rest of the PR yet...

danielpeintner:
> > Should we rename the CLI tool? The current name, wot-servient, is potentially surprising for regular users as it differs from the package name. I propose renaming it to node-wot to align with the project name and make it feel less surprising.
> 
> I like this :) I could not review the rest of the PR yet...

I am fine with changing it. IF WE DO IT, we should publish first a node-wot version and increase the major version to make it clear it is a breaking change.

packages/cli/src/parsers/config-file-parser.ts
```diff
@@ -0,0 +1,31 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
danielpeintner: VERY Minor: I think we _could_ update all copyright years for all new files but I am not sure it is worth it

packages/cli/src/wot-servient-schema.conf.json
```diff
@@ -114,28 +121,12 @@
...
-                {
-                    "properties": {
-                        "level": {
-                            "type": "string",
-                            "enum": ["debug", "info", "warn", "error"]
-                        }
-                    }
-                }
-            ]
+        "logLevel": {
```
danielpeintner: Do we want to list such term changes, so people know about it (`log.level` vs. `logLevel`)?
Not sure if needed...

packages/cli/README.md
```diff
@@ -56,79 +56,44 @@ The `-h` option explains the functionality and also how node-wot can be configur
 The `-h` help option shows the following output:
 
 ```
-Usage: wot-servient [options] [files...]
+Usage: wot-servient [options] [command] [files...]
```
danielpeintner:
```suggestion
Usage: wot-servient [options] [commands] [files...]
```

Use plural?

relu91: confirmed it is autogenerated

packages/binding-mqtt/README.md
```diff
@@ -200,11 +200,11 @@ Please setup node-wot as described at the [node-wot main page](https://github.co
 
 ```
 
-Start the script by the command `wot-servient mqtt-publish.js` or `node ../../packages/cli/dist/cli.js mqtt-publish.js`.
+Start the script by the command `node-wot mqtt-publish.js` or `node ../../packages/cli/dist/cli.js mqtt-publish.js`.
 
 -   example-mqtt-subscription.js: Shows how node-wot consumes a Thing Description to do MQTT subscription on the provided event (=latest counter value) as well as initiate the action (reset counter).
 
-Start the script by the command `wot-servient -c mqtt-subscribe.js` or `node ../../packages/cli/dist/cli.js -c mqtt-subscribe.js`.
+Start the script by the command `node-wotù -c mqtt-subscribe.js` or `node ../../packages/cli/dist/cli.js -c mqtt-subscribe.js`.
```
danielpeintner: There is a typo, but somehow the GH UI doesn't allow me to propose a change

danielpeintner (commented):
Minor issue

danielpeintner:
@relu91 mentioned on Telegram that vm2 efforts revived 
https://github.com/patriksimek/vm2/issues/533#issuecomment-3444774921

Honestly, I don't know which path to go is best... in any case.. even if we would stick to vm2 we should take over some changes Cristiano did in this PR...

packages/binding-mqtt/README.md
relu91:
```suggestion
Start the script by the command `node-wot -c mqtt-subscribe.js` or `node ../../packages/cli/dist/cli.js -c mqtt-subscribe.js`.
```
Trying this probably I didn't enable editing.

relu91:
As stated offline, I think we can leave vm2 for now. We can re-introduce it if we get real feedback from our users to have isolation between wot scripts and the current Node.js project. So far, the usage has been limited.

relu91:
@danielpeintner sadly, I had to re-generate package lock :/ Let me know if you are ok to merge it with these unrelated changes, or if I should try to rebase the whole history from main. That's probably why CI is failing due to coap?

packages/cli/src/utils/load-env-variables.ts
```diff
@@ -0,0 +1,34 @@
...
+ * http://www.eclipse.org/legal/epl-2.0, or the W3C Software Notice and
+ * Document License (2015-05-13) which is available at
+ * https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR W3C-20150513
+ ********************************************************************************/
+import * as dotenv from "dotenv";
+import ErrnoException = NodeJS.ErrnoException;
+
+export function loadEnvVariables(prefix: string = "WOT_SERVIENT_"): { [key: string]: string } {
```
danielpeintner: We change "wot-servient" to "node-wot" as the command.

Shall we do the same here?
NODE_WOT_SERVIENT ?

packages/cli/src/cli.ts
```diff
@@ -54,148 +54,45 @@ Run a WoT Servient in the current directory.
...
-                .map((o: ErrorObject) => o.message)
-                .join("\n")}`
-        );
-    }
-    // Concatenate validated parameters
-    let result = previous ?? {};
-    result = _.merge(result, obj);
-    return result;
+{
+    "$schema": "./node_modules/@node-wot/cli/dist/wot-servient-schema.conf.json"
```
danielpeintner: Wonder again about the name ... shall we rename the file or keep it as is?

packages/binding-mqtt/README.md
```diff
@@ -315,11 +315,11 @@ Please setup node-wot as described at the [node-wot main page](https://github.co
 
 ```
 
-Start the script by the command `wot-servient mqtt-publish.js` or `node ../../packages/cli/dist/cli.js mqtt-publish.js`.
+Start the script by the command `node-wot mqtt-publish.js` or `node ../../packages/cli/dist/cli.js mqtt-publish.js`.
 
 -   example-mqtt-subscription.js: Shows how node-wot consumes a Thing Description to do MQTT subscription on the provided event (=latest counter value) as well as initiate the action (reset counter).
 
-Start the script by the command `wot-servient -c mqtt-subscribe.js` or `node ../../packages/cli/dist/cli.js -c mqtt-subscribe.js`.
+Start the script by the command `node-wot -c mqtt-subscribe.js` or `node ../../packages/cli/dist/cli.js -c mqtt-subscribe.js`.
```
danielpeintner, line 322: Note: I did a search and there are more places that need to be updated
* \Dockerfile
* packages\browser-bundle\web-test-runner.config.mjs
* more?

relu91: Dockerfile is already updated: https://github.com/eclipse-thingweb/node-wot/pull/1456/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

Do I miss something?

egekorkan (approved):

PR #1461, opened 2026-01-11
build: add missing test dependencies to package workspaces to fix lint warnings

### LINKED ISSUE #1442: ESLint Warning extraneous-dependencies/no-extraneous-dependencies
With https://github.com/eclipse-thingweb/node-wot/pull/1424 we introduced some *new* eslint warnings.

One warning is `extraneous-dependencies/no-extraneous-dependencies`

https://github.com/eclipse-thingweb/node-wot/blob/736f2803f3eceac36919d0455d7713272bbcad46/eslint.config.mjs#L92

Which shows for test files the following

Image

and wants that test dependencies are installed as dependency!?!

Not sure what we should do here...

18bharathkumar:
## Summary

This PR fixes `extraneous-dependencies/no-extraneous-dependencies` warnings while preserving the monorepo’s single-source-of-truth policy for shared dependencies.
closes #1442 

Test libraries remain declared only in the root `package.json`. ESLint is updated to be monorepo-aware so workspace packages can correctly resolve hoisted dependencies.

## Changes

- Removed duplicate test dependency declarations from workspace `package.json` files  
- Updated `eslint.config.mjs` to validate dependencies against both root and workspace manifests

## Result

- ✅ `npm run lint` passes  
- ✅ `npm run check:versions` passes  
- ✅ Dependency versions remain centralized and consistent

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,3 +1,4 @@
+import fs from "fs";
 import path from "path";
 import { fileURLToPath } from "url";
 
@@ -24,6 +25,17 @@ const compat = new FlatCompat({
     recommendedConfig: js.configs.recommended,
     allConfig: js.configs.all,
 });
+
+const packageDirs = [__dirname];
+const packagesRoot = path.join(__dirname, "packages");
+if (fs.existsSync(packagesRoot)) {
+    fs.readdirSync(packagesRoot).forEach((dir) => {
+        const pkgPath = path.join(packagesRoot, dir, "package.json");
+        if (fs.existsSync(pkgPath)) {
+            packageDirs.push(path.join(packagesRoot, dir));
+        }
+    });
+}
 
 export default defineConfig([
     ...compat.extends(
@@ -89,7 +101,13 @@ export default defineConfig([
             "n/hashbang": "warn",
 
             // *************** Ensure that only used dependencies are imported ***************
-            "extraneous-dependencies/no-extraneous-dependencies": "warn", // https://github.com/eclipse-thingweb/node-wot/issues/1430
+            // Default rule for root files
+            "extraneous-dependencies/no-extraneous-dependencies": [
+                "warn",
+                {
+                    packageDir: [__dirname],
+                },
+            ],
 
             // *************** Code style and best practices ***************
             "unused-imports/no-unused-imports": "error",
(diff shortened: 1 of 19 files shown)


eslint.config.mjs
```diff
@@ -89,7 +89,7 @@ export default defineConfig([
             "n/hashbang": "warn",
 
             // *************** Ensure that only used dependencies are imported ***************
-            "extraneous-dependencies/no-extraneous-dependencies": "off", // https://github.com/eclipse-thingweb/node-wot/issues/1430
+            "extraneous-dependencies/no-extraneous-dependencies": "warn", // https://github.com/eclipse-thingweb/node-wot/issues/1430
```
danielpeintner:
```suggestion
            "extraneous-dependencies/no-extraneous-dependencies": "warn",
```

danielpeintner (commented):
Thank you 👍

There is a CI consistency check that fails, see https://github.com/eclipse-thingweb/node-wot/actions/runs/20891164482/job/60048975528?pr=1461

eslint.config.mjs
```diff
@@ -25,6 +26,17 @@ const compat = new FlatCompat({
...
+const packagesRoot = path.join(__dirname, "packages");
+if (fs.existsSync(packagesRoot)) {
+    fs.readdirSync(packagesRoot).forEach((dir) => {
+        const pkgPath = path.join(packagesRoot, dir, "package.json");
+        if (fs.existsSync(pkgPath)) {
+            packageDirs.push(path.join(packagesRoot, dir));
+        }
+    });
+}
+
```
relu91, line 39: Any reasons why we are collecting all the package directories?

eslint.config.mjs
```diff
@@ -89,7 +101,13 @@ export default defineConfig([
...
 
             // *************** Ensure that only used dependencies are imported ***************
-            "extraneous-dependencies/no-extraneous-dependencies": "off", // https://github.com/eclipse-thingweb/node-wot/issues/1430
+            // Default rule for root files
+            "extraneous-dependencies/no-extraneous-dependencies": [
+                "warn",
+                {
+                    packageDir: [__dirname],
+                },
+            ],
```
relu91, line 110: Why we don't simply apply it to all?

eslint.config.mjs
18bharathkumar: We collect all package directories to make the ESLint configuration monorepo-aware.

Even though we currently use only __dirname to validate dependencies against the root package.json, collecting package directories establishes package boundaries and allows the rule to be easily extended to validate files against their own package’s package.json without changing the discovery logic.

eslint.config.mjs
18bharathkumar: sorry i didn't get the question

eslint.config.mjs
danielpeintner: Note: I propose to remove the Github link

eslint.config.mjs
danielpeintner: I think @relu91 wants to say "why" we filter out directories. We could apply the rules simply to all ...
@relu91 please step in if you wanted to say something else.

Issue #1458, opened 2025-12-18
I think deriving an object’s URL from its title is a problem


pierre-josselin:
I haven’t found any way to define an object’s URL path other than deriving it from its title, and if that’s really the only option, I think it’s a problem.

I am referring to the mechanism where the title of a Thing Description (e.g. “My counter”) is automatically converted into a URL such as: http://localhost:8080/my-counter

Why is this a problem?

1. **Multiple objects could share the same title**
    If several objects are created with the same title, it becomes nearly impossible to determine their correct URLs.

    [I’ve read that](https://github.com/eclipse-thingweb/node-wot/issues/1376#issue-3045097219) once a slug is generated, if the title is already in use, a suffix (_2, _3, etc.) is appended. However, in a production environment, objects won’t be created manually but imported from a database that may contain thousands of entries.
    
    If I create 50 objects with the same title, there’s no reliable way to know which URL corresponds to which object, except by relying on the order in which they were created, which is neither practical nor robust.

    Finally, if the order of my objects has changed in my database and I restart the WOT server, my-counter_2 might become my-counter_3

2. **Slug generation depends on the tool being used**
    [slugify](https://www.npmjs.com/package/slugify) is used to transform a title into a human-readable, URL-friendly identifier. If I don’t use the exact same slugification library or rules, I may end up with a different result and won’t be able to access the object.
    
    For example:
    [slugify](https://www.npmjs.com/package/slugify) converts "unicode ♥ is ☢" into "unicode-love-is-radioactive"
    [sluga](https://www.npmjs.com/package/sluga) converts "unicode ♥ is ☢" into "unicode-is"
    
    In this case, even knowing the exact title is not enough to reliably reconstruct the URL if the slugification algorithm differs.

3. Using an unique id in the title defeats the purpose of having a meaningful title
    Indeed, it's possible to use the title to define a unique ID that I know won't be altered by slug (for example, a UUID). That's what I'm currently doing, but I lose the entire advantage of having an explicit and human-readable title, precisely to easily identify my object.

**What would be expected?**

The object’s URL should use the ID defined in the thing description exactly as-is, rather than derived from the title. The thing description ID would naturally be required to follow a specific format (here an URN) and be guaranteed to be unique.

If no ID is explicitly provided, the system could fall back to the existing behavior and generate the URL from the title using the current slug mechanism.

**Conclusion**

[...]

PR #1464, opened 2026-01-16
Support ID-based HTTP paths for exposed Things

relu91:
This PR implements support for using Thing id fields to generate HTTP endpoint paths, addressing production concerns with title-based URL derivation. While keeping backward compatibility, it introduces a configurable approach to handle URL path generation.

### Problem Statement
As explained in #1458, the current HTTP binding derives endpoint URLs exclusively from Thing titles through slug generation, which creates several issues:

- Multiple Things with identical titles result in unpredictable URL collisions requiring numeric suffixes
- Slug generation is library-dependent, making URLs unreliable across different implementations
- Developers must embed unique identifiers in titles to ensure stable URLs, defeating the purpose of human-readable titles
Solution

### Proposed behaviour:

Introduce a new configuration parameter in `HttpServer` to use or not the title slugfication to generate Thing paths. For backward compatibility, this feature is true by default. So basically, if the node-wot user wants to have only paths generated with the `thing.id`, it has to:
- Set devFriendlyUri: false in the HTTP server configuration
- Always provide Things with explicit id values (URN format recommended) -> node-wot will default to title if no id is provided.

### Key Changes
- Added devFriendlyUri configuration option to HttpConfig (defaults to true)
- Modified expose() method to support dual-path registration: one from title (when enabled), one from id (when available)
- Both paths route to the same Thing, with proper URL encoding
- Updated destroy() method to clean up both registered paths

> [!NOTE]
> Remember that `id` is optional in [Thing Description 1.1](https://www.w3.org/TR/wot-thing-description11/#thing), therefore, even with `devFriendlyUri` set to `false`, node-wot might generate Things path using the title.


packages/binding-http/src/http-server.ts
```diff
@@ -267,21 +269,32 @@ export default class HttpServer implements ProtocolServer {
...
-            urlPath = uniqueUrlPath;
-        }
-
         if (this.getPort() !== -1) {
-            debug(`HttpServer on port ${this.getPort()} exposes '${thing.title}' as unique '/${urlPath}'`);
-            this.things.set(urlPath, thing);
+            const paths: string[] = [];
+            // If not id is given we create the path using the title even if devFriendlyUri is false.
+            // in Thing Description 1.1 id is optional
+            if (this.devFriendlyUri || thing.id == null) {
```
danielpeintner, line 276: Maybe we can add thing.id == "" as well ?

packages/binding-http/src/http-server.ts
```diff
@@ -267,21 +269,32 @@ export default class HttpServer implements ProtocolServer {
...
+                        uniqueUrlPath = urlPath + "_" + nameClashCnt++;
+                    } while (this.things.has(uniqueUrlPath));
+                    urlPath = uniqueUrlPath;
+                }
+                this.things.set(urlPath, thing);
+                paths.push(urlPath);
+                debug("HttpServer on port %d exposes %s as unique '/%s'", this.getPort(), thing.name, urlPath);
+            }
+
+            if (thing.id != null) {
```
danielpeintner, line 293: If we handle _empty_ thing.id above we need to handle it here as well.

packages/binding-http/src/http-server.ts
```diff
@@ -267,21 +269,32 @@ export default class HttpServer implements ProtocolServer {
...
+                this.things.set(thing.id, thing);
+                paths.push(thing.id);
```
danielpeintner, line 295: Does it make sense to add a debug message if the thing.id has been used before (an we are replacing it). I know it shouldn't be the case ...

packages/binding-http/src/http-server.ts
relu91: Is it a valid TD? Should ID be a URI, right?

danielpeintner: You are right! I checked the schema

PR #1466, opened 2026-01-26
chore: increase version numbers to 1.0.0 for next release

danielpeintner:
FYI: As usual, every time we update the node-wot package versions, we also newly create `package-lock.json` to update dependencies.
Hence, one can see many changes for `package-lock.json` ...

NOTE: please take a look at the changes in `interaction-output.ts`
Without any changes we get
`error TS2322: Type 'Buffer<ArrayBufferLike>' is not assignable to type 'ArrayBuffer'.`
The reason is that buffer (specifically `Buffer<ArrayBufferLike>`) is no longer directly compatible with ArrayBuffer.
There are at least 2 possible fixes.. the one I used is forcing type conversion
The alternative would be wrapping the buffer in a Uint8Array ... causes a test to fail, but that could be fixed.
Opinions?

### DIFF
--- a/packages/core/src/interaction-output.ts
+++ b/packages/core/src/interaction-output.ts
@@ -76,7 +76,7 @@ export class InteractionOutput implements WoT.InteractionOutput {
 
         const data = this.#valueBuffer ?? (await this.#content.toBuffer());
         this.dataUsed = true;
-        const isPooled = data.byteLength < 4096 && data.buffer.byteLength > data.byteLength;
+        const isPooled = data.buffer.byteLength > data.byteLength;
 
         if (isPooled) {
             this.#buffer = data.buffer.slice(data.byteOffset, data.byteOffset + data.byteLength);


packages/core/src/interaction-output.ts
```diff
@@ -75,9 +75,10 @@ export class InteractionOutput implements WoT.InteractionOutput {
 
         const data = await this.#content.toBuffer();
         this.dataUsed = true;
-        this.#buffer = data;
+        // this.#buffer = new Uint8Array(data).buffer
+        this.#buffer = data as unknown as ArrayBuffer;
```
danielpeintner: Which version do people prefer?

relu91: I got deeper into the issue. This is an overview of the problem (polished with AI):

While `this.#buffer = data as unknown as ArrayBuffer` technically "works" in our current ES2015 target—because Buffer happens to implement the same byteLength property and slice() method—it introduces significant issues:
- Leaky Abstractions: If a consumer calls console.log or checks the object type, it reveals a Node.js Buffer instead of the expected ArrayBuffer. This can break environment-agnostic code or unexpected serialization logic.
- Semantic Mismatch: A Buffer is a view into memory, whereas an ArrayBuffer represents the actual memory allocation. Treating them as identical ignores how memory is actually managed under the hood.
- Of course, if we happen to move to later ES targets, the ArrayBuffer interface starts to differ (it has a `transfer` method, for example).

Now, a **proper solution** is to use the underlying memory associated with the Buffer and properly slicing it depending on the length (sometimes, Nodejs pools Buffer allocations). This means that calling `arrayBuffer` method can involve some more work than expected (copying memory). However, I think we can live with that, as my impression is that it is not a commonly used method. Basically, we do this:
```ts
        if (this.#buffer) {
            return this.#buffer;
        }

        if (this.dataUsed) {
            throw new Error("Can't read the stream once it has been already used");
        }

        const data = await this.#content.toBuffer();
        this.dataUsed = true;
        const isPooled = data.byteLength < 4096 && data.buffer.byteLength > data.byteLength;

        if (isPooled) {
           this.#buffer = data.buffer.slice(
                data.byteOffset,
                data.byteOffset + data.byteLength
            )
        } else {
            this.#buffer = data.buffer;
        }
        return this.#buffer;
```
To properly work, you also need to specialize the Buffer returned by the readStreamFully method (which is correct as we do not return a `SharedArrayBuffer`: 
```
static readStreamFully(stream: NodeJS.ReadableStream): Promise<Buffer<ArrayBuffer>> {
```

Next, we need to optimize the value function, as it caches the buffer to be consumed as a buffer if the user wants. We have these options:
1. Don't cache the buffer and don't allow calling `arrayBuffer` after `value` succeeds (need to check the scripting api)
2. Cache it, but don't copy it until the `arrayBuffer` method is called (if it is not called, we don't waste any memory nor cpu cycle)
3. Allow constant overhead for the sake of code readability and simplicity.

relu91: @danielpeintner, if you like, I can commit my full proposal to this branch.

danielpeintner: Sure, please go ahead!

relu91: Done! In the future, we might want to evaluate if the `ArrayBuffer` is the best return type for the `output.arrayBuffer()` method. Had we chosen UInt8Array as the return type, we probably wouldn't have to do that small memory copy overhead.

packages/core/src/interaction-output.ts
```diff
@@ -73,11 +74,17 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
-        this.#buffer = data;
+        const isPooled = data.byteLength < 4096 && data.buffer.byteLength > data.byteLength;
 
-        return data;
+        if (isPooled) {
+            this.#buffer = data.buffer.slice(data.byteOffset, data.byteOffset + data.byteLength);
+        } else {
+            this.#buffer = data.buffer;
+        }
+        //this.#buffer = data as unknown as ArrayBuffer
```
danielpeintner: We can remove this line...

packages/core/src/interaction-output.ts
```diff
@@ -73,11 +74,17 @@ export class InteractionOutput implements WoT.InteractionOutput {
             throw new Error("Can't read the stream once it has been already used");
         }
 
-        const data = await this.#content.toBuffer();
+        const data = this.#valueBuffer ?? (await this.#content.toBuffer());
         this.dataUsed = true;
-        this.#buffer = data;
+        const isPooled = data.byteLength < 4096 && data.buffer.byteLength > data.byteLength;
```
danielpeintner: Can we get rid of 4096 check?

packages/core/src/interaction-output.ts
danielpeintner: removed myself with https://github.com/eclipse-thingweb/node-wot/pull/1466/changes/0a32e9a2b64c1b47c4ac8ff2257bca1db04abcf0

packages/core/src/interaction-output.ts
danielpeintner: Thanks 👍

PR #1474, opened 2026-02-04
Web Thing Protocol Websocket Client Implementation

jthomas-krs:
During the Plugfest we had to implement the Node-WoT web thing protocol websockets to talk to another device. This is the is the work done there. This was a quick draft with testing done against a single device so there could be some edge-cases.


relu91:
I fixed the history of the commits. I couldn't find an easy way to remove myself as a committer of the three commits. However, if you look at the commits themselves, the original author is preserved. I hope @jthomas-krs doesn't mind. This allows us to keep going with this quickly.

danielpeintner:
The only thing that I don't like is the `mode === "wot"` vs. _generic_ WebSocket differentiation 

Anyhow, shall we merge the PR?

Issue #1475, opened 2026-02-06
Support MQTT over websocket


relu91:
Currently, the client library we are using can connect with websocket. However, we are missing a way to properly either describe it in TD or implement it in core. 

To describe, we can:
- `ws` scheme +.`subprotocol:  mqtt`
- `mqtt+ws` (not recommended, but we can handle it): other examples coap+ws, modbus+tcp

To implement:
- We lack support to register multiple schemes. We cannot register `mqtt` and `mqtt+ws`
- Currently subprotocol is handled by the main client protocol. So it means that if we go for `ws` URI scheme + `subprotocol : mqtt`, the code handling mqtt should be copied or imported to the `websocket-binding`, which is not the right way forward. Then we need a way to register combos of URI+subprotocol.

siddharthbaleja7:
@relu91 can you assign this issue to me?

PR #1477, opened 2026-02-10
feat(binding-mqtt): add WebSocket subprotocol and composite scheme 

jona42-ui:
Implemented support for MQTT over WebSocket using both ws:// + subprotocol:"mqtt"
and mqtt+ws:// composite scheme notations, addressing the need for flexible
WebSocket-based MQTT connections.

chnages:

core package

1. **ProtocolClientFactory Interface**: Added `getSupportedProtocols(): Array<[string, string?]>` method returning tuples

mqtt binding

1. **Consolidated Factories**: Single `MqttClientFactory` handles all MQTT variants (mqtt, mqtts, ws+mqtt, wss+mqtt)
2. **MqttClientFactory.getSupportedProtocols()**: Returns `[["mqtt"], ["mqtts"], ["ws", "mqtt"], ["wss", "mqtt"]]`
3. **MqttsClientFactory**: Deprecated, now extends `MqttClientFactory` for backward compatibility

Updated all binding factories to implement `getSupportedProtocols()`:

- HTTP, HTTPS, CoAP, CoAPS, WebSocket (ws/wss), MQTT, Modbus, Netconf, OPC UA, File, M-Bus

We also added tests

Closes https://github.com/eclipse-thingweb/node-wot/issues/1475

### DIFF
--- a/packages/binding-mqtt/src/mqtts-client-factory.ts
+++ b/packages/binding-mqtt/src/mqtts-client-factory.ts
@@ -17,30 +17,15 @@
  * Protocol test suite to test protocol implementations
  */
 
-import { ProtocolClientFactory, ProtocolClient, createDebugLogger } from "@node-wot/core";
 import { MqttClientConfig } from "./mqtt";
-import MqttClient from "./mqtt-client";
+import MqttClientFactory from "./mqtt-client-factory";
 
-const debug = createDebugLogger("binding-mqtt", "mqtts-client-factory");
-
-export default class MqttsClientFactory implements ProtocolClientFactory {
-    public readonly scheme: string = "mqtts";
-    private readonly clients: Array<ProtocolClient> = [];
-
-    constructor(private readonly config: MqttClientConfig) {}
-    getClient(): ProtocolClient {
-        const client = new MqttClient(this.config, true);
-        this.clients.push(client);
-        return client;
-    }
-
-    init(): boolean {
-        return true;
-    }
-
-    destroy(): boolean {
-        debug(`MqttClientFactory stopping all clients for '${this.scheme}'`);
-        this.clients.forEach((client) => client.stop());
-        return true;
+/**
+ * @deprecated Use MqttClientFactory instead. MqttClientFactory now handles both secure and non-secure MQTT protocols.
+ * This class is kept for backward compatibility and simply wraps MqttClientFactory.
+ */
+export default class MqttsClientFactory extends MqttClientFactory {
+    constructor(config: MqttClientConfig) {
+        super(config);
     }
 }
--- a/packages/core/src/protocol-interfaces.ts
+++ b/packages/core/src/protocol-interfaces.ts
@@ -89,6 +89,26 @@ export interface ProtocolClientFactory {
     getClient(): ProtocolClient;
     init(): boolean;
     destroy(): boolean;
+
+    /**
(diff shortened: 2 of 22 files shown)


relu91:
Hi @jona42-ui, welcome! Thanks for proposing changes. Before diving into the review, just notice that the #1475  is not marked as a `good-first-issue`, this usally mean that it requires a good understanding of `node-wot` architecture and the WoT specification. There is nothing inherently bad about contributing to harder issues as newcomers, but it would be better to interact with the maintainers' team before diving into the changes, as they might guide you to what is expected. 

You can reach out to them either on [Discord](https://discord.com/channels/1081253871688622181/1357381845120978984) or [Telegram](t.me/nodewot) (again, not strictly required but recommended for this kind of "architectural" issues). You can also interact with the issue itself.

packages/core/src/servient.ts
```diff
@@ -161,16 +161,41 @@ export default class Servient {
...
+            if (schemes && schemes.includes(normalizedScheme)) {
+                debug(
+                    `Servient found client factory '${factory.scheme}' supporting extended scheme '${normalizedScheme}'`
+                );
+                return factory.getClient();
+            }
+        }
+
+        throw new Error(`Servient has no ClientFactory for scheme '${normalizedScheme}'`);
     }
```
relu91, line 270: Don't use an iterative search. We had a map that got us a nice, constant access time. Suggestion: use two level map: Map<Protocol, Map<Subprotocol, factory>. Therefore, you also need to change how registration of the Clientfactory is done.

packages/core/src/protocol-interfaces.ts
```diff
@@ -89,6 +89,12 @@ export interface ProtocolClientFactory {
     getClient(): ProtocolClient;
     init(): boolean;
     destroy(): boolean;
+
+    /** @returns all schemes supported by this factory (e.g., mqtt, mqtt+ws, mqtts) */
+    getSchemes?(): string[];
+
+    /** @returns true if this factory can handle the scheme + subprotocol combination */
+    supportsSubprotocol?(scheme: string, subprotocol: string): boolean;
```
relu91: Again, we had a "declarative" approach to avoid looping through all the client factories. I would avoid adding a `supportSubprotcol` method. We should better find a way to declare that this factory can support some specific subprotocols. We could decide to stick with `scheme+subprotocol` pattern; however, it is not really recommended (wdyt @egekorkan ?). Having two properties:
- `getSchemes`
- `getSubprotocols` 

would later create some challenges, as some subprotocols might apply to just one of the schemes. Another option would be to return a tuple `[protocol, subprotocol?]`, similar to `scheme+subprotocol` but more structured.

packages/binding-mqtt/src/mqtts-client-factory.ts
```diff
@@ -28,6 +28,15 @@ export default class MqttsClientFactory implements ProtocolClientFactory {
...
     constructor(private readonly config: MqttClientConfig) {}
+
+    getSchemes(): string[] {
+        return ["mqtts", "mqtt+wss"];
+    }
+
+    supportsSubprotocol(scheme: string, subprotocol: string): boolean {
+        return scheme === "wss" && subprotocol === "mqtt";
+    }
+
```
relu91: Given that we move to support multiple schemes, I think it is better to have just one client-factory that handles all the MQTT protocols and sub-protocols permutations.

Issue #1445, opened 2025-10-09
ExposedThing handleUnsubscribeEvent and handleUnobserveProperty should be async


danielpeintner:
In `ExposedThing` we have 

https://github.com/eclipse-thingweb/node-wot/blob/8dc6ed8d7a7fd24ef97d703c018a5ec1cd76481a/packages/core/src/exposed-thing.ts#L575

and

https://github.com/eclipse-thingweb/node-wot/blob/8dc6ed8d7a7fd24ef97d703c018a5ec1cd76481a/packages/core/src/exposed-thing.ts#L642

Both are not `async` while the counterpart of `handleSubscribeEvent` and `handleObserveProperty` are `async`.
Both call async functions (`unsubscribe`/`unobserveHandler`)

I think we ought to change them. Opinions?

danielpeintner:
> Shall we rely on the linter only ?

I hope that's enough. I _found_ the problem with https://github.com/eclipse-thingweb/node-wot/pull/1446

PR #1480, opened 2026-02-12
fix(core): make handleUnsubscribeEvent and handleUnobserveProperty async

Pranav-0440:
## Fix: Make handleUnsubscribeEvent and handleUnobserveProperty async (#1445)

This PR updates the following methods in `ExposedThing`:

- `handleUnsubscribeEvent`
- `handleUnobserveProperty`

### Motivation

These methods call async unsubscribe/unobserve handlers but were not declared as async, leading to potential floating promises and inconsistency with:

- `handleSubscribeEvent`
- `handleObserveProperty`

This change:

- Makes both methods `async`
- Updates return types to `Promise<void>`
- Awaits the corresponding handlers
- Preserves API symmetry
- Ensures better async correctness

### Validation

- Full build successful
- All tests passing across core and all bindings
- No linting issues

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -698,10 +698,10 @@ export default class CoapServer implements ProtocolServer {
 
         res.on("finish", (err: Error) => {
             error(`CoapServer on port ${this.port} failed on observe with: ${err.message}`);
-            thing.handleUnobserveProperty(affordanceKey, listener, interactionOptions);
+            void thing.handleUnobserveProperty(affordanceKey, listener, interactionOptions);
         });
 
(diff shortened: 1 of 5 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -698,10 +698,10 @@ export default class CoapServer implements ProtocolServer {
 
         res.on("finish", (err: Error) => {
             error(`CoapServer on port ${this.port} failed on observe with: ${err.message}`);
-            thing.handleUnobserveProperty(affordanceKey, listener, interactionOptions);
+            void thing.handleUnobserveProperty(affordanceKey, listener, interactionOptions);
```
danielpeintner, line 701: Shouldn't we handle error cases?

PR #1481, opened 2026-02-13
Add AI agent instructions for repository contributors

egekorkan:
Did it during the meeting to improve AI assisted contributions and reviews

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -3,7 +3,7 @@
 ### **Repository Context**
 
 -   **Project**: Eclipse Thingweb Node-WoT
--   **Language**: TypeScript (97.5%), JavaScript (2.4%)
+-   **Language**: TypeScript, JavaScript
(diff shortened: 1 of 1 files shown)


danielpeintner:
> Browser test is failing but that should not be relevant here

Re-running the failed test made the issue disappear

.github/copilot-instructions.md
```diff
@@ -0,0 +1,187 @@
+## **AI Agent Instructions for eclipse-thingweb/node-wot**
+
+### **Repository Context**
+
+-   **Project**: Eclipse Thingweb Node-WoT
+-   **Language**: TypeScript (97.5%), JavaScript (2.4%)
```
danielpeintner: Do we need such statements?

egekorkan: It seems to be common but the percentages are not needed I think and they change.

egekorkan:
```suggestion
-   **Language**: TypeScript, JavaScript
```

PR #1482, opened 2026-02-13
chore(core): added `@typescript-eslint/no-unnecessary-condition` rule and removed redundant existence checks

### LINKED ISSUE #1177: Enable `@typescript-eslint/no-unnecessary-condition` rule

ViRUS-0-0:
Fixes #1177 : Node-wot/core package

## PR Summary
This PR primarily focuses on cleaning up code in` packages/core` by removing unnecessary conditional checks and adopting safer property access patterns. These changes address potential warnings regarding redundant checks where types guarantee presence.

1. Code Cleanup: Removed checks for properties like `thing.properties`, `thing.actions`, and `evt.forms` where they are guaranteed to exist, simplifying the logic in content-serdes.ts and serdes.ts.
2. Safety Improvements: Replaced direct property checks with `Object.prototype.hasOwnProperty.call` in `consumed-thing.ts`, `exposed-thing.ts`, and `helpers.ts` to ensure safer property existence verification.
3. Test Updates: Updated `content-serdes-test.ts` and `server-test.ts` to align with the simplified logic, removing unnecessary optional chaining or default values that are no longer needed.
4. Refactoring: General cleanup of conditional logic to be more concise and type-safe across the core package.

### DIFF
--- a/packages/core/eslint.config.mjs
+++ b/packages/core/eslint.config.mjs
@@ -1,10 +1,10 @@
-import rootConfig from '../../eslint.config.mjs';
+import rootConfig from "../../eslint.config.mjs";
 
 export default [
-  ...rootConfig,
-  {
-    rules: {
-      '@typescript-eslint/no-unnecessary-condition': 'warn'
-    }
-  }
+    ...rootConfig,
+    {
+        rules: {
+            " @typescript-eslint/no-unnecessary-condition": "warn",
+        },
+    },
 ];
--- a/packages/core/eslint.config.mjs
+++ b/packages/core/eslint.config.mjs
@@ -1,0 +1,10 @@
+import rootConfig from '../../eslint.config.mjs';
+
+export default [
+  ...rootConfig,
+  {
+    rules: {
+      '@typescript-eslint/no-unnecessary-condition': 'warn'
+    }
+  }
+];
--- a/packages/core/eslint.config.mjs
+++ b/packages/core/eslint.config.mjs
@@ -4,7 +4,7 @@ export default [
     ...rootConfig,
     {
         rules: {
-            " @typescript-eslint/no-unnecessary-condition": "warn",
+            "@typescript-eslint/no-unnecessary-condition": "error",
         },
     },
 ];
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -50,8 +50,5 @@
     "directories": {
         "test": "test"
     },
-    "keywords": [],
-    "eslintConfig": {
-        "extends": "../../.eslintrc.js"
-    }
+    "keywords": []
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,7 +40,7 @@
         "build": "tsc -b",
         "lint": "eslint .",
         "lint:fix": "eslint . --fix",
-        "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" \"**/*.json\"",
+        "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" \"**/*.json\" \"*.mjs\"",
         "test": "mocha --require ts-node/register --extension ts"
     },
     "bugs": {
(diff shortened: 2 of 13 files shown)


packages/core/eslint.config.mjs
JKRhb, line 1: Seems like there is still a formatting issue present in this file. (I assume it is about the single quotes that are used instead of double-quotes?)

packages/core/src/content-serdes.ts
```diff
@@ -57,29 +57,28 @@ export class ContentSerdes {
...
+        this.instance.addCodec(new TextCodec("image/svg+xml"));
+        // Base64
+        this.instance.addCodec(new Base64Codec("image/png"));
+        this.instance.addCodec(new Base64Codec("image/gif"));
+        this.instance.addCodec(new Base64Codec("image/jpeg"));
+        // OctetStream
+        this.instance.addCodec(new OctetstreamCodec());
+
         return this.instance;
     }
```
JKRhb, line 85: I think this change actually breaks the singleton pattern that is used with ContentSerdes and should be reverted or rewritten.

JKRhb: The fact that the linter fails indicates that we may want to think about keeping the singleton pattern here, or whether we rather want to replace it with a different approach. As this is a bigger decision to be taken here, we may want to ignore this particular check and resolve the issue in a follow-up PR.

ViRUS-0-0: Hey @JKRhb 

I pushed a [change](https://github.com/eclipse-thingweb/node-wot/pull/1482/changes/a6f5dba7c98fee63ab47bc3323ad8bb63cbd2d32). This fixes the lint issue. Can you please take a look and please let me know if this works?

>As this is a bigger decision to be taken here, we may want to ignore this particular check and resolve the issue in a follow-up PR

I'd be happy to  follow-up on this, once we decide on the approch, if my latest [commit](https://github.com/eclipse-thingweb/node-wot/pull/1482/changes/a6f5dba7c98fee63ab47bc3323ad8bb63cbd2d32) doesn't fix it.

packages/core/package.json
```diff
@@ -40,7 +40,7 @@
         "build": "tsc -b",
         "lint": "eslint .",
         "lint:fix": "eslint . --fix",
-        "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" \"**/*.json\"",
+        "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" \"**/*.json\" \"*.mjs\"",
```
danielpeintner, line 43: Do we have .mjs files?

ViRUS-0-0: Yeah, the new `eslint.config.mjs` file for the core package. The file was having a format issue, so I considered adding `.mjs` in the format script.

packages/core/src/consumed-thing.ts
```diff
@@ -130,8 +130,12 @@ class InternalPropertySubscription extends InternalSubscription {
...
     ) {
         super(thing, name, client);
-        const index = this.thing.properties?.[name].forms.indexOf(form as Form);
-        if (index === undefined || index < 0) {
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        if (!this.thing.properties[name].forms) {
+            throw new Error(`Property '${name}' has no forms`);
+        }
+        const index = this.thing.properties[name].forms.indexOf(form as Form);
+        if (index < 0) {
```
JKRhb: I think you might be able to simplify the code to something like this:

```suggestion
        const index = this.thing.properties?.[name].forms.indexOf(form as Form) ?? -1;
        if (index < 0) {
```

I think there might be other places where you have such a pattern.

Issue #876, opened 2022-11-21
Adding selfHost documentation for mqtt


egekorkan:
Currently, the MQTT binding allows a convenience function to self host the mqtt broker. This is visible in https://github.com/eclipse/thingweb.node-wot/blob/master/packages/binding-mqtt/src/mqtt-broker-server.ts#L69 however it is not documented for the users to try it out.

Also  it seems that uri is mandatory now but the example in the readme is not up to date

Something like this should be put in the config of the Readme example:
```ts
servient.addServer(new MqttBrokerServer({
    uri:"localhost:1883",
    selfHost: true
}));
```

PR #1489, opened 2026-02-17
docs(binding-mqtt): Add selfHost documentation for MQTT broker 

jona42-ui:
Added the selfhost documentation example for the mqtt binding

Closes https://github.com/eclipse-thingweb/node-wot/issues/876


relu91 (approved):
Good to go, thanks!

PR #1490, opened 2026-02-17
Support ID-based CoAP paths for exposed Things

Pranav-0440:
This PR implements ID-based path support for the CoAP binding, similar to the HTTP implementation introduced in #1464.

Changes:
- Added `devFriendlyUri` configuration option (default: true)
- Implemented dual-path registration (slugified title + id-based path)
- Ensured proper cleanup of all aliases in `destroy()`
- Added tests verifying reachability via both title and id paths

This maintains backward compatibility while enabling stable, production-safe URIs for CoAP-exposed Things.
Fixes -  #1458

### DIFF
--- a/packages/binding-coap/src/coap-server.ts
+++ b/packages/binding-coap/src/coap-server.ts
@@ -145,14 +145,14 @@ export default class CoapServer implements ProtocolServer {
 
     public async expose(thing: ExposedThing, tdTemplate?: WoT.ExposedThingInit): Promise<void> {
         const port = this.getPort();
-        const paths = this.createThingUrlPaths(thing);
+        const urlPaths = this.createThingUrlPaths(thing);
 
         if (port === -1) {
             warn("CoapServer is assigned an invalid port, aborting expose process.");
             return;
         }
 
-        for (const urlPath of paths) {
+        for (const urlPath of urlPaths) {
             this.fillInBindingData(thing, port, urlPath);
 
             debug(`CoapServer on port ${port} exposes '${thing.title}' as unique '/${urlPath}'`);
@@ -161,10 +161,19 @@ export default class CoapServer implements ProtocolServer {
         }
     }
 
+    /**
+     * Creates the URL paths under which a Thing is exposed.
+     * If devFriendlyUri is enabled, a slugified title-based path is created.
+     * If the Thing has a non-empty ID, an additional ID-based path is created.
+     *
+     * If devFriendlyUri is disabled and the Thing has no ID,
+     * a title-based path is still created to ensure the Thing remains accessible.
+     */
     private createThingUrlPaths(thing: ExposedThing): string[] {
-        const paths: string[] = [];
+        const urlPaths: string[] = [];
+
         // Title-based path
-        if (this.devFriendlyUri || thing.id == null) {
+        if (this.devFriendlyUri || !thing.id) {
             let urlPath = slugify(thing.title, { lower: true });
 
             if (this.things.has(urlPath)) {
@@ -176,13 +185,19 @@ export default class CoapServer implements ProtocolServer {
                 urlPath = uniqueUrlPath;
             }
 
-            paths.push(urlPath);
+            urlPaths.push(urlPath);
         }
+
         // ID-based path
         if (typeof thing.id === "string" && thing.id.length > 0) {
-            paths.push(thing.id);
+            if (this.things.has(thing.id)) {
+                warn(`CoapServer path collision detected for id '${thing.id}'.`);
+            }
+
+            urlPaths.push(thing.id);
         }
-        return paths;
+
+        return urlPaths;
     }
 
     private fillInBindingData(thing: ExposedThing, port: number, urlPath: string) {
--- a/packages/binding-coap/src/coap.ts
+++ b/packages/binding-coap/src/coap.ts
@@ -28,8 +28,27 @@ export * from "./coaps-client-factory";
 export * from "./coaps-client";
 
 export interface CoapServerConfig {
+    /**
+     * Port on which the CoAP server listens.
+     * Defaults to 5683.
+     */
     port?: number;
+
+    /**
+     * Network address to bind to.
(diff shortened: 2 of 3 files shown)


packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
 
     public async expose(thing: ExposedThing, tdTemplate?: WoT.ExposedThingInit): Promise<void> {
         const port = this.getPort();
-        const urlPath = this.createThingUrlPath(thing);
+        const paths = this.createThingUrlPaths(thing);
```
JKRhb:
```suggestion
        const urlPaths = this.createThingUrlPaths(thing);
```

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
 
         if (port === -1) {
             warn("CoapServer is assigned an invalid port, aborting expose process.");
             return;
         }
 
-        this.fillInBindingData(thing, port, urlPath);
+        for (const urlPath of paths) {
```
JKRhb:
```suggestion
        for (const urlPath of urlPaths) {
```

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
-        this.setUpIntroductionMethods(thing, urlPath, port);
+            this.setUpIntroductionMethods(thing, urlPath, port);
+        }
     }
 
-    private createThingUrlPath(thing: ExposedThing) {
-        let urlPath = slugify(thing.title, { lower: true });
+    private createThingUrlPaths(thing: ExposedThing): string[] {
+        const paths: string[] = [];
+        // Title-based path
```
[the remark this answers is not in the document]
JKRhb: I think this comment could rather be used as the basis for a doc comment for the function. Generally speaking, though, I don't think the mirroring of the HTTP binding needs to be mentioned that often.

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
 
-        this.setUpIntroductionMethods(thing, urlPath, port);
+            this.setUpIntroductionMethods(thing, urlPath, port);
+        }
     }
 
-    private createThingUrlPath(thing: ExposedThing) {
-        let urlPath = slugify(thing.title, { lower: true });
+    private createThingUrlPaths(thing: ExposedThing): string[] {
+        const paths: string[] = [];
```
JKRhb:
```suggestion
        const urlPaths: string[] = [];
```

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
+                let uniqueUrlPath;
+                let nameClashCnt = 2;
+                do {
+                    uniqueUrlPath = urlPath + "_" + nameClashCnt++;
+                } while (this.things.has(uniqueUrlPath));
+                urlPath = uniqueUrlPath;
+            }
 
-        return urlPath;
+            paths.push(urlPath);
```
JKRhb:
```suggestion
            urlPaths.push(urlPath);
```

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
-            urlPath = uniqueUrlPath;
-        }
+            if (this.things.has(urlPath)) {
+                let uniqueUrlPath;
+                let nameClashCnt = 2;
+                do {
+                    uniqueUrlPath = urlPath + "_" + nameClashCnt++;
+                } while (this.things.has(uniqueUrlPath));
+                urlPath = uniqueUrlPath;
+            }
```
JKRhb, line 186: Hmm. I was wondering whether we could end up in the following situation here:
- Starting point: two (partial) TDs have the same title, but only the second one has an ID
- For the second TD, a nameclash regarding the title is detected because the first TD did not have an ID and therefore an identical URL path
- The title of the second TD is appended with a `_2`
- The urlPath of the second ID is further expanded by the ID

This is certainly an edge case (I am not actually not certain at the moment whether it is possible to have a missing ID at this point or whether that one is going to be set by the Servient), but I still wanted to mention it here.

packages/binding-coap/src/coap-server.ts
```diff
@@ -143,34 +145,44 @@ export default class CoapServer implements ProtocolServer {
...
 
-        return urlPath;
+            paths.push(urlPath);
+        }
+        // ID-based path
+        if (typeof thing.id === "string" && thing.id.length > 0) {
```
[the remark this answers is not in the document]
JKRhb: Oh, that is actually another interesting observation.

packages/binding-coap/src/coap.ts
```diff
@@ -30,6 +30,7 @@ export * from "./coaps-client";
 export interface CoapServerConfig {
     port?: number;
     address?: string;
```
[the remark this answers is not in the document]
JKRhb: The reference to the HTTP binding could be removed here.

packages/binding-coap/test/coap-server-test.ts
```diff
@@ -73,6 +73,42 @@ class CoapServerTest {
...
+
+        const resp1 = await coapClient.readResource(new Form(uriByTitle));
+        expect((await resp1.toBuffer()).toString()).to.equal('"OK"');
+
+        const resp2 = await coapClient.readResource(new Form(uriById));
+        expect((await resp2.toBuffer()).toString()).to.equal('"OK"');
+
+        await coapClient.stop();
+        await coapServer.stop();
+    }
```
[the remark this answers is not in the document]
JKRhb: Seems like a valid addition at first glance.

PR #1499, opened 2026-02-23
feat(core): implement Thing-level data schema mapping extraction

### LINKED ISSUE #1411: Adding Data Mapping per Thing
As a possible helper for #1409 , we can think of an additional arguement in `consume()` where a "value path" is given. In EdiTDor project we are working on a new PR at https://github.com/eclipse/editdor/pull/136 where a json path can be given for the values to be stripped out of a JSON object. So if the Thing (here the gateway) sends `{value:123, id:abcd, timestamp:2024}` we can put `/value` where only `123` will be presented to the UI. The other parameters are like header information (telemetry) that are inserted by the gateway that are not relevant for the application, so the TD should not have an object. 

This is basically out of band data mapping where having this in the TD 2.0 would remove the need for the implicit knowledge. This is the same thing with Webthings and other such solutions. Allowing a new parameter like this as an argument for the `consume()` method would allow settings this for the entire Thing and can help with OPCUA content type issue too. Application of this value path should happen after content serdes but before passing to the validation.

siddharthbaleja7:
Fixes #1411 

This PR implements the experimental `nw:dataSchemaMapping` feature at the **Thing-level**, allowing developers to extract specific nested values from Thing responses using dot-notation or JSON pointers.

As discussed with the maintainers, this feature strictly targets Thing **responses** ([readProperty](cci:1://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/src/consumed-thing.ts:565:4-591:5), [invokeAction](cci:1://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/src/consumed-thing.ts:732:4-776:5) outputs, and [subscribeEvent](cci:1://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/src/consumed-thing.ts:835:4-890:5) emissions). Input payloads are unaffected.

**Key Implementation Details:**
* **Order of Operations:** Extraction occurs *after* content deserialization but *before* JSON Schema validation inside [interaction-output.ts](cci:7://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/src/interaction-output.ts:0:0-0:0).
* **Servient-Level Injection:** Added `dataSchemaMapping` to the Servient configuration. If a consumed/produced Thing Description lacks a mapping, the Servient's default mapping is automatically injected.
* **Typings:** Added `@experimental` tags to all `nw:` prefixed types in `thing-description.ts`.
* **Zero Dependencies:** Parsing logic is isolated inside [helpers.ts](cci:7://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/src/helpers.ts:0:0-0:0) without relying on external libraries.
* **Strict Typing:** All type casting is safely handled natively without relying on `eslint-disable` `any` casting.

## Testing
* **Added 20+ Unit Tests** in [server-test.ts](cci:7://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/test/server-test.ts:0:0-0:0) and [client-test.ts](cci:7://file:///Users/siddharthbaleja/Desktop/opensource%282026%29/node-wot/packages/core/test/client-test.ts:0:0-0:0) verifying extraction logic, overriding precedence, and error handling.
* Ran and passed all 136 core tests (`npm run test -w packages/core`).
* Passed extreme strict verification (`npm run lint --workspaces` and `npm run test:browser` locally).

## Documentation
* Added an "Experimental Features: Data Mapping per Thing" section to the main `README.md`.


packages/core/src/consumed-thing.ts
```diff
@@ -594,13 +594,22 @@ export default class ConsumedThing extends Thing implements IConsumedThing {
     private handleInteractionOutput(
         content: Content,
         form: Form,
-        outputDataSchema: WoT.DataSchema | undefined
+        outputDataSchema: WoT.DataSchema | undefined,
+        interactionType?: "nw:property" | "nw:action" | "nw:event"
```
danielpeintner: Does it make sense to define those strings (or the enum) once and point to it? Otherwise, we keep adding them at several places as string ...

packages/core/src/consumed-thing.ts
```diff
@@ -620,13 +629,22 @@ export default class ConsumedThing extends Thing implements IConsumedThing {
...
         this.checkMediaTypeOrThrow(content, form);
-        return new ActionInteractionOutput(content, form, outputDataSchema, synchronous);
+
+        let valuePath: string | undefined;
+        const mappedThing = this as unknown as { "nw:dataSchemaMapping"?: Record<string, { "nw:valuePath": string }> };
+        const mapping = mappedThing["nw:dataSchemaMapping"];
+        if (interactionType && mapping && mapping[interactionType] && mapping[interactionType]["nw:valuePath"]) {
+            valuePath = mapping[interactionType]["nw:valuePath"];
+        }
+
```
danielpeintner: I am seeing this (same) code 2 times. Correct?
Does it make sense to move it to a dedicated function?

packages/core/src/consumed-thing.ts
danielpeintner: The same comment goes to all other `nw:` terms

packages/core/src/interaction-output.ts
```diff
@@ -152,8 +160,14 @@ export class InteractionOutput implements WoT.InteractionOutput {
...
     synchronous?: boolean;
 
-    constructor(content: Content, form?: WoT.Form, schema?: WoT.DataSchema, synchronous?: boolean) {
-        super(content, form, schema);
+    constructor(
+        content: Content,
+        form?: WoT.Form,
+        schema?: WoT.DataSchema,
+        synchronous?: boolean,
+        mapping?: DataSchemaMapping
```
danielpeintner: Minor: I wonder whether we should keep the order of the arguments more consistent..
In handleActionInteractionOutput(...) we used to have synchronous after mapping. We could do the same here

packages/core/src/wot-impl.ts
```diff
@@ -107,6 +107,13 @@ export default class WoTImpl {
     async consume(td: WoT.ThingDescription): Promise<WoT.ConsumedThing> {
         try {
             const thing = parseTD(JSON.stringify(td), true);
+            const mapping = { ...(this.srv.dataSchemaMapping ?? {}), ...(thing["nw:dataSchemaMapping"] ?? {}) };
+
+            // If none mapping is configured, the property will be left undefined
```
danielpeintner:
```suggestion
            // If no mapping is configured, the property will be left undefined
```

README.md
```diff
@@ -319,6 +319,51 @@ Below are small explanations of what they can be used for:
 -   Smart Clock: It simply has a property affordance for the time. However, it runs 60 times faster than real-time to allow time-based decisions that can be easily tested.
 -   Simple Coffee Machine: This is a simpler simulation of the coffee machine above.
 
+## Experimental Features
+
+### Data Mapping per Thing
+
+node-wot allows configuration of "Data Mapping" which extracts specific values from a Thing's response object (e.g., getting `123` from a wrapper `{ value: 123, timestamp: ... }`). This is useful when the Interaction only cares about an inner value but the Thing returns a wrapper object.
+
+This is configured using the experimental `nw:dataSchemaMapping` vocabulary in the Thing Description. It can be defined at the Thing level or globally injected via the `Servient` configuration:
```
danielpeintner:
```suggestion
This is configured using the experimental node-wot `nw:dataSchemaMapping` vocabulary in the Thing Description. It can be defined at the Thing level or globally injected via the `Servient` configuration:
```

Otherwise, it sounds like it is an experimental TD feature to me

danielpeintner:
Thanks @siddharthbaleja7 @relu91

PR #1511, opened 2026-03-29
Fix( mqtt ) mqv filter operator precedence

### LINKED ISSUE #1508: mqv:filter is silently ignored due to operator precedence bug
the `mqv:filter` is being "silently ignored" in cases where `requestUri.pathname.slice(1)` returns an empty string "" (which is falsy), but the nullish coalescing operator ?? only triggers on null or undefined, not on empty strings.

foexample:

URL: mqtt://broker.example.com/
Form has: { "mqv:filter": "sensors/+/temperature" }

 Current behavior:

requestUri.pathname = "/"
requestUri.pathname.slice(1) = ""  (empty string, not null/undefined)
filter = "" (mqv:filter is ignored!)

Expected behavior:

 filter should be "sensors/+/temperature"

jona42-ui:
use the logical OR operator (||) instead of nullish coalescing (??)

Fixes https://github.com/eclipse-thingweb/node-wot/issues/1508

### DIFF
--- a/packages/binding-mqtt/src/mqtt-client.ts
+++ b/packages/binding-mqtt/src/mqtt-client.ts
@@ -65,7 +65,10 @@ export default class MqttClient implements ProtocolClient {
         const brokerUri: string = `${this.scheme}://` + requestUri.host;
         // Keeping the path as the topic for compatibility reasons.
         // Current specification allows only form["mqv:filter"]
-        const filter = requestUri.pathname.slice(1) ?? form["mqv:filter"];
+        const filter = requestUri.pathname.slice(1) || form["mqv:filter"];
+        if (filter == null) {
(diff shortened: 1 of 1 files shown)


relu91 (changes requested):
Thank you for the PR! Changes look good. Do you mind adding two tests:
1. Verify that the bug described in #1508 is solved
2. Verify if the form has no path and no filter; the client throws an error. 

I would add them both [here](https://github.com/eclipse-thingweb/node-wot/blob/master/packages/binding-mqtt/test/mqtt-client-subscribe-test.unit.ts). Thanks!

danielpeintner:
@jona42-ui do you still plan to work on some tests? Thank you!

PR #1517, opened 2026-04-23
refactor: update uuid library to resolve a moderate severity alert

danielpeintner:
replaces https://github.com/eclipse-thingweb/node-wot/pull/1516

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15149,12 +15149,16 @@
             }
         },
         "node_modules/uuid": {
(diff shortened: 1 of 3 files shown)


danielpeintner:
Mhh, forgot about eslint rule..

Adding something like teh following does not make sense, right?

// @ts-expect-error
import uuid = require("uuid");

relu91:
solved by #1543

PR #1533, opened 2026-07-16
Add Node.js version 26.x to CI workflow

### LINKED ISSUE #1532: Node.js 26 is available
We might want to test it

danielpeintner:
fixes https://github.com/eclipse-thingweb/node-wot/issues/1532

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
         strategy:
             matrix:
                 os: [macos-latest, ubuntu-latest, windows-2022]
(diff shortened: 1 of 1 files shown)


danielpeintner:
We seem to run into node-mbus issues

danielpeintner:
I created https://github.com/Apollon77/node-mbus/issues/171 to keep track

danielpeintner:
I think 20 October 2026 (about 7 weeks from now) Node.js 26 becomes LTS.

I suggest to wait whether node-mbus is able to fix the problem till then... and otherwise we might need to think about dropping support for it.

PR #1543, opened 2026-07-24
Replace `uuid` dependency with native `crypto.randomUUID()`

relu91:
Replaced the external uuid package with Node.js/Web native crypto.randomUUID().

Why:
- Removes unnecessary third-party dependency (reduces bundle size and supply-chain risk).
- Improves performance by using native Web Crypto / Node.js standard APIs.
- Fully compliant with RFC 4122 v4 UUIDs across modern engines.
- Fixes issues with new ECMAModule compatibility as shown in https://github.com/eclipse-thingweb/node-wot/pull/1539 and https://github.com/eclipse-thingweb/node-wot/pull/1516

The only reasons to keep it are if we want to have other UUID versions or utils (NIL UUID parsing etc.). Currently, we don't use any of those, so I suggest re-evaluating if that is the case.


danielpeintner (approved):
Thanks great!

Issue #1503, opened 2026-03-18
CounterThing example on main page does not work because observeProperty cannot keep up


jspaith:
I copy/pasted the [examples](https://github.com/eclipse-thingweb/node-wot?tab=readme-ov-file#examples) on the welcome readme.md for the counter thing and client.

Repro
* Install various NPM packages 
* Copy the snippets into `counterThing.js` and `counterClient.js`
* Run node on both of these in separate terminals

Behavior -- there are only two notifications fired and the client hangs forever
```
~/developer/jspaith/scratch/wot/test1 >node ./counterClient.js
count: 2
count: 4
```

This suggested a timing issue where the client isn't receiving updates and is hanging because its waiting for 5 before terminating.  It seems that the `observeProperty` is not keeping up. Brute force, but this updated script does work.  

I'm not doing a PR directly to the home page because I don't know if the script needs this paradigm or something is off with underlying `observeProperty`.

```javascript
// client.js
// Required steps to create a servient for a client
const { Servient } = require("@node-wot/core");
const { HttpClientFactory } = require("@node-wot/binding-http");

const servient = new Servient();
servient.addClientFactory(new HttpClientFactory(null));

servient.start().then(async (WoT) => {
    const td = await WoT.requestThingDescription("http://localhost:8080/counter");
    // Then from here on you can consume the thing
    let thing = await WoT.consume(td);

    // Read initial value
    let read = await thing.readProperty("count");
    console.log("initial count:", await read.value());

    for (let i = 0; i < 5; i++) {
        await thing.invokeAction("increment");
        read = await thing.readProperty("count");
        console.log("count:", await read.value());
    }

    // Clean shutdown
    servient.shutdown();
}).catch((err) => { console.error(err); });
```

After this change client works fine
```
~/developer/jspaith/scratch/wot/test1 >node ./counterClient.js
initial count: 0
count: 1
count: 2
count: 3
count: 4
count: 5
~/developer/jspaith/scratch/wot/test1 >
```

Environment:
* WSLv2 / Ubuntu 22
* Windows 11 host
* Node version v18.20.8
* npm version 10.8.2
* I'm using the readme.md associated with #13ea375d323a93646b44bc79386b37703e53f598 in this repo.

PR #1548, opened 2026-08-16
fix: servient.shutdown does not close open http connections

### LINKED ISSUE #1506: `servient.shutdown` does not close open http connections
As stated in https://github.com/eclipse-thingweb/node-wot/issues/1503#issuecomment-4091231642: 
> The hard thing is that await servient.shutdown(); should also work (instead of using the subscription object and stopping it). However, it does not close the subscription as it only calls HTTPFactory destroy -> which does nothing. Probably this is a separate issue, and this one can be closed once the example is updated as above.

A trivial solution could be to keep a reference to the HTTP client created in the HTTPFactory and close it when it is destroyed. But I noticed that we are currently caching clients also at the `ConsumeThing` level, which is a little bit counterintuitive as it should be the binding to decide what and when to cache. Probably a proper review of the current core -> protocol bindings interaction pattern is needed before closing this issue.

DhairyaMajmudar:
## Description

`Servient.shutdown()` did not close active HTTP property observations. Although shutdown completed, long-polling or SSE connections could remain open and prevent the Node process from exiting.

The reason for this was that HTTP and HTTPS client factories did not retain the clients they created, and their `destroy()` methods did not stop those clients.

## References

Closes: https://github.com/eclipse-thingweb/node-wot/issues/1506
Related to: https://github.com/eclipse-thingweb/node-wot/issues/1503#issuecomment-4091231642

cc: @relu91 @egekorkan


packages/binding-http/src/http-client-factory.ts
```diff
@@ -28,20 +28,24 @@ export default class HttpClientFactory implements ProtocolClientFactory {
     public readonly scheme: string = "http";
     private config: HttpConfig | null = null;
     private oAuthManager: OAuthManager = new OAuthManager();
+    private readonly clients = new Set<ProtocolClient>();
```
danielpeintner, line 31: I have a question about why using Set instead of an array. I wonder whether it can happen that a HttpClient may be treated as duplicate... hence creating for example 2 HttpClient instances but having just one entry in the set...

DhairyaMajmudar: @danielpeintner , I used a Set to avoid accidentally tracking the same instance more than once, but I’m happy to use an array for consistency with MqttClientFactory if you prefer.

danielpeintner: @relu91 any opinion

DhairyaMajmudar: Hi @relu91 @danielpeintner, just a small nudge on this one. Thank you!

danielpeintner: FYI @relu91 will be back next week.
I hope it is okay for you to wait a bit...

DhairyaMajmudar: Thank you, @danielpeintner. Sure, no hurry at all.

relu91: Using a Set is not a deal breaker. One may argue it is a bit of overkill since we are looping anyway and closing them all. Anyhow, this work will be obsolete when the new core-protocol interface is finished (see  https://github.com/relu91/thingweb.node-wot/tree/feat/new-protocol-interface). We can merge it now if people need this fix asap.

danielpeintner: I would use an array to align with MqttClientFactory and merge this PR as is.

DhairyaMajmudar: Thanks, both! I’m happy to switch to an array to align with MqttClientFactory and keep the current implementation consistent.

Given that this is a relatively small fix and the new protocol interface is expected to supersede this code anyway, I’m also happy to merge it as-is for now so the issue is addressed in the meantime.

danielpeintner:
The PR looks fine, and @relu91 gave his okay as well. 👍 

I do see one issue in our CI pipeline. 
see https://github.com/eclipse-thingweb/node-wot/actions/runs/34992932503/job/104461873456?pr=1548

Anyhow, this is unrelated to this PR and I will fix it in a follow-up!

danielpeintner (approved):

PR #1553, opened 2026-08-31
chore(binding-opcua): upgrade node-opcua family to 2.179.0

erossignon:
Bumps all node-opcua-* packages to 2.179.0 (node-opcua-crypto to 5.9.0).

These packages release together and must be upgraded as a set, so per-package dependabot PRs can't build. Replaces #1549 and #1551 — both can be closed.

Needed by crypto 5: x509 identity uses clientCertificateManager.getPrivateKey(), and skipLibCheck in the examples tsconfig for the transitive asn1-schema ESM typings.

npm run ncu:opcua is the upgrade procedure, now documented in the binding README. 

All 73 binding-opcua tests pass.


packages/examples/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "outDir": "dist",
         "module": "Node16",
         "rootDir": "src",
+        "skipLibCheck": true,
```
danielpeintner, line 7: This setting is a bit unfortunate, but since it is for `examples` only, I am fine with it.

danielpeintner (approved):

packages/examples/tsconfig.json
danielpeintner: @relu91 do you think it is fine as well?

CONTRIBUTING.md
```diff
@@ -114,6 +114,9 @@ To reduce the size of the installation from about 800 MByte down to about 200 MB
     5. Install the packages with `npm link @node-wot/<module>`
 -   Build error around `prebuild: npm run bootstrap`
     -   This has been seen failing on WSL. Try using a more recent Node.js version
+-   `TS2769: No overload matches this call` on `app.use(...)` in `binding-http` tests, where the argument is a handler that "is not assignable to parameter of type `Application`"
+    -   Two copies of `@types/express` (4.x and 5.x) are installed. `binding-http` uses Express 4, but `@types/cookies` (pulled in by the browser-bundle test runner) asks for `@types/express: "*"`, and a lockfile regeneration can hoist 5.x to the root, where the `express-oauth-server` typings pick it up.
+    -   The root `package.json` pins `@types/express` to 4.x to keep a single copy. Check with `npm ls @types/express`. Remove the pin once `binding-http` moves to Express 5.
```
danielpeintner: @relu91 do you have any advice?

PR #1565, opened 2026-09-15
chore(binding-opcua): upgrade node-opcua family to 2.184.0 + address new OPCUA JSON 1.04 version module

erossignon:
# chore(binding-opcua): upgrade node-opcua to 2.184.4, adopt the split OPC UA JSON API (pinned to 1.04)

## Summary

Upgrades the `node-opcua` family in `@node-wot/binding-opcua` from 2.179.x to **2.184.4**.

This is more than a version bump. Three things land together, because they cannot be
separated — 2.184 will not compile without the first two:

1. **node-opcua is now an ES module**, which `moduleResolution: "node16"` rejects outright.
2. **`node-opcua-json` split its public API** into edition-specific entry points and changed
   every encoder/decoder signature.
3. **The binding is explicitly pinned to the OPC UA 1.04 JSON encoding.** 1.05 is now
   available and is a different wire format; moving to it is deliberately left to a
   follow-up PR.

No behavioural change is intended. The JSON produced on the wire is byte-for-byte what it
was before — see [Why the payloads do not change](#why-the-payloads-do-not-change).

---

## 1. node-opcua 2.184 ships as ES modules

Release [v2.184.0 "Modules, All The Way Down"](https://github.com/node-opcua/node-opcua/releases/tag/v2.184.0)
made all 122 node-opcua packages declare `"type": "module"`. Under the repo-wide
`module`/`moduleResolution: "Node16"` inherited from the root `tsconfig.json`, every import
of node-opcua now fails to compile:

```
error TS1479: The current file is a CommonJS module whose imports will produce 'require'
calls; however, the referenced file is an ECMAScript module and cannot be imported with
'require'. Consider writing a dynamic 'import("node-opcua")' call instead.
```

### The fix: `nodenext`, not an ESM conversion

The upstream [ESM migration guide](https://github.com/node-opcua/node-opcua/blob/master/documentation/migrating_to_esm.md)
is explicit that consumers should *not* convert:

> You do not convert your project to ESM, and it stays CommonJS… `nodenext` is the setting
> that knows a modern Node can `require()` an ES module; `node16` is the one that does not.

So `packages/binding-opcua/tsconfig.json` overrides just the two settings:

```jsonc
"module": "nodenext",
"moduleResolution": "nodenext"
```

This works because Node has supported `require()` of an ES module natively since **22.12**,
node-opcua's engine floor is now **22.13**, and node-opcua gates against top-level `await`
anywhere in its dependency graph (its `check:tla` CI job) precisely so that this stays true.

**The binding itself remains CommonJS.** `dist/index.js` still emits `"use strict"` and
`__createBinding` helpers. `@node-wot/core`, `packages/examples`, `packages/cli` and the
browser bundle are untouched and unaffected. The override is scoped to this one package —
no other workspace package changes its module settings.

<details>
<summary>The alternative that was tried and rejected</summary>

Converting `binding-opcua` itself to ESM (`"type": "module"`) also compiles, but it requires
`.js` extensions on all 34 relative imports across `src/` and `test/`, ESM-interop unwrapping
for the CommonJS `ajv` / `ajv-formats` default imports, and `rxjs/Subscription` →
`rxjs/Subscription.js`. It would also make `binding-opcua` the only ESM package in the
monorepo and change how downstream consumers import it. Since the two-line `nodenext`
override achieves the same result with **zero source changes**, that route was dropped.
</details>

---

## 2. `node-opcua-json` — breaking API change

This is the part that actually touches source code.

### 2a. The package now has edition-specific entry points

| Entry point | Behaviour |
|---|---|
| `node-opcua-json` (root) | Dispatches on `JsonEncodingScheme` at runtime; accepts **either** edition |
| `node-opcua-json/104` | Locked to OPC UA 1.04 — `Reversible`/`NonReversible`, `Type`/`Body`, `TypeId`/`Body`, NodeIds as objects |
| `node-opcua-json/105` | Locked to OPC UA 1.05 — `Compact`/`Verbose`, `UaType`/`Value`, `UaTypeId`/`UaBody`, NodeIds as strings |

Both versioned entry points export the *same version-free names* (`opcuaJsonEncodeDataValue`
and friends), so the edition is chosen once by the import specifier rather than threaded
through every call site.

All binding imports moved from the root to `node-opcua-json/104`. Choosing the root entry
point instead would have made the encoding depend on a runtime scheme argument — a decision
we are not ready to expose through Thing Descriptions yet.

### 2b. Every encoder/decoder signature changed

```diff
-opcuaJsonEncodeDataValue(dataValue, true)
+opcuaJsonEncodeDataValue(dataValue, JsonEncoderMode.Reversible, namespaceArray)

-opcuaJsonDecodeDataValue(pojo)
+opcuaJsonDecodeDataValue(pojo, builder, namespaceArray)
```

- The opaque `boolean` became an explicit **`JsonEncoderMode`** enum. `true`/`false` read as
  nothing at the call site and could not express 1.05's `Compact`/`Verbose`.
- Decoders now take an **`ExtensionObjectBuilder`**, which resolves a DataType NodeId to a
  constructor so structured (ExtensionObject) payloads can be rebuilt.
- Encoders and decoders take a **`namespaceArray: string[]`** for translating
  namespace-qualified NodeIds and QualifiedNames.

Both breaks are **loud, not silent** — `boolean` is not assignable to the enum, and the new
parameters are required, so the compiler flags every affected call.

### Why the payloads do not change

`JsonEncoderMode104` was given numeric values chosen to match the booleans it replaces:

```ts
export declare enum JsonEncoderMode104 {
    Reversible = 1,      // was `true`
    NonReversible = 0,   // was `false`
}
```

So `Reversible` ↔ `true` and `NonReversible` ↔ `false` are exactly value-preserving. Combined
with staying on the `/104` entry point, the encoded JSON is identical to 2.182's output. The
existing codec round-trip tests assert this and pass unchanged.

---

## 3. OPC UA 1.04 vs 1.05 JSON — staying on 1.04 for now

2.184 is the first release to offer both JSON editions. They are **not** compatible on the
wire; OPC UA 1.05 Part 6 Annex H replaces the 1.04 forms outright:

| | 1.04 | 1.05 |
|---|---|---|
| Encoding modes | `Reversible` / `NonReversible` | `Compact` / `Verbose` |
| Variant fields | `Type`, `Body` | `UaType`, `Value` |
| ExtensionObject fields | `TypeId`, `Body` | `UaTypeId`, `UaBody` |
| NodeId | object | string |
| Null ExtensionObject | `null` | omitted entirely in `Compact` |

**This PR deliberately stays on 1.04.** Every consumer of `application/opcua+json` today —
Thing Descriptions, `dataschema` definitions, downstream payload consumers — is written
against the 1.04 shapes. Switching editions inside a dependency upgrade would silently
change every payload the binding emits and reads, which is not something to bundle with an
ESM migration.

Pinning to `node-opcua-json/104` makes that choice **explicit and greppable** rather than
implicit in whatever the root entry point happens to default to.

### Follow-up: 1.05 support (separate PR)

Migrating is a design question, not a mechanical one, and needs its own review:

- How is the edition selected — a form field, a content-type parameter
  (`application/opcua+json;version=1.05`), or a binding-wide option?
- The JSON schemas in `src/codecs/opcua-data-schemas.ts` validate 1.04 shapes and need 1.05
  counterparts.
- What is the deprecation path for existing 1.04 Thing Descriptions?

---

## Other changes in this PR

- **`tsconfig.test.json` (new).** The build config deliberately excludes `test/`, so a type
  error in the test suite previously surfaced only as a confusing mocha/ts-node loader
  failure. `npm run check:types:test` now type-checks the tests directly.
- **`_resolveOutputArguments` — dropped the unused `session` parameter.** It was threaded
  through but never read.
- **String concatenation → template literals** in error messages, and `!parameters || !parameters.dataType`
  → `!parameters?.dataType`. Cosmetic, no behaviour change.
- **`ncu:opcua` script** now passes `--all` so the whole node-opcua family is bumped in step
  rather than just the direct entries.

---

## Known limitations (pre-existing, unchanged by this PR)

Two arguments the new API introduces are currently supplied as stubs. Both are worth
tracking, but neither is a regression — the old API had no way to express them at all:

- **`ExtensionObjectBuilder` is a throwing stub.** `getExtensionObjectConstructor()` raises
  `"Not implemented"`, so decoding a structured (ExtensionObject) payload throws. This
  matches the previous behaviour, where such payloads could not be decoded either. Wiring it
  to the session's DataType factory would be a genuine feature addition.
- **`namespaceArray` is always `[]`.** Namespace-qualified NodeIds and QualifiedNames inside
  payloads are not translated against the server's namespace array. Again unchanged from
  2.182, but the API now makes the gap visible.

---

## Compatibility

⚠️ **node-opcua 2.184 requires Node.js ≥ 22.13**, both for its own engine floor and for the
`require(ESM)` interop this upgrade depends on (Node ≥ 22.12). The repo currently declares no
`engines` field anywhere, so npm will not warn a user on Node 20 — they will hit a runtime
failure instead. Adding `"engines": { "node": ">=22.13.0" }` to `packages/binding-opcua` is
worth considering, either here or as a follow-up. **Happy to add it to this PR if reviewers
prefer.**

---

## Testing

| Check | Result |
|---|---|
| `tsc -b --force` (binding-opcua) | clean |
| `npm run check:types:test` | clean |
| `npm test` (binding-opcua) | **97 passing**, 1 pending |
| `npm run build` (whole monorepo) | clean |
| `npm run lint -w packages/binding-opcua` | clean |

No test was modified to accommodate the upgrade. The only test changes are the mechanical
call-site updates to the new `node-opcua-json` signatures — the assertions, including the
JSON round-trip and schema-validation tests that pin the exact payload shapes, are untouched
and still pass. That is the main evidence that the wire format is unchanged.

---

## Note for reviewers

Because `composite: true` is set, TypeScript does **not** invalidate `.tsbuildinfo` when an
*extended* tsconfig changes. After pulling this branch, run `tsc -b --force` (or delete
`tsconfig.tsbuildinfo`) once, or you may see stale errors that do not reflect the new
settings. This is called out explicitly in the node-opcua 2.184 release notes.

The review is best read in two passes: `tsconfig.json` on its own (the ESM fix, two lines),
then the `src/codecs/**` changes (the JSON API migration), which are repetitive by nature —
the same two-argument-to-three-argument transform at every call site.


danielpeintner (approved):
The changes in binding-opcua look fine.
Anyhow, let's give others some time to review also...

Question: We have some special process when updating package-lock because of node-opcua
see https://github.com/eclipse-thingweb/node-wot/blob/ad9bc99919944d19f52d751dc945265580838a84/README.md?plain=1#L522
Is this still relevant/necessary?

erossignon:
> 
> Question: We have some special process when updating package-lock because of node-opcua see
> 
> https://github.com/eclipse-thingweb/node-wot/blob/ad9bc99919944d19f52d751dc945265580838a84/README.md?plain=1#L522
> 
> 
> Is this still relevant/necessary?

No, this is not relevant if the minimum node version>=22 and npm >=11.

The issue was present with older version of npm as I recall, I haven't add to run `npm dedupe` for a long time now. npm behalves appropriately now.

erossignon:
this PR is required for me to unblock resolution of #1400

danielpeintner:
PR is merged, so everything should be fine.

PR #1572, opened 2026-09-21
feat(core): register codecs per URI scheme (completes #1412)

### LINKED ISSUE #1409: Allow different Codecs for the same contentType

erossignon:
Fixes #1409. Builds on @danielpeintner's #1412: his commit is kept as the first commit, rebased onto master.

**Problem.** `ContentSerdes` holds one codec per media type for the whole servient. `application/octet-stream` is the Modbus-style codec, so an OPC UA form using it gets that codec too. On write, core serializes the value before the binding is called, so the binding cannot step in (#1400).

**Change.** A codec can be registered for a media type and a URI scheme: `servient.addMediaType(codec, false, "opc.tcp")`. On forms whose href uses that scheme it wins; everywhere else the generic codec is used, as today. Core now passes the form's scheme on reads (`InteractionOutput`), writes and actions (`ConsumedThing`), server responses (`ExposedThing`) and notifications (`ProtocolListenerRegistry`).

Second commit, on top of #1412:

| gap in #1412 | fix |
|---|---|
| core never passed a scheme (the TODO) | all call sites above pass `ContentSerdes.schemeOf(form.href)` |
| a media type registered only for a scheme was never found on read | one `findCodec()` for every lookup |
| `isSupported()` ignored the scheme, and `InteractionOutput` checks it before decoding | `isSupported(contentType, scheme?)` |
| scheme-specific codecs were listed as offered | not offered: they belong to one binding |

**Tests.** `content-serdes-test.ts`: precedence, fallback to the generic codec, scheme-only media type, not offered, `schemeOf`. `client-test.ts`: a write and a read through `ConsumedThing` use the scheme's codec. Revert the call sites in `consumed-thing.ts` and `interaction-output.ts` and it fails with `expected 'generic-encoded' to equal 'testdata-encoded'`.

Caveats:
- `addCodec`, `isSupported`, `contentToValue` and `valueToContent` gain an optional last parameter; existing calls are unchanged.
- The key is the href's URI scheme. With MQTT over WebSocket (`ws://` plus `subprotocol: "mqtt"`, #1475) the scheme alone does not identify the binding. Not addressed here.
- Two CBOR tests in core (`Insufficient data`) fail on master as well, locally on Node 26; unrelated.


danielpeintner (approved):
LGTM

@relu91 @egekorkan any concern/feedback?

